PR #1, opened 2019-09-10
Fix build with gcc-7 on linux (WIP) 

falkTX:
Not very clean, needs improvement.
But already gives you an idea of what is needed for compatibility

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,13 @@ build-debug
 .vscode
 perf.data
 perf.data.old
+
+Makefile
+CMakeCache.txt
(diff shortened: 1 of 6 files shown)


paulfd:
Hi, thanks. I like the latest version with the `__has_include`. 

If you want to make it completely compatible with gcc-7 there will be some issues with `<optional>` and `<string_view>`. That being said, since I already rely on abseil, they provide c++11 versions of both and they default to the standard library when compiled with c++17. Maybe using `absl::optional` and `absl::string_view` everywhere would be easier for older compiler versions.

PR #20, opened 2019-12-29
Make the parser include absl headers publicly

jpcima:
It requires absl headers to include Parser from an external program, so make this dependency public.


paulfd:
Thanks, good point! If you're testing things though I suggest the `develop` branch, the `master` is quite old now and we will try to have a first release soon ;-) I overhauled the parser to avoid deep recursions in the regexes in particular...

PR #25, opened 2020-01-17
Make the LV2 volume automatable

jpcima:
It was requested on the channel sfztools.


paulfd:
You're right there's no reason for this port property! Thanks.

PR #26, opened 2020-01-19
Support updated crossfading during note playback

michaelwillis:
@paulfd I took a crack at this, I'm not sure if I got it right though.


paulfd:
Hey! Thanks alot, that's exactly what I had in mind so yeah you got it right :) I just corrected a couple things, so if you agree with the 2 commits I can merge. 

Did you try it? Does it sound as expected?

michaelwillis:
@paulfd I didn't get the results that I expected, but I think it was because of the mistakes that you corrected. Let's go ahead and merge it, and I'll let you know how it goes.

paulfd:
Ok let's see!

Issue #30, opened 2020-01-28
The filters


jpcima:
I attach an implementation of filters which matches ARIA near perfectly. (when implemented there)

It uses some very generic resonant filters, it showed to be quite easy to reproduce them.

Current:
[sfz-filters](https://github.com/jpcima/sfz-filters-experimental)
~[sfzfilters.dsp.gz](https://github.com/sfztools/sfizz/files/4125719/sfzfilters.dsp.gz)~

Reference:
https://sfzformat.com/opcodes/fil_type
https://sfzformat.com/opcodes/fil2_type

- 2 pole filters, cascaded for 4 pole and 6 pole
https://shepazu.github.io/Audio-EQ-Cookbook/audio-eq-cookbook.html

Status:
- [x] lpf_1p
- [x] hpf_1p (buggy in current Plogue ARIA, reported and fixed)
- [x] lpf_2p
- [x] hpf_2p
- [x] bpf_2p
- [x] bpf_1p
- [x] brf_1p (equivalent of brf_2p in Plogue ARIA)
- [x] brf_2p
- [x] lpf_4p
- [x] hpf_4p
- [x] lpf_6p
- [x] hpf_6p

New ARIA extensions:
- [x] lsh
- [x] hsh
- [x] peq

Not implemented by ARIA:
- [x] apf_1p
- [x] lpf_2p_sv
- [x] hpf_2p_sv
- [x] bpf_2p_sv
- [x] brf_2p_sv
- [ ] pkf_2p
- [ ] comb
- [x] pink

Implemented by us, not in specification:
- [x] bpf_4p
- [x] bpf_6p

redtide:
`brf_2p` is listed in SFZ1 [fil_type](https://sfzformat.com/opcodes/fil_type), I kept them separated.

jpcima:
@redtide edited.
Also, added the All-pass and Notch.

jpcima:
Here, I implemented the Hal Chamberlin state-variable filter.
This one fits requirements to implement the `*_sv` opcodes, but needs way more precaution.
[chamberlin.dsp.gz](https://github.com/sfztools/sfizz/files/4127020/chamberlin.dsp.gz)

Reference: https://www.earlevel.com/main/2003/03/02/the-digital-state-variable-filter/

As noted from the article:
[...]

PR #32, opened 2020-01-30
Filters

jpcima:
#30

This is the multi-mode filter.
It has been renamed and moved into the `sfz` namespace.

I have also added the generated code, which is an important amount of code.
To make the PR better, I isolated it in its own commit.


jpcima:
Just a fix for an error just noticed.
Copying the wrong size in case of filter set to none.

```
-                std::memcpy(ch_out, ch_in, NCh * nframes * sizeof(float));
+                std::memcpy(ch_out, ch_in, nframes * sizeof(float));
```

src/sfizz/SfzFilter.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+template <unsigned NCh>
+template <class F>
+void Filter<NCh>::Impl::process(F &filter, const float *const in[NCh], float *const out[NCh], float cutoff, float q, float pksh, unsigned nframes)
+{
+    filter.setCutoff(cutoff);
+    filter.setQ(q);
+    filter.setPkShGain(pksh);
+    filter.compute(nframes, const_cast<float **>(in), const_cast<float **>(out));
```
paulfd, line 125: I'm not a super big fan of the const casts... Is there a reason Faust generates the interfaces as `float**`?

src/sfizz/SfzFilter.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+template <unsigned NCh>
+class Filter {
+public:
+    Filter();
+    ~Filter();
+
+    /**
+       Set up the filter constants.
+       Run it exactly once after instantiating.
```
paulfd, line 31: Is it true? It seems to me this could be changed live.

src/sfizz/SfzFilter.h
```diff
@@ -0,0 +1,100 @@
...
+    void clear();
+
+    /**
+       Process one cycle of the filter without modulating cutoff or Q.
+       `cutoff` is a frequency expressed in Hz.
+       `q` is a resonance expressed in dB.
+       `pksh` is a peak/shelf gain expressed in dB.
+       `in[i]` and `out[i]` may refer to identical buffers, for in-place processing
+     */
+    void process(const float *const in[NCh], float *const out[NCh], float cutoff, float q, float pksh, unsigned nframes);
```
paulfd, line 47: Would you be OK using AudioSpans and Spans in these interfaces? To make it more consistent with the rest of the codebase. I can translate them.

paulfd (approved):
For the rest I think it is OK for now, we can tweak the interface while hooking the code to the rest of the codebase. Thanks!

src/sfizz/SfzFilter.cpp
jpcima: I'm afraid it's the way it is for a long time, unfortunately.
I don't see for it any good reason, it's a thing to deal with.

src/sfizz/SfzFilter.h
jpcima: This sets up only constants dependent on the sample rate.
The sample rate is not changed live. (or is it?)
Run it multiple times then, it should be fine.

src/sfizz/SfzFilter.h
jpcima: On examination, I can't say I love AudioSpan particularly.

It fattens significantly the calling protocol.
It has always 2 pointers (MaxChannels), and 2 longs per instance.

---

`processModulated` Mono
- `raw` : 5 pointers, 1 int
- `AudioSpan` : 10 pointers, 10 longs

---

`processModulated` Stereo
- `raw` : 7 pointers, 1 int
- `AudioSpan` : 10 pointers, 10 longs

---

As I see, it's far of what I would call "0-cost" feature, and my point of view is rather not use this extra expressivity, which requires on such cost.

I think especially of ARM machines which benefits of light call frames, for register-only passing.

In bodies of functions, then yes maybe, but not as part of the function's declaration.

paulfd: There's slightly less than this and it's true that for such big interfaces the cost of possibly redundant lengths add up. As I said we can discuss it later anyway and I am interested about the ARM argument 🙂

PR #34, opened 2020-01-31
Make the build sytem MinGW-compatible

jpcima:
Fixed some aspects of the build system so it can work with MinGW as toolchain.

Notes:
- search for sndfile using pkg-config
- apply `-ffast-math` and friends to all gcc targets, not only `UNIX` ones

Other problems at this time:
- abseil cannot build in MinGW without patching, for various reasons
  - weak symbols not working (which anyway we don't need, could remove if it was an option)
  - `pragma comment (lib)` not supported on this compiler, so it fails linking
- it's impossible to build in 32-bit (not Windows only, all platforms)
  - `#error "Unknown CPU architecture."` in `external/atomic_queue/defs.h`


clients/CMakeLists.txt
```diff
@@ -1,11 +1,10 @@
 project (sfizz)
 
-if (APPLE)
-	include_directories (SYSTEM /usr/local/opt/jack/include)
-	link_directories (/usr/local/opt/jack/lib)
-endif()
+find_package(PkgConfig REQUIRED)
```
paulfd, line 3: For this one I'll check the compatibility of the JACK part with MSVC later in the WE; maybe we need to disable all this in this case...

src/external/atomic_queue/defs.h
```diff
@@ -6,7 +6,8 @@
 
 #include <atomic>
 
-#if defined(__x86_64__) || defined(_M_X64)
+#if defined(__x86_64__) || defined(_M_X64) || \
+    defined(__i386__) || defined(_M_IX86)
```
paulfd, line 10: As with ARM 32bits I am not sure whether this restriction was intentional or just untested in the original `atomic_queue`, for ARM it seemed to work ok...

paulfd:
Thanks for the PR. I'll delay merging until I can test the MSVC windows build later this WE :slightly_smiling_face:

clients/CMakeLists.txt
jpcima: JACK support is not important in Windows. I disable it.
I think, it would be good to default JACK to OFF on windows.

In fact, another possibility is to check for JACK_FOUND regardless of platform, and enable depending on this condition.

src/external/atomic_queue/defs.h
jpcima: The intrinsic _mm_pause is part of SSE2. I didn't see a reason why it should be disabled.
It's also why to enable "-msse2" in gcc flags, as it's not part of defaults on 32bit.

I wonder, if one wants a generic `spin_loop_pause` for all platforms, can we fallback to OS syscalls? (cf. `pthread_yield`, `sched_yield`) Then, I can enable more platforms to be compatible, like powerpc.

paulfd: There's also (https://en.cppreference.com/w/cpp/thread/yield) but I don't know exactly how all of these differ. Most conversations I've seen tend to suggest not to assume too much about the behavior of these calls across platforms.

jpcima: When you use syscall, rather than instructions, you are hoping that the OS will handle this in a RT-compliant manner, which may or may not be, and it's a high chance that this call does this.

Also I find we can do as in boost, which implements for various cpus:
https://www.boost.org/doc/libs/1_64_0/boost/fiber/detail/cpu_relax.hpp

paulfd: Heh the boost solution looks nice. This "wait 0us" is a bit strange though, this looks like an assumption on how the scheduler will handle the particular case of a zero-duration sleep. At this point using `std::this_thread::yield()` directly has the correct semantic at least.

jpcima: Ok, I agree totally, I think to add it as a following PR, not this.
This one is already loaded as it is, + it has the 32bit which is not strictly a thing of MinGW.

paulfd:
Okay I forgot how the cmake files were included, there's no issue to do it this way, sorry. Merged, thanks !

jpcima:
other thing regarding MinGW: all problems noted in the top post were fixed in abseil-cpp master branch
Do you think it's acceptable to bump the submodule from lts to master ?

PR #35, opened 2020-01-31
Compatibility with CMake 3.5

jpcima:
I edit the build system such that Linux builds are possible with lower CMake versions.
Tried myself on the version 3.5.0.

It's important because without such compability, some Debian-based music distros can't package this. (kx studio, librazik, ...)

If the effective CMake is < 3.9, IPO is skipped and disabled.

Fix
- setting `ENABLE_LTO` to OFF doesn't override the cache variable, unless it's with option `FORCE`
- it was forgotten to check `USE_LIBCPP`
- for the libc++ case, too bad, `target_link_options` will fail on old CMake


paulfd:
Seems good thanks!

PR #36, opened 2020-02-01
Other patches to make it build on old systems and ARM

jpcima:
It should be clear enough for most.

Regarding ARM, some older targets don't have `yield`, and and some v6 it's present as an equivalent of `nop`.


paulfd:
Discussed on chat, this looks good!

PR #38, opened 2020-02-02
Travis build with MinGW (better)

jpcima:
- added environment.sh, a global script for such things as `PATH` across scripts
- cross compiler and libs in docker container
- added a helper to run commands under container, or machine, depending on build type

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,11 @@ jobs:
     - CONTAINER=cross
 
   - os:   linux
+    env:
+    - CROSS_COMPILE=mingw64
+    - CONTAINER=cross
+
+  - os:   linux
     arch: amd64
     dist: bionic
     addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ env:
 jobs:
   include:
   - os:   linux
+    env:
+    - CROSS_COMPILE=mingw32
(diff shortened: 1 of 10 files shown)


paulfd:
For me it looks OK but I'm not really that much of a Travis expert ! Side note though: can we use a similar idea to build the statically linked versions for Linux using docker?

redtide:
It looks OK to me, I was not sure about one fail to find a file but it seems an authorization issue because the fork, so I think we can merge.
@paulfd yes, sure.

redtide (approved):
OK for me.

paulfd:
For the change in ghc, is the warning bad? I'd rather either push this upstream or just ignore the warning "manually", it may cause problems for when we update the header from upstream.

jpcima:
I took this liberty because:
- it doesn't affect the build result; updating would just make the warning come back
- it's a warning from a header, so it propagates its occurrence into many compile units
- we can submit this to upstream as PR, in the meantime
- it's a result of -Wextra, as result of a warning which Windows dynamic loading forces on us and we don't really have any way to work around.

There is some information about this particular warning:
https://debarshiray.wordpress.com/2019/04/01/about-wextra-and-wcast-function-type/

paulfd:
Okay let's propose this upstream and merge meanwhile then! Thanks!

PR #40, opened 2020-02-04
Parse all parameters in opcodes

paulfd:
In preparation for the filters and more complex opcodes, you need not only the store the "back" parameter but also other numbers in the opcode name (e.g. eqN_gainccXX). This PR separates the back parameters from the rest since its position does convey meaning, and assumes that the other parameters are ordered within the opcode.

One side effect of the implementation is that it will properly parse say `eq_NgainccXX`. No a priori check is made that the opcodes is of "known form with respect to parameter parsing". If this is desirable, it would require changes to the way the parsing of opcodes is done (e.g. by checking existence and form upon construction).

### DIFF
--- a/src/sfizz/Opcode.cpp
+++ b/src/sfizz/Opcode.cpp
@@ -6,21 +6,37 @@
 
 #include "Opcode.h"
 #include "StringViewHelpers.h"
-
+#include "absl/strings/charconv.h"
+#include <cctype>
 sfz::Opcode::Opcode(absl::string_view inputOpcode, absl::string_view inputValue)
     : opcode(inputOpcode)
     , value(inputValue)
 {
-    const auto lastCharIndex = inputOpcode.find_last_not_of("1234567890");
-    if (lastCharIndex != inputOpcode.npos) {
-        int returnedValue;
-        absl::string_view parameterView = inputOpcode;
-        parameterView.remove_prefix(lastCharIndex + 1);
-        if (absl::SimpleAtoi(parameterView, &returnedValue)) {
-            parameter = returnedValue;
-            opcode.remove_suffix(opcode.size() - lastCharIndex - 1);
-        }
-    }
     trimInPlace(value);
     trimInPlace(opcode);
+    size_t firstCharIndex { 0 };
+    auto firstNumIndex = opcode.find_first_of("1234567890");
+    while (firstNumIndex != opcode.npos) {
+        lettersOnlyHash = hash(opcode.substr(firstCharIndex, firstNumIndex - firstCharIndex), lettersOnlyHash);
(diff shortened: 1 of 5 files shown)


jpcima:
I thought also that we can extract these numbers into some named variables, which they are a number of (say `int X,Y,Z,N`)
Then, maybe usage of them will be more straight forward?

On examination of the opcode set thought, I see the naming scheme to be a mess sometimes.
For instance, there exist sometimes CC numbers refered to as N, sometimes as X in other contexts.

paulfd:
Yes we could put multiple with names for sure. I saw at most 3 but the naming scheme is indeed a big mess. We really hit the limit of designing your own mark up language for this...

jpcima (approved):

redtide (approved):

paulfd:
In the end there's at most 3 parameters now: a last one, a first one, and possibly a middle one, so I put functions to return optionals with these on top of the full parameter list for direct access.

I also kept a check for the CC-type "back" parameter at the end of the opcode name, but I also store the parameter positions within the opcode stripped of digits. The idea is to integrate at some point a check on these positions, e.g. using a dictionary of correct positions through something like https://github.com/mapbox/eternal
```cpp
constexpr hashmap<uin64_t, std::vector<int>>correctPositions {
 { hash("fil_gain"), { 3 } },
{ hash("fil_gaincc"), { 3, 10 } },
...
};
```
and when parsing the opcodes
```cpp
...
case hash("fil_gaincc"):
    if (opcode.parameterPositions != correctPositions[opcode.lettersOnlyHash])
        return false; // unknown opcode: the parameters are not where they should be
...
```

jpcima:
Yes, it's a good idea for correctness to check positions also.
I wonder about the constexpr map thing and its purpose.

As I saw, opcodes are identified by FNV hash.
Why not generate a minimal perfect hash of them?

Then it can store their accompanying metadata into some ordinary array.
As for the switch, the compiler can be able to generate a jump table, instead of binary-tree dispatch. No?

(just an idea like this, I think perfect hash is typical for this kind of use, in the same way as mnemonic lists of asm instructions for example)

paulfd:
You're probably right, the requirements is mostly that it should be quick to lookup given the hash we already compute. By minimal perfect hash you mean "wrapping" the hashes into a small table we can directly index?

Such a table could also serve to store default values and proper ranges, unlike what I do now which is creating a bunch of constants...

jpcima:
By minimal perfect hash, it means to compress it an 0-N range the smallest possible, where the 0-N may have some gaps.
I think gperf to be software which generates such things. (with various degrees of success, but it's a thing to try)

paulfd:
Indeed, gperf seems like the perfect tool for this use case, no pun intended. I put it up in the todolist.

PR #41, opened 2020-02-04
Parse EQ and filter parameters in the Regions

paulfd:
Another piece in hooking up the filters and EQs in the processing. Makes all parameters available except those related to channel aftertouch.

### DIFF
--- a/src/sfizz/Defaults.h
+++ b/src/sfizz/Defaults.h
@@ -125,22 +125,29 @@ namespace Default
 	constexpr Range<float> rtDecayRange { 0.0f, 200.0f };
 
     // Performance parameters: Filters
+    constexpr float numFilters { 2 };
     constexpr float filterCutoff { 0 };
     constexpr float filterResonance { 0 };
+    constexpr float filterGain { 0 };
     constexpr int filterKeytrack { 0 };
     constexpr uint8_t filterKeycenter { 60 };
     constexpr int filterRandom { 60 };
     constexpr int filterVeltrack { 0 };
     constexpr int filterCutoffCC { 0 };
     constexpr float filterResonanceCC { 0 };
-    constexpr Range<float> filterCutoffRange { 0.0f, 96000.0f };
+    constexpr float filterGainCC { 0 };
+    constexpr Range<float> filterCutoffRange { 0.0f, 20000.0f };
     constexpr Range<int> filterCutoffModRange { -9600, 9600 };
+    constexpr Range<float> filterGainRange { -96.0f, 24.0f };
+    constexpr Range<float> filterGainModRange { -96.0f, 24.0f };
     constexpr Range<int> filterKeytrackRange { 0, 1200 };
     constexpr Range<int> filterRandomRange { 0, 9600 };
     constexpr Range<int> filterVeltrackRange { -9600, 9600 };
-    constexpr Range<float> filterResonanceRange { 0, 40 };
+    constexpr Range<float> filterResonanceRange { 0.0f, 40.0f };
+    constexpr Range<float> filterResonanceModRange { 0.0f, 40.0f };
 
     // Performance parameters: EQ
+    constexpr float numEQs { 3 };
     constexpr float eqBandwidth { 1.0f };
     constexpr float eqBandwidthCC { 0.0f };
     constexpr float eqFrequencyUnset { 0.0f };
--- a/src/sfizz/Defaults.h
+++ b/src/sfizz/Defaults.h
@@ -138,13 +138,13 @@ namespace Default
     constexpr float filterGainCC { 0 };
     constexpr Range<float> filterCutoffRange { 0.0f, 20000.0f };
     constexpr Range<int> filterCutoffModRange { -9600, 9600 };
-    constexpr Range<float> filterGainRange { -96.0f, 24.0f };
-    constexpr Range<float> filterGainModRange { -96.0f, 24.0f };
+    constexpr Range<float> filterGainRange { -96.0f, 96.0f };
+    constexpr Range<float> filterGainModRange { -96.0f, 96.0f };
     constexpr Range<int> filterKeytrackRange { 0, 1200 };
     constexpr Range<int> filterRandomRange { 0, 9600 };
     constexpr Range<int> filterVeltrackRange { -9600, 9600 };
-    constexpr Range<float> filterResonanceRange { 0.0f, 40.0f };
-    constexpr Range<float> filterResonanceModRange { 0.0f, 40.0f };
+    constexpr Range<float> filterResonanceRange { 0.0f, 96.0f };
+    constexpr Range<float> filterResonanceModRange { 0.0f, 96.0f };
 
     // Performance parameters: EQ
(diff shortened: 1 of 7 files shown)


src/sfizz/Defaults.h
```diff
@@ -124,6 +124,41 @@ namespace Default
...
+    // Performance parameters: Filters
+    constexpr float filterCutoff { 0 };
+    constexpr float filterResonance { 0 };
+    constexpr int filterKeytrack { 0 };
+    constexpr uint8_t filterKeycenter { 60 };
+    constexpr int filterRandom { 60 };
+    constexpr int filterVeltrack { 0 };
+    constexpr int filterCutoffCC { 0 };
+    constexpr float filterResonanceCC { 0 };
+    constexpr Range<float> filterCutoffRange { 0.0f, 96000.0f };
```
jpcima: Filter's maximum at 96 kHz will not do.
It should be capped at Fs/2 maximum if possible, even a small bit lower.
With a minimum Fs at 44100 Hz, it seems not unreasonable to set it at 20kHz.

src/sfizz/Region.cpp
```diff
@@ -381,6 +393,226 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
...
+                case hash("bpf_2p_sv"): filters[filterIndex].type = FilterType::kFilterBpf2pSv; break;
+                case hash("brf_2p_sv"): filters[filterIndex].type = FilterType::kFilterBrf2pSv; break;
+                case hash("lpf_4p"): filters[filterIndex].type = FilterType::kFilterLpf4p; break;
+                case hash("hpf_4p"): filters[filterIndex].type = FilterType::kFilterHpf4p; break;
+                case hash("lpf_6p"): filters[filterIndex].type = FilterType::kFilterLpf6p; break;
+                case hash("hpf_6p"): filters[filterIndex].type = FilterType::kFilterHpf6p; break;
+                case hash("pink"): filters[filterIndex].type = FilterType::kFilterPink; break;
+                case hash("lsh"): filters[filterIndex].type = FilterType::kFilterLsh; break;
+                case hash("hsh"): filters[filterIndex].type = FilterType::kFilterHsh; break;
+                case hash("peq"): filters[filterIndex].type = FilterType::kFilterPeq; break;
```
jpcima, line 537: I think `peq` to mean the same as `pkf_2p`, maybe we can alias the 2 of them.
No certainty, given that aria doesn't implement.

A thing I know : `peq` is for Peaking EQ, probably not Parametric EQ that is indicated in docs

Whatever is in aria currently is matching to Peaking EQ from RBJ biquad formulas.

src/sfizz/Region.cpp
```diff
@@ -381,6 +393,226 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
...
+                default:
+                     filters[filterIndex].type = FilterType::kFilterNone;
+                    DBG("Unknown filter type: " << std::string(opcode.value));
+            }
+        }
+        break;
+    // Performance parameters: EQ
```
jpcima, line 545: I don't see `fil_gain` handled, opcode serves as level parameter to lsh/hsh/peq.

src/sfizz/Defaults.h
```diff
@@ -124,6 +124,41 @@ namespace Default
...
+    constexpr Range<int> filterCutoffModRange { -9600, 9600 };
+    constexpr Range<int> filterKeytrackRange { 0, 1200 };
+    constexpr Range<int> filterRandomRange { 0, 9600 };
+    constexpr Range<int> filterVeltrackRange { -9600, 9600 };
+    constexpr Range<float> filterResonanceRange { 0, 40 };
```
jpcima: Not that it matters a lot, but here literals want to have ".0f" to be consistent with the rest of style.

src/sfizz/Region.cpp
```diff
@@ -381,6 +393,226 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
         setValueFromOpcode(opcode, rtDecay, Default::rtDecayRange);
         break;
 
+    // Performance parameters: filters
+    case hash("cutoff"):
+        {
+            const auto filterIndex { backParameter.value_or(1) - 1 };
+            if (!extendIfNecessary(filters, filterIndex + 1))
```
jpcima: Can we these vectors be reserved with some reasonable initial capacity?

src/sfizz/Defaults.h
paulfd: In the end I think this parameter is more the real frequency anyway so 20 kHz is a much saner choice.

src/sfizz/Region.cpp
paulfd: Hm, yes, but then it would mean all regions with no EQ/filter would allocate space for them. The reasonable choices would probably be 2 for filters and 3 for EQs, since these are the SFZ v1 defaults.

On possibility would be to reserve upon first seeing a fitler/eq opcode, so as to have only 1 allocation. I updated doing this.

src/sfizz/Defaults.h
```diff
@@ -124,6 +124,48 @@ namespace Default
...
+    constexpr uint8_t filterKeycenter { 60 };
+    constexpr int filterRandom { 60 };
+    constexpr int filterVeltrack { 0 };
+    constexpr int filterCutoffCC { 0 };
+    constexpr float filterResonanceCC { 0 };
+    constexpr float filterGainCC { 0 };
+    constexpr Range<float> filterCutoffRange { 0.0f, 20000.0f };
+    constexpr Range<int> filterCutoffModRange { -9600, 9600 };
+    constexpr Range<float> filterGainRange { -96.0f, 24.0f };
+    constexpr Range<float> filterGainModRange { -96.0f, 24.0f };
```
jpcima: This "peak/shelf gain" designates the level where the peak, or shelf, is located with relation to 0dB.
Not a proper a "gain" as such, just a level.

As such -96,+24 dB is not appropriate as range.
Possibly, peq is not expected to be used with a negative level, which makes it effectively a notch filter.
If negative levels are kept, at least they should be made symmetric around the 0dB. (eg ±24)

It's the same for shelfs, except instead of boosting a band, it will attenuate it in negatives.

As example. a "-20dB lowshelf"
Capture du 2020-02-05 20-43-16

vs a "+20dB" one
Capture du 2020-02-05 20-43-55

jpcima: I will add ; +- 24dB is restrictive compared to what aria allows.
Based on some testing, it permits at least +- 80 dB of range.

src/sfizz/Defaults.h
```diff
@@ -124,6 +124,48 @@ namespace Default
...
+    constexpr Range<int> filterKeytrackRange { 0, 1200 };
+    constexpr Range<int> filterRandomRange { 0, 9600 };
+    constexpr Range<int> filterVeltrackRange { -9600, 9600 };
+    constexpr Range<float> filterResonanceRange { 0.0f, 40.0f };
```
jpcima: Aria also has a higher upper bound in resonance (of at least 80)

jpcima (approved):

PR #42, opened 2020-02-05
Add the equalizer band filter

jpcima:
This adds the EQ filter. No wrapper at this point, but just here for reference.

In summary:
- the equalizer band is a `peq`
- `eqN_freq` is equivalent of `cutoff`
- `eqN_gain` is equivalent of `fil_gain`
- `eqN_bw` is without equivalent ; it computes a fitting `Q` according to the formula

The formula is "FYI: The relationship between bandwidth and Q is ­— digital filter with BLT"
from the document https://www.w3.org/2011/audio/audio-eq-cookbook.html

After a few tries, it seems matching to ARIA.

Note: the `sinh` part of this formula, which is a k-rate update, is a potential candidate for tabulating


paulfd:
Thanks for this, I'll let you merge!

jpcima:
There is a class FilterEq available for the SFZv1 eq_* control style.

paulfd (approved):

PR #43, opened 2020-02-06
Add the LFO

jpcima:
This adds LFO.
Most functionality is present, except:
- lfoN_count
- modulation of frequency

Some implementation notes in the header.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,6 +28,7 @@ set (SFIZZ_SOURCES
     sfizz/effects/Rectify.cpp
     sfizz/effects/Gain.cpp
     sfizz/effects/Width.cpp
+    sfizz/LFO.cpp
 )
 include (SfizzSIMDSourceFiles)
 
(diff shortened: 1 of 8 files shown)


jpcima:
Note: try this pseudo-sine LFO, 6 flops to implement, vectorizable

http://www.musicdsp.org/en/latest/Synthesis/167-another-cheap-sinusoidal-lfo.html

```
template <> inline float LFO::eval<LFO::kSine>(float phase)
{
    float x = phase + phase - 1;
    return 4 * x * (1 - std::fabs(x));
}
```

jpcima:
Regarding an implementation of `processModulated`, if I understand correctly, following properties should be modulable:

- lfoN_delay
- lfoN_fade
- lfoN_freq
- lfoN_phase
- lfoN_wave
- lfoN_smooth

I think, there is also `lfoN_smooth` missing in current implementation.

paulfd:
This is rebased now. I'll add the wrapping code in the region after the next point release.

PR #44, opened 2020-02-07
Filter benchmarks

paulfd:
This adds a couple benchmarks to:

1. Test the "modulation time step", meaning the number of samples to process before updating the filter coefficients from a purely computation efficiency point of view. I compare also with a simple straightforward "virtual analog" one pole implementation that I did earlier, for my own education if anything.
2. Test whether we should do 2x mono filters vs a stereo filter, once again purely from a computatione efficiency point of view. This has an impact on the complexity of the code since we need to change the processing logic from mono samples to stereo samples and the current filters have a templated number of channels.

This is mostly for supporting further decisions and has no direct impact on the actual library. None of these evaluate the "signal processing quality" of the underlying algorithms or whether large periods in updating the parameters leads to artifacts or instability in the filters.

The names of the benchmark elements should be mostly clear, the number at the end represents the period at which we recompute filter parameters. I will post the perf results on the platforms I have access to.

PR #45, opened 2020-02-07
Filters bis

jpcima:
Filter interface without use of templates.
Mono and stereo are instanciated within a single object. (cf. setChannels)

Add the filter demo program as Jack-Qt client.
It's to check everything to be OK after changes.

Also, the demo program supports the `FilterEq`.

### DIFF
--- a/src/sfizz/SfzFilter.cpp
+++ b/src/sfizz/SfzFilter.cpp
@@ -8,8 +8,8 @@
 #include "SfzFilter.h"
 #include "SfzFilterImpls.cxx"
 #include <cstring>
-#include <cassert>
 #include "SIMDHelpers.h"
+#include "Debug.h"
 
 namespace sfz {
 
@@ -160,7 +160,7 @@ unsigned Filter::channels() const
 
 void Filter::setChannels(unsigned channels)
 {
-    assert(channels < Impl::maxChannels);
+    ASSERT(channels < Impl::maxChannels);
     if (P->fChannels != channels) {
         P->fChannels = channels;
         clear();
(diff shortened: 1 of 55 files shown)


src/sfizz/SfzFilter.cpp
```diff
@@ -16,317 +16,335 @@ namespace sfz {
...
-    case kFilterHpf2pSv: P->processModulated(P->fDspHpf2pSv, in, out, cutoff, q, pksh, nframes); break;
-    case kFilterBpf2pSv: P->processModulated(P->fDspBpf2pSv, in, out, cutoff, q, pksh, nframes); break;
-    case kFilterBrf2pSv: P->processModulated(P->fDspBrf2pSv, in, out, cutoff, q, pksh, nframes); break;
-    case kFilterLsh: P->processModulated(P->fDspLsh, in, out, cutoff, q, pksh, nframes); break;
-    case kFilterHsh: P->processModulated(P->fDspHsh, in, out, cutoff, q, pksh, nframes); break;
-    case kFilterPeq: P->processModulated(P->fDspPeq, in, out, cutoff, q, pksh, nframes); break;
-    default: assert(false);
+void Filter::setChannels(unsigned channels)
+{
+    assert(channels < Impl::maxChannels);
```
paulfd: If you want you can use the uppercase ASSERT so that the check appears only on debug builds.

paulfd (approved):

paulfd:
Cf #44, we may only need to use stereo filters in the end. Thanks!

jpcima:
I'll merge now to put in place new API.
It needs to remove the template parameters in the benchmarks, use `setChannels` instead.
I'll update it.

PR #46, opened 2020-02-08
Filter benchmarks

jpcima:
This is the filter benchmarks #44, rebased on the updated filters API.
I added a few things things on top.

Side note: should we not enable `ScopedFTZ` for all benchmarks ?


paulfd:
Yes indeed we should !

paulfd:
I added some things to make it OK with libc++/STL, as well as FTZ. In the course of this I also made an intermediate target for the SIMD helpers which speeds up compilation of the benchmarks.

jpcima:
Ok let's merge it, then I'll reintegrate part of the changes I made in another PR.
(eg. regarding libsamplerate)

PR #51, opened 2020-02-09
Simplify benchmarks CMakeLists.txt and find libsamplerate

jpcima:


paulfd:
It's may not be used but we could link `bm simd` and `bm_ftz` in all benchmarks, it would remove a sizable number of lines from the CMakeLists

paulfd (approved):

Issue #50, opened 2020-02-09
Request: no configure-time detection


jpcima:
I wish to eliminate the detection of features such as `HAVE_INTRIN_H`.

Usually, in gcc at least, there already exist some predefined macros which permit to detect CPU features without resorting to such checks.

If configure-time checks are gone, sfizz is more easily included from systems which do not use cmake.
For instance, DPF uses only GNU make, and I would like to provide some make module that plugin makers include to embed it easier into the plugin.

Also, in the part below, it's not advisable to override a user's flags with a fixed architecture.
It disables the ability of creating portable ARM packages.
(moreover I don't think -march=native and -mtune or -mfpu together to make sense)
```
    add_compile_options (-DHAVE_ARM_NEON_H)
    add_compile_options (-mfpu=neon)
    add_compile_options (-march=native)
    add_compile_options (-mtune=cortex-a53)
```

paulfd:
Fine for me :+1:

PR #52, opened 2020-02-09
Detect SIMD availability from compiler

jpcima:
#50

Remove configure-time checks for SIMD features, and replace with compile-time ones.

Add SIMDConfig.h where detection magic happens.

Add all 3 source files to all builds inconditionnally, skip the parts not needed with use of preprocessor conditionals.
`    sfizz/SIMDSSE.cpp    sfizz/SIMDNEON.cpp    sfizz/SIMDDummy.cpp`


paulfd:
Actually all the stuff from https://github.com/sfztools/sfizz/blob/efccb7ae446145376a59016781ece093aebb2a47/src/sfizz/Config.h#L56 could go into `SIMDConfig.h` and be per architecture. That was kinda the goal at first.

paulfd:
BTW remove the ARM compiler flags too, I copied them from a raspberry forum but they brought me only pain and sadness...

jpcima:
At this point, it is nearly the file `SfizzSIMDSourceFilesCheck.cmake` which can go altogether.

paulfd:
Indeed it's not very useful anymore !

paulfd (approved):

PR #53, opened 2020-02-09
Initial filter processing

paulfd:
The processing seems to happen, now we have to debug :slightly_smiling_face: 
- Since we're still working on the envelopes and such, the filter modulation is "once per block" for now so not the smoothest around. On the other hand it conforts me into the new event handling where the midi state will be the center reference for all events within the block.
- ~~I went with a full stereo processing, which reduces to something I find quite clear in the way it's handled within the voice. Regions store the descriptions of filters, when a voice starts it borrows the needed number of filters and initialize them based on the description, and all filters get called in series in the rendering callback.~~

Ref #30

### DIFF
--- a/src/sfizz/EQPool.cpp
+++ b/src/sfizz/EQPool.cpp
@@ -8,7 +8,7 @@ using namespace std::chrono_literals;
 sfz::EQHolder::EQHolder(const MidiState& state)
 :midiState(state)
 {
-    eq.setChannels(2);
+
 }
 
 void sfz::EQHolder::reset()
@@ -16,9 +16,10 @@ void sfz::EQHolder::reset()
     eq.clear();
 }
 
-void sfz::EQHolder::setup(const EQDescription& description, uint8_t velocity)
+void sfz::EQHolder::setup(const EQDescription& description, unsigned numChannels, uint8_t velocity)
 {
     reset();
+    eq.setChannels(numChannels);
     this->description = &description;
     const auto normalizedVelocity = normalizeVelocity(velocity);
 
@@ -80,7 +81,7 @@ sfz::EQPool::EQPool(const MidiState& state, int numEQs)
     setnumEQs(numEQs);
 }
 
-sfz::EQHolderPtr sfz::EQPool::getEQ(const EQDescription& description, uint8_t velocity)
+sfz::EQHolderPtr sfz::EQPool::getEQ(const EQDescription& description, unsigned numChannels, uint8_t velocity)
 {
     AtomicGuard guard { givingOutEQs };
     if (!canGiveOutEQs)
@@ -93,7 +94,7 @@ sfz::EQHolderPtr sfz::EQPool::getEQ(const EQDescription& description, uint8_t ve
     if (eq == eqs.end())
         return {};
 
-    (**eq).setup(description, velocity);
+    (**eq).setup(description, numChannels, velocity);
     return *eq;
 }
 
--- a/src/sfizz/EQPool.cpp
+++ b/src/sfizz/EQPool.cpp
@@ -1,0 +1,137 @@
+#include "EQPool.h"
+#include "AtomicGuard.h"
+#include <thread>
+#include "absl/algorithm/container.h"
+#include "SIMDHelpers.h"
+using namespace std::chrono_literals;
+
+sfz::EQHolder::EQHolder(const MidiState& state)
+:midiState(state)
+{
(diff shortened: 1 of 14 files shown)


src/sfizz/EQPool.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+sfz::EQPool::EQPool(const MidiState& state, int numEQs)
+: midiState(state)
+{
+    setnumEQs(numEQs);
+}
+
+sfz::EQHolderPtr sfz::EQPool::getEQ(const EQDescription& description, uint8_t velocity)
+{
+    AtomicGuard guard { givingOutEQs };
```
jpcima, line 86: I wonder, what is the purpose of `AtomicGuard`?
does it provide more value that attempting try-lock on a regular mutex?

paulfd: In the way I'd like it to work, not much, but I wasn't sure that locking the mutex if the try-lock succeeds would be realtime-safe on all platforms. Apparently linux uses futexes which should be (?) but I don't know how it all works inside and if it's also the case for MacOS and Windows. Atomics looked a bit safer, although I'm not super fond of the usage of these classes.

If you have better ways I'm actually interested to know more :)

jpcima: No very much detail on it, it's one of many obscure aspects of low level audio programming.
Fwiw, I think it to be generally accepted in the common audio programming wisdom.

Some articles with mentions
http://www.rossbencina.com/code/real-time-audio-programming-101-time-waits-for-nothing
http://atastypixel.com/blog/four-common-mistakes-in-audio-development/

You can find it also in Ardour at various places.
https://github.com/Ardour/ardour/search?q=try_lock&unscoped_q=try_lock

paulfd: If it's good enough for Ardour, it's probably good enough for me, I'll make a todo of this to try it out at some point.

jpcima:
I was previously thinking of a model to process correctly the modulation.

Suppose:
- modulator: EG, LFO, CC, Var, ...
- modulation sink: Filter cutoff, EG volume, ...

A **modulator** is identified by {type, number}.
A **modulation sink** is identified by {opcode}, and it may connected to a **modulation source**.
A **modulation source** is a list of **modulators** which modulate a target together.

A modulation sink needs be computed once per cycle; a modulator may be reused several times, because it may be applied to multiple sinks.

Hence, I think a cache of modulators, that starts fresh every cycle, and can be queried for a given modulator output, computed lazily as it is requested.

`Span<float> modCache.query({type, number})`
→ exists: return it
→ otherwise: compute a modulator cycle and cache it. (includes smoothcc processing)

The cache reserves as many Buffers as there exist modulators to process, a number known once having parsed the SFZ file.

Example: FM on LFO
```
modSource ← lfo.getFrequencyModulationSource()

freqMod ← {0,...,0}
for modKey ∈ modSource
    freqMod += modCache.query(modKey)

freqMod *= rangeCents

for i ∈ [0,numFrames]
    freq ← lfo.baseFreq * cents2ratio(freqMod[i])
    /* LFO */
```

paulfd:
Yes I agree completely, but CCs and LFOs are slightly different in that CCs are shared between voices, whereas LFOs are only within a voice as far as I understood. 

For the filters/EQs I added more of a "pull" model from the `MidiState` now for this. I have quite alot of work-in-progress to finish now that the filters are in for this, but the approach I wanted for CC is exactly what you described. The `MidiState` is the modulation source, it will memoize its CCs envelopes before or on the first query, and modulation targets have a reference to it. Targets are responsible for querying their needed envelopes from the source.

For EGs and LFOs I guess you could to the same within the voice, where you would not need a semi-global like the `MidiState` to synchronize every voice.

I need to switch all existing modulation to the pull model. Also it should be much cleaner to have multiple CCs modulating targets like this, whereas now it is limited to 1.

paulfd:
BTW I still think `_smoothcc` is per-target, not per-source, so the target has to do and save the filtering.

jpcima:
Yes you're correct about this part, however _before_ conversion to cents,
(as suggested by the experiment on cakewalk regarding the pitchbend)

paulfd:
In the end since the grafting code is quite short, 56e0212 is a proposal to actually differentiate mono and stereo processing. After thinking more about it, the gains of stereo-only processing are on systems that are already fast and do not really care about tiny increases in performance (cf #44).

PR #56, opened 2020-02-09
Add pan width and position using the new helpers

paulfd:
Corrected some bugs in the modulation of width and position in the course of the implementation.


src/sfizz/Voice.cpp
```diff
@@ -63,19 +63,19 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, uint8_t value
     basePan = normalizeNegativePercents(region->pan);
     auto pan { basePan };
     if (region->panCC)
-        pan += normalizeCC(midiState.getCCValue(region->panCC->first)) * normalizeNegativePercents(region->panCC->second);
+        pan += normalizeCC(midiState.getCCValue(region->panCC->first)) * region->panCC->second / 100.0f;
```
jpcima: division is slower, better to multiply in cases of a constant.
In gcc, `fast-math` will resolve this to multiplication, but maybe not so on MS where I believe the fast-math equivalent is not available.

jpcima:
I plotted mid and side coeffs, resulting in this data. This is correct?

Capture du 2020-02-10 00-26-09

Logarithmic

Capture du 2020-02-10 00-30-43

paulfd:
Nope, s***. There it should be better, I forgot the pan table has pi/2 and not pi/4. It should be between ~0.707  and 1 for the mid and ~-0.707 and .707 for the side.

jpcima:
New: linear and log

Capture du 2020-02-10 00-52-04
Capture du 2020-02-10 00-52-19

paulfd:
OK I hope I did not screw things up with the git thingy. Should be OK apparently?

src/sfizz/Voice.cpp
```diff
@@ -364,33 +354,28 @@ void sfz::Voice::processStereo(AudioSpan<float> buffer) noexcept
     copy<float>(rightBuffer, span1);
     add<float>(leftBuffer, rightBuffer);
     subtract<float>(span1, leftBuffer);
-    applyGain<float>(sqrtTwoInv<float>, leftBuffer);
-    applyGain<float>(sqrtTwoInv<float>, rightBuffer);
+
+    // Add const aliases to be slightly more readable
+    const auto midBuffer = leftBuffer;
+    const auto sideBuffer = rightBuffer;
```
jpcima: shouldn't these 2 be defined as sum and difference of L/R channels?
mid=L+R
side=L-R

paulfd: So the mixing happened a couple lines above but it may have been reversed. I corrected to try to be a bit clearer, does it make sense to you now?

jpcima:
It's incorrect. pan=0 width=100 should reproduce the signal identically.

Sine 220Hz on the left, 880Hz on the right.

Capture du 2020-02-10 16-27-22

paulfd:
There's a problem with the mixing constants it seems, the residual sines ~15-20db spaced?

jpcima:
Here, the width formula rewritten.
[stereo.dsp.gz](https://github.com/sfztools/sfizz/files/4182072/stereo.dsp.gz)

It's just written out of personal intuition, not off the paper.
It does exactly what the opcode says: mono at 0, identical at +1, reversed at -1

jpcima (approved):

PR #58, opened 2020-02-12
Fix Sfizz crash in Ardour when no file is loaded

AlexMitchellMus:
stop sfizz from crashing host (ardour) if no SFZ file is loaded and project is reloaded

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -708,6 +708,9 @@ fs::file_time_type sfz::Synth::checkModificationTime()
 {
     auto returnedTime = modificationTime;
     for (auto file: getIncludedFiles()) {
+        if (file.empty()) {
+            return fs::file_time_type::min();
+            }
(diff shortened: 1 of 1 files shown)


jpcima:
oh, probablty it raises an exception from ghc::filesystem.
You can inihibit the exception by using the function signature which accepts `error_code`.

I suggest to try something like this.
```
std::error_code ec;
const auto fileTime = fs::last_write_time(file, ec);
if (ec)
    return fs::file_time_type::min();
```

AlexMitchellMus:
If you can do the commit I'm fine with that.

jpcima:
I reproduced the problem, and amended the commit as c6cc205.

Now a non-existing pathname can't be entered into `last_write_time`.
Yet, there may be cases where `fs` can result in raised exceptions, and it needs to be reviewed.

For instance, a file may exist and not be accessible for reading; or there may be a race condition between when file a checked and its access, if the file was removed meanwhile.

It's best to strenghten the program against all such cases.

PR #60, opened 2020-02-13
Add floating point helpers, and the fast ilog2

jpcima:
This adds some utility functions which permits to work with floating point at bits level.

The purpose is to implement fast logarithm base 2 with integer result.

I think of using this helper for implementing more efficient oscillator wavetables later.
Here is an article from a series on this topic.
https://www.earlevel.com/main/2012/05/09/a-wavetable-oscillator-part-3/

The motivation is because the variable-rate resampler is not really fit for big changes of pitch that would be necessary for a single wavetable. As such, I think to go with the multisamples as in the article, and moreover this makes lerp reasonable using this method.

When pitch is changed under this method, there is potentially a need to jump between multiple tables to lookup from. Tables are searched by pitch ; it's in this case where log2 in just a couple instructions becomes useful.

The variable-rate resampler will be reserved for non-oscillator cases.

Includes test and benchmark (here a speedup around factor 10)

### DIFF
--- a/benchmarks/BM_mathfuns.cpp
+++ b/benchmarks/BM_mathfuns.cpp
@@ -22,6 +22,7 @@ public:
         std::uniform_real_distribution<float> dist { 0.1f, 1.0f };
         source = std::vector<float>(state.range(0));
         result = std::vector<float>(state.range(0));
+        intResult = std::vector<int>(state.range(0));
         std::generate(source.begin(), source.end(), [&]() { return dist(gen); });
     }
 
@@ -31,6 +32,7 @@ public:
 
     std::vector<float> source;
     std::vector<float> result;
+    std::vector<int> intResult;
 };
 
 BENCHMARK_DEFINE_F(MyFixture, Dummy)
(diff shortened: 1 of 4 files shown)


benchmarks/BM_mathfuns.cpp
```diff
@@ -133,6 +135,29 @@ BENCHMARK_DEFINE_F(MyFixture, SIMDCos)
...
 
+BENCHMARK_DEFINE_F(MyFixture, ScalarLibmFloorLog2)
+(benchmark::State& state)
+{
+    for (auto _ : state) {
+        for (size_t i = 0, n = source.size(); i < n; ++i) {
+            intResult[i] = static_cast<int>(
+                std::floor(std::log2(std::fabs(source[i]))));
+        }
+        benchmark::DoNotOptimize(intResult);
```
paulfd, line 146: Do you need this? If it's declared in the fixture I think it will assume there are side effects. I could be wrong though.

paulfd:
Neat trick! Thanks. So there isn't a `nextpow2` type function in the C++ standard library?

paulfd (approved):

benchmarks/BM_mathfuns.cpp
jpcima: Honestly, no idea. If you ask me why it's there, I copied the method of doing from other test cases above.

jpcima:
There isn't `nextpow2` in standard C++, you would have to add it if you need.

benchmarks/BM_mathfuns.cpp
paulfd: Ok so it's because of past-me then!

Issue #55, opened 2020-02-09
Generator wavetables


jpcima:
This is the wavetable generator with IFFT for saw, square, triangle.
It's based on BSD-licensed kissfft.
https://gist.github.com/jpcima/9e43f08ed3a39e0363115c1ed449740e

In sfizz, it will need to implement adequate resampling so these waves can be generated with good quality.

jpcima:
EDIT changed size*2+1, a benign mistake I copied everywhere, to size/2+1
also made the triangle wave start upwards instead of downwards

About this I don't know, didn't check against real oscillator, but I remember the LFO triangle was starting going up.

**EDIT** checked against ARIA, all waves starting upwards, fixed the square also

jpcima:
It's implemented in the PR #61 
For a short summary of how this thing works:

- there exists a class `HarmonicProfile` which is a spectrum generator for any kind of periodic wave.
You invoke this with successive indices (DFT bin), this returns the amplitude and phase as complex number. It starts at index 1 for fundamental `f0`, then `2*f0`, `3*f0`, up to any size.
The spectrums of conventional waves extend to infinity, that we are able to truncate at whichever point we want.

- it can `generate` into an array of any size, by filling a spectrum of the matching size and doing IFFT.
When parameter `cutoff` is set at 0.5, the whole harmonic range is generated up to Nyquist, or <0.5 to cut it sooner.

- there exist multisamples, which wrap this to generate sets of wavetables for different frequency ranges (currently 1 per octave). For each table, this adjusts `cutoff` such that, when the table is repitched to the range's highest frequency, the scaled harmonics do not exceed Nyquist frequency.

- one can plot a multisamples with `sfizz_test_wavetables` and option `-m`. Pictured below.
When all is well, the final range of 10-20kHz reduces the signal to a single sine wave.
![plot](https://user-images.githubusercontent.com/17614485/74530670-eb1d2d00-4f2a-11ea-8b28-16ea4be351b0.png)

- the demo program permits to listen to an oscillator generated in realtime with a frequency sweep.
When frequency modulation causes a jump from a wavetable to another, this creates a discontinuity, but I don't find it very perceptible.
A thing of note: in Cakewalk's oscillators, there exists a quality setting called `Dual-High`, which permits to interpolate from 2 adjacent wavetables of a multisample, and avoids discontinuity created by such jumps.

- regarding custom oscillators from sound files, it's not too hard to modify the system to add it.

[...]

PR #61, opened 2020-02-14
Add the wavetable oscillator

jpcima:
This is the wavetable oscillator using mipmaps and linear interpolation.

Includes the previous fast ilog2 PR, adds KissFFT (BSD license).

It generates the waves for the waveforms of SFZ with inverse FFT. #55
The usage is quite simple and shown in the demo program.

A series of articles explain the method well.
- https://www.earlevel.com/main/2012/05/04/a-wavetable-oscillator-part-1/
- https://www.earlevel.com/main/2012/05/08/a-wavetable-oscillator%e2%80%94part-2/
- https://www.earlevel.com/main/2012/05/09/a-wavetable-oscillator-part-3/


paulfd:
Thanks! Just a note as this will need to be rebased using the ilog function. If that's OK with you, let's keep this around until we merge and debug the filters. I'll tag it along with the CC envelopes.

jpcima:
>Just a note as this will need to be rebased using the ilog function. If that's OK with you, let's keep this around until we merge and debug the filters.

I'm not sure of what you mean by this, the branch is already based on ilog.
(never mind the force-push just happened, I was confused and I did a wrong manip)

On this wavetable stuff, I will add the mipmapping code later that accepts sound input from files.
I don't add to this PR which is already big enough as it is.

paulfd:
I'm still seeing the floating point helpers in the changed files, it's a bit strange. That was the reason behind this comment.

jpcima:
because the #60 is not yet merged.

paulfd:
I need some sleep 😪 Sorry.

jpcima:
Below is a sfz sample, of which you can uncomment any generator you want.
Aria has "*tri" and "*triangle" as synonym. (from what I can find, it's the only one in that case)

(Aria also mentions a sample=*oscillator whose purpose is not known yet)

```
<region>
lokey=0
hikey=127
sample=*sine
//sample=*saw
//sample=*tri
//sample=*triangle
//sample=*square
```

src/sfizz/Wavetables.h
```diff
@@ -0,0 +1,177 @@
...
+   Multisample of a wavetable, which is a collection of FFT-filtered mipmaps
+   adapted for various playback frequencies.
+ */
+class WavetableMulti {
+public:
+    // number of tables in the multisample
+    unsigned tableSize() const { return _tableSize; }
+
+    // number of tables in the multisample
+    static constexpr unsigned multiSize() { return WavetableRange::countOctaves; }
```
paulfd: The name is not super clear to me. Maybe `octaveCounts` or `countOctaves` or `numTables`?

src/sfizz/Wavetables.h
```diff
@@ -0,0 +1,177 @@
...
+    //     octave 9: 10240 Hz - 20480 Hz
+};
+
+/**
+   Multisample of a wavetable, which is a collection of FFT-filtered mipmaps
+   adapted for various playback frequencies.
+ */
+class WavetableMulti {
+public:
+    // number of tables in the multisample
```
paulfd, line 131: I guess this is number of samples in each table?

src/sfizz/Wavetables.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    void processModulated(const float* frequencies, float* output, unsigned nframes);
+
+private:
+    /**
+       Interpolate a value from a part of table, with delta in 0 to 1 excluded.
+       There are `TableExtra` elements available for reading.
+       (cf. WavetableMulti)
+     */
+    static float interpolate(const float* x, float delta);
```
paulfd, line 56: This one could be a good candidate for MathHelpers? If only to have it public for testing. Or, since it's quite simple, maybe a free function in the `.cpp` file?

src/sfizz/Wavetables.h
```diff
@@ -0,0 +1,177 @@
...
+    void fillExtra();
+
+    // length of each individual table of the multisample
+    unsigned _tableSize = 0;
+
+    // number X of extra elements, for safe interpolations up to X-th order.
+    static constexpr unsigned _tableExtra = 4;
+
+    // internal storage, having `multiSize` rows and `tableSize` columns.
+    std::unique_ptr<float[]> _multiData;
```
paulfd: Should we use the Buffer here too, just in case?

paulfd:
I pushed some changes on the way the wavetables are added and handled in the resource. Can you check if it's OK/revert/amend? Thanks.

paulfd:
The rest would be more easily tested through sfz rendering, not sure if there are more unit-type tests that could be super relevant here...

src/sfizz/Wavetables.h
jpcima: `numTables` sounds fine

src/sfizz/Wavetables.h
jpcima: Yes exactly, it's an error in the comment.

src/sfizz/Wavetables.h
jpcima: This a function where you can introduce an interpolation function of choice if you'd like to test others.

Since it has `tableExtra=4`, you could do interpolations to the 4th order.
In fact `tableExtra` can be set to 1 if it is kept linear only.

(linear can be considered sufficient for wavetables, they are all lowpassed in advance)

src/sfizz/Wavetables.h
jpcima: No problem yes

jpcima:
>I pushed some changes on the way the wavetables are added and handled in the resource. Can you check if it's OK/revert/amend? Thanks.

The goal was to keep them shared among instances to avoid consuming extra memory.
This proposed change would undo that, is that correct?

paulfd:
Yes it would, although that's not the intended effect. My main idea was to hide the tables within a pool so as not to clutter the `Resources` structure. I'm a bit shaky on the `static` semantics because I know there are weird cases. If we declare the variables as `static const`, possibly further initializing them right within the `Wavetables.h` file rather than in a constructor, would this be fine? This effectively makes them a pure singleton class..

jpcima:
Starting with C++11 the static local variables have thread-safe inits.
This should not be ever a problem.
Of course if that disturbs you, you're free change them into actual globals.

src/sfizz/Wavetables.h
paulfd: Ok since it kinda depends on the table structure let's keep it here for now.

paulfd (approved):

paulfd:
If it's good for you I'll merge this.

PR #62, opened 2020-02-14
Optimize random

jpcima:
There isn't need for quality PRNG for musical uses.
The linear-congruential formula will be faster: https://godbolt.org/z/2dKbeR


paulfd (approved):

paulfd:
Yeah probably, nothing security critical here 🙂

jpcima:
It ran in the LFO branch well, so I assume it fine.

I did another bizarre observation although, uniform_real_distribution seems slow on clang compiler for some reasons. I'll check later to reproduce it in compiler explorer.

As this branch is concerned, good for me.

PR #64, opened 2020-02-16
Filter smoothing

paulfd:
It appears that the spaceship sounds comes from the initial smoothing from 0 to whatever parameters the filter starts on. This is a transient thing, there's no issue once the smoothing stops which is also why you can only really see it on spectrograms/envelopes.

There are I guess different ways to handle the problem:

- In this PR, I went with a smaller transition period, which implies going in Faust from `si.smoo` for smoothing to `si.smooth(ba.tau2pole(desired value for tau in seconds))`. The parameters appear to take between 5 and 10 `tau` to stabilize, the theoretical value being around 7 (@jpcima feel free to correct).
- The "best" solution would be to only smooth after the first initialization of the parameters, but apparently it is complicated right now for Faust-generated code at least. We could do this by dry-running the filters for about `5 * tau` on initialization but there's a substantial cost.
- Another possibility would be to not smooth within Faust at all and smooth the parameter changes within sfizz. We'll have to do this in a couple other places anyway, but it might be better for filter stability to smooth the biquad's parameters rather than the input parameters of the filters.

As said, this PR takes the first approach which might be the best compromise in the short term.

### DIFF
--- a/src/sfizz/dsp/filters/filters_modulable.dsp
+++ b/src/sfizz/dsp/filters/filters_modulable.dsp
@@ -66,7 +66,8 @@ ap1Smooth(s,f) = fi.iir((a,1.),(a)) with {
 //------------------------------------------------------------------------------
 // Example
 //------------------------------------------------------------------------------
-// process = rbjLpfSmooth(si.smoo, cutoff, 0.0, resonance : ba.db2linear) with {
-//   cutoff = hslider("[1] Cutoff [unit:Hz] [scale:log]", 440.0, 50.0, 10000.0, 1.0);
-//   resonance = hslider("[2] Resonance [unit:dB]", 0.0, 0.0, 40.0, 0.1);
-// };
+process = rbjPeakingEqSmooth(si.smooth(ba.tau2pole(0.001)), cutoff, gain, resonance) with {
+  cutoff = hslider("[1] Cutoff [unit:Hz] [scale:log]", 50.0, 50.0, 10000.0, 1.0);
+    resonance = hslider("[2] Resonance [unit:dB]", 1.0, 0.0, 40.0, 0.1);
+	gain = hslider("[2] gain [unit:dB]", 1.0, -80.0, 80.0, 0.1);
+ };
--- a/src/sfizz/dsp/filters/filters_modulable.dsp
+++ b/src/sfizz/dsp/filters/filters_modulable.dsp
@@ -62,12 +62,3 @@ hp1Smooth(s,f) = fi.iir((0.5*(1.+p),-0.5*(1+p)),(0.-p)) with {
 ap1Smooth(s,f) = fi.iir((a,1.),(a)) with {
   a = (-1.+2.*ma.PI*f/ma.SR) : s;
 };
-
-//------------------------------------------------------------------------------
-// Example
-//------------------------------------------------------------------------------
-process = rbjPeakingEqSmooth(si.smooth(ba.tau2pole(0.001)), cutoff, gain, resonance) with {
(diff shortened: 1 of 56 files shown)


src/sfizz/dsp/filters/filters_modulable.dsp
```diff
@@ -66,7 +66,8 @@ ap1Smooth(s,f) = fi.iir((a,1.),(a)) with {
 //------------------------------------------------------------------------------
 // Example
 //------------------------------------------------------------------------------
-// process = rbjLpfSmooth(si.smoo, cutoff, 0.0, resonance : ba.db2linear) with {
-//   cutoff = hslider("[1] Cutoff [unit:Hz] [scale:log]", 440.0, 50.0, 10000.0, 1.0);
-//   resonance = hslider("[2] Resonance [unit:dB]", 0.0, 0.0, 40.0, 0.1);
-// };
+process = rbjPeakingEqSmooth(si.smooth(ba.tau2pole(0.001)), cutoff, gain, resonance) with {
```
jpcima: Ah ok, so it was about this code. You can consider this part obsolete entirely, and remove, it was for early tests.

Beside it has an error: RBJ formulas accepts a gain specification in dB, and Q specification in linear.
From SFZ, it inputs both of them in dB. It means it was incorrect to remove `db2linear` on the parameter `resonance`.

You can generate a filter by a simpler way.
Pass the function name as `-pn`. (make it use another entry function instead of `process`)
```
faust2jack -double -pn sfzPeq src/sfizz/dsp/filters/sfz_filters.dsp
```

jpcima:
I'm good with setting tau at much lower.
I think to update the filters_demo such that I enable it to modulate F and Q by LFO, in order to put the stability to test a little more.

paulfd:
I'm OK with it too, but knowing how low we can go can be nice for sure.

jpcima:
Using the filters demo + LFO, instability is rare but some cases can be found.

A reproducible one:
- type: Brf1p
- cutoff: 500 Hz
- resonance: 0 dB
- cutoff mod speed: 1Hz
- cutoff mod range: 24 semitones

from here, try increasing the cutoff; it will destabilize nearing 5 kHz and up.

paulfd:
24 semitones is a factor of 4, should we consider this a regular use case?

jpcima:
I've seen ±2400 in various places of sfz docs, so I made it that.
(on the other hand, I know ARIA to not limit the range at all in various cases)

The thing of factor 4 and ~5kHz is we are approaching/exceeding nyquist frequency.
It validates the idea that some filters are not resistant to `Fc>=k*Fs/2`.
That makes it a good idea to clamp `Fc`, and pick a `k` at 0.9 or what we'll find to be appropriate.

Then it's too check in practice how low `tau` can be set, perhaps below ms will be fine.

jpcima:
In my testing, it's the Apf1 and Brf1 which are especially sensitive (still failing), and also state-variable ones. (Brf1 is based on 2 Apf1s cascaded)

There exist 2 models of Ap1f that I know of. They have a little different phase responses.
1. https://ccrma.stanford.edu/~jos/pasp/First_Order_Allpass_Interpolation.html
2. https://ccrma.stanford.edu/realsimple/DelayVar/Phasing_First_Order_Allpass_Filters.html

Currently `ap1Smooth` is implemented as the first, but didn't have better luck stabilizing using the second.
Below is an implementation of the second:
```
ap1Smooth(s,f) = fi.iir((a,-1.),(0.-a)) with {
  t = tan(2.*ma.PI*f/ma.SR);
  a = ((1.-t)/(1.+t)) : s;
};
```

paulfd:
Merging this then, we'll handle possible instabilities if they appear in (reasonable) use cases! Thanks!

Issue #63, opened 2020-02-15
One-shot voices stick and fill up the polyphony


jpcima:
reported by @alcomposer : when you play 64 notes of a one-shot generator, the polyphony is filled and you can't play anymore unless resetting the plugin.

Should the voices be cleared once the envelope has reached the end?
also can we permit new notes to steal a slot more aggressively, in case the polyphony is filled?

```
<group>
lovel=0
hivel=127

<region>
trigger=release
sample=*noise
loop_mode=one_shot
ampeg_attack=1
ampeg_decay=1
ampeg_release=0
ampeg_sustain=0
```

AlexMitchellMus:
`trigger=release` is what stops the voice from being deactivated. Otherwise the example works fine without that line.

paulfd:
OK, so we can assume that in this case (`trigger=release`) the envelope should basically run its course without sustaining at all, would this be right?

If so, the envelope needs a `freeRunning` parameter in the envelope and get into release state as soon as it hits the sustain state. As a note for me (I may have time to do this in the WE) or if someone wants to try, this should be as simple as having a boolean check to switch the status to `Release` [here](https://github.com/sfztools/sfizz/blob/cf940980b45cd8b0f80c7b03daf6eb845e3ee2ab/src/sfizz/ADSREnvelope.cpp#L154) and [here](https://github.com/sfztools/sfizz/blob/cf940980b45cd8b0f80c7b03daf6eb845e3ee2ab/src/sfizz/ADSREnvelope.cpp#L83), along with proper setup depending on the `Region` trigger value and setting the `freeRunning` value [there](https://github.com/sfztools/sfizz/blob/cf940980b45cd8b0f80c7b03daf6eb845e3ee2ab/src/sfizz/Voice.cpp#L112).

However, this is a good opportunity to refactor [that](https://github.com/sfztools/sfizz/blob/cf940980b45cd8b0f80c7b03daf6eb845e3ee2ab/src/sfizz/ADSREnvelope.cpp#L15) and the `prepareEGEnvelope` function above which are ugly and will not scale to flex EGs easily. The solution would be to have it more like [this function](https://github.com/sfztools/sfizz/blob/cf940980b45cd8b0f80c7b03daf6eb845e3ee2ab/src/sfizz/FilterPool.cpp#L20) in the filter pool: give the EG a reference to the `Region`, the `MidiState`, as well as the initial delay and velocity, and let it compute its parameters itself.

[...]

PR #66, opened 2020-02-17
Corrected #63

paulfd:
- The envelope does not sustain if the region has `trigger=release`
- Change the envelope setup to pass the region rather than
individual envelope parameters

### DIFF
--- a/benchmarks/BM_ADSR.cpp
+++ b/benchmarks/BM_ADSR.cpp
@@ -11,45 +11,61 @@
 #include <random>
 #include <numeric>
 
-constexpr int fixedAmount { 12 };
-constexpr int envelopeSize { 2 << 16 };
-constexpr int attack = static_cast<int>(envelopeSize / 4) - fixedAmount;
-constexpr int decay = attack;
-constexpr int release = attack;
-constexpr int releaseTime = envelopeSize - static_cast<int>(envelopeSize / 4);
+constexpr int fixedAmount{12};
+constexpr float sampleRate{100.0f};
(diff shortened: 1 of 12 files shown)


jpcima:
This makes sense to me.

What about more aggressive voice stealing, would it be good to do something about it?
(eg. one note arrives as polyphony is full, kill oldest voice and replace it)

jpcima (approved):

paulfd:
I have a mechanism now in place. The idea is to take all voices that are in release state and take the voice with the lowest score -- and still below a threshold.
Reading it I realize I'm sorting the wrong array, and that there's a sort of mixup of logic there so I'll fix it :P

This mechanism was never active because the voice was never in a release state. Saying this I also realize that the above fix will still not put the voice in release state, so I'll hold off merging for now.

paulfd:
Ah no I'm sorting by the longest running voice that can be stolen and then finding one that has power below a threshold. See [here](https://github.com/sfztools/sfizz/blob/8941f0a8004d2af54966a245e7ad13a84ed87097/src/sfizz/Synth.cpp#L340).

Maybe just taking the one with the lowest power would be simpler actually. Even if it ran for a shorter time, we don't really care.

jpcima:
I would say to elect the voice which has the greatest age, first among the releasing ones, and second among the others.

paulfd:
Ok now both the voice and the envelope should be in release state together.

paulfd:
Benchmarks updated now, so I'll merge this as soon as the CI ends.

PR #70, opened 2020-02-19
Add method to set up initial filter coeffs

jpcima:
This adds a method `prepare` into the filter and eq wrappers.
It should set initial coeffs of the filter without being affected by smoothing.

The trick is to use an external boolean fed from C++ to Faust code.
When the boolean is on, it forces the pole to 0, hence disable smoothing.
When it runs a single dummy cycle of the filter, it is forced to take the immediate value.

By this method, the spaceship sound can be reliably eliminated.

(generated files not included in commit, don't forget to remake them)

### DIFF
--- a/src/sfizz/SfzFilter.cpp
+++ b/src/sfizz/SfzFilter.cpp
@@ -105,6 +105,29 @@ void Filter::clear()
         dsp->instanceClear();
 }
 
+void Filter::prepare(float cutoff, float q, float pksh)
+{
+    sfzFilterDsp *dsp = P->getDsp(P->fChannels, P->fType);
+
+    if (!dsp)
+        return;
+
+    // compute a dummy 1-frame cycle with smoothing off
+
+    float buffer[Impl::maxChannels] = {0};
+    float *inout[Impl::maxChannels];
+    bool en = dsp->isSmoothingEnabled();
+
+    for (unsigned i = 0; i < Impl::maxChannels; ++i)
+        inout[i] = &buffer[i];
+
+    dsp->instanceClear();
+    dsp->configureStandard(cutoff, q, pksh);
+    dsp->setSmoothingEnabled(false);
+    dsp->compute(1, inout, inout);
+    dsp->setSmoothingEnabled(en);
+}
+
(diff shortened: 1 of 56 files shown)


jpcima:
Added just some edits to make code shorter and let it inline better.
Not that it matters a lot in this case, but done anyway.

src/sfizz/SfzFilter.cpp
```diff
@@ -274,6 +297,29 @@ void FilterEq::clear()
...
+        return;
+
+    // compute a dummy 1-frame cycle with smoothing off
+
+    float buffer[Impl::maxChannels] = {0};
+    float *inout[Impl::maxChannels];
+    bool en = dsp->isSmoothingEnabled();
+
+    for (unsigned i = 0; i < Impl::maxChannels; ++i)
+        inout[i] = &buffer[i];
```
paulfd, line 314: I've been thinking about a way to make this whole idiom of constructing arrays of float* to pass to filters a bit simpler. It's a bit tedious and verbose, although I'm sure the compiler crushes it in release. We'd need some type that would decay to `float**` or `const float* const *`, but `{ ... }` don't fit the bill.

paulfd (approved):

paulfd:
I added the code to use it in the filters, with a way to avoid duplicate code in the process. May not be perfect, we'll see how it integrates with the proper modulation matrix.

I tested it, if it's OK for you you can merge it directly. Do you want to regenerate the filters in a separate commit?

jpcima:
>I tested it, if it's OK for you you can merge it directly. Do you want to regenerate the filters in a separate commit?

I note we have different versions of faust compilers, which makes it best if one of us only generates this files, to avoid needless diff verbosity.

paulfd:
Good point, maybe we could use a Docker image set at a specific faust version to build the filters?

jpcima:
No let's just agree on a common version and use that.

paulfd:
I'm using the one in the ubuntu 19.10 repos, I will check which it is, but it's not a stable release so I don't think it's a great choice.

paulfd:
19.10 is faust 2.15, the next LTS is going to be 2.20.2 and I think arch is there too now, so we can keep this one. I regenerated them with this version.

PR #72, opened 2020-02-20
Enable static builds without vcpkg for linux platforms and strip symbols from the LV2 plugin

paulfd:
- Using `SFIZZ_STATIC_LIBSNDFILE` the linker will use the static pkgconfig. This option is meant to be used in containers and CI for distributing the LV2 binary. You need to compile the `libsndfile` libraries using `-fPIC` manually, disable the generation of shared versions of the libraries, and provide proper pkg-config files.
- On Linux's linkers, the `-Wl,--version-script...` linker option allows to specify the global symbols. Using it with a simple `lv2.version` file allows to only show the `lv2_descriptor()` symbol in the LV2 plugin file, and not stray symbols from other libraries. This is required because even if you compile all the dependent libraries with `-fvisibility=hidden`, for some reason with GCC some symbols are still left in the end, whereas clang hides them properly. Regardless, with this version script the `-fvisibility` flag is not needed for the dependent libraries.


paulfd:
Example script for a container using `apt-get source`
```Dockerfile
FROM ubuntu:18.04
RUN sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list
RUN apt-get update
RUN apt-get install -y build-essential wget pkg-config automake gettext libtool-bin git libjack-jackd2-dev cmake

WORKDIR /tmp
RUN apt-get source libogg-dev libvorbis-dev libflac-dev libsndfile

ENV CFLAGS="-fPIC"
ENV CXXFLAGS="-fPIC"

RUN cd /tmp/libogg-1.3.2 \
    && ./configure --disable-shared --enable-static \
    && make -j$(nproc) \
    && make install

RUN cd /tmp/libvorbis-1.3.5 \
    && ./autogen.sh \
    && ./configure --disable-shared --enable-static \
    && make -j$(nproc) \
    && make install

RUN cd /tmp/flac-1.3.2 \
    && ./autogen.sh \
    && ./configure --disable-shared --enable-static --disable-silent-rules \
    && make -j$(nproc) \
    && make install

RUN cd /tmp/libsndfile-1.0.28 \
    && ./configure --disable-shared --enable-static --disable-full-suite --disable-silent-rules \
    && make -j$(nproc) \
    && make install

RUN mkdir build \
    && cd build \
    && cmake .. -DSFIZZ_JACK=OFF -DSFIZZ_SHARED=OFF -DSFIZZ_STATIC_LIBSNDFILE=ON -DCMAKE_BUILD_TYPE=Release \
    && make -j$(nproc)
```

jpcima:
Maybe check here, regarding the link of static libraries into plugins.
It's a solution I've used that works across all OS.
https://github.com/jpcima/ADLplug/blob/master/cmake/LinkHelpers.cmake

Use `require_symbols` on non-Windows, on Windows use a def file.
DEF file: https://github.com/jpcima/ADLplug/blob/master/sources/link/LV2.def

paulfd:
Not sure the .def file is needed for LV2 though? The plugin seems to load in ardour, and the dependency walker seems to show the `lv2_descriptor()` symbol properly.

image

cmake/SfizzConfig.cmake
```diff
@@ -45,7 +45,11 @@ else()
     find_package(PkgConfig REQUIRED)
     pkg_check_modules(SNDFILE "sndfile" REQUIRED)
     target_include_directories(sfizz-sndfile INTERFACE ${SNDFILE_INCLUDE_DIRS})
-    target_link_libraries(sfizz-sndfile INTERFACE ${SNDFILE_LIBRARIES})
+    if (SFIZZ_STATIC_LIBSNDFILE)
+        target_link_libraries(sfizz-sndfile INTERFACE ${SNDFILE_STATIC_LDFLAGS})
```
jpcima: CMake's documention indicates there should exist a variable `SNDFILE_STATIC_LIBRARIES`.
Perhaps this was meant instead of `LDFLAGS`?

jpcima:
This makes .def optional I suppose.
Maybe it's sufficient that there exist dllexport annotations, so it will stop gcc exporting everything?

paulfd:
Maybe it's worth mentioning that this result is when compiled with MSVC on a windows host. Mingw builds might be different.

jpcima (approved):

cmake/SfizzConfig.cmake
paulfd: Yes, I remember using STATIC_LIBRARIES and switching to LDFLAGS for some reason, but I think it was because I had the libraries in a non-standard location maybe.

Using STATIC_LIBRARIES seems to work now so I switched it back.

redtide (approved):
I suppose if there are no additional checks needs on current state we could merge and eventually correct/add the rest later, for sure I think we need a setup for the lv2 instead using plain zips and test the GH releases deployment.

Issue #75, opened 2020-02-23
Assigning client name via command line


lminiero:
Hi all,

first of all, thanks a lot for working on this! I've recently discovered sfizz thanks to Michael Wellis' excellent Ardour template for VPO, and I started using it for all my SFZ needs, especially considering LinuxSampler was crashing a bit too often on my old laptop.

I was wondering if you could add a command line option to `sfizz_jack` to provide a custom name to use for the Jack client: at the moment it defaults to an hardcoded `sfizz`

    auto defaultName = "sfizz";
    jack_status_t status;
    client = jack_client_open(defaultName, JackNullOption, &status);

which can be problematic when you open multiple clients at the same time, or want to be able to quickly create connections with other applications. This looks like a simple task so I can take care of that as well and prepare a pull request, if you let me know which branch you'd like me to target.

Thanks!

lminiero:
Just prepared this very short diff on the (slightly older) commit of the `develop` branch I'm on, and it seems to be working as expected:

```
diff --git a/clients/jack_client.cpp b/clients/jack_client.cpp
index 12344e2..00387bf 100644
--- a/clients/jack_client.cpp
+++ b/clients/jack_client.cpp
@@ -163,6 +163,7 @@ static void done(int sig [[maybe_unused]])
     // exit(0);
 }
 
+ABSL_FLAG(std::string, client_name, "sfizz", "Jack client name");
 ABSL_FLAG(std::string, oversampling, "1x", "Internal oversampling factor (value values are x1, x2, x4, x8)");
 ABSL_FLAG(uint32_t, preload_size, 8192, "Preloaded value");
 
@@ -171,10 +172,12 @@ int main(int argc, char** argv)
     // std::ios::sync_with_stdio(false);
     auto arguments = absl::ParseCommandLine(argc, argv);
     auto filesToParse = absl::MakeConstSpan(arguments).subspan(1);
+    const std::string clientName = absl::GetFlag(FLAGS_client_name);
     const std::string oversampling = absl::GetFlag(FLAGS_oversampling);
     const uint32_t preload_size = absl::GetFlag(FLAGS_preload_size);
 
     std::cout << "Flags" << '\n';
+    std::cout << "- Client name: " << clientName << '\n';
     std::cout << "- Oversampling: " << oversampling << '\n';
     std::cout << "- Preloaded Size: " << preload_size << '\n';
     const auto factor = [&]() {
@@ -216,9 +219,8 @@ int main(int argc, char** argv)
[...]

PR #76, opened 2020-02-23
New command line option for custom Jack client name (sfizz_jack)

lminiero:
Addresses #75. This adds a new `-client_name` property to choose a custom Jack client name, and the default if nothing is provided is `sfizz`, which keeps the previous behaviour:

    -client_name (Jack client name); default: "sfizz";
    -oversampling (Internal oversampling factor (value values are x1, x2, x4,
      x8)); default: "1x";
    -preload_size (Preloaded value); default: 8192;

### DIFF
--- a/clients/jack_client.cpp
+++ b/clients/jack_client.cpp
@@ -163,6 +163,7 @@ static void done(int sig [[maybe_unused]])
     // exit(0);
 }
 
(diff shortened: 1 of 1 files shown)


paulfd:
Thanks!

PR #81, opened 2020-02-26
Added bend and filter/eq opcodes

paulfd:
I hope I didn't forget too many. These lists are hard to maintain :D


redtide (approved):
Indeed. Looking at that new extension I still wonder about aliases, talking with @jpcima turns out that we should avoid to add those not in v1 and v2 like ARIA does, but what about new ones? (then also a `filN_gain_ccX` and `filN_gain_onccX`).

paulfd:
Yeah it's true that slews of aliases are not making things clear for anyone. For "new" opcodes I guess we should try to make them logical as much as possible. The filter ones are a bit annoying because of `filN_type` and `cutoffN` that apply to the same filter... I would argue for `filN_gainccX`, to match the `eqN_gainccX`.

PR #84, opened 2020-02-27
Effects

jpcima:
Add effects and their routing. Work in progress

- Effect factory

Designed such that it can receive user-defined effects.
An effect is designated by its "type" opcode.

- Routing with main bus and up to 256 effect buses
- Mixing per-region
- Basic "lofi" effect to start which is relatively CW-compliant

**TODO**
- review the mixing behavior in case of use of main/fx buses together, or one of these only

There are still some things and dirty and it's not tried.
Regardless, I submit the code for review if you would like to have a peek.

Remark: I would like if we can introduce HIIR downsampler separately of this PR

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -734,18 +734,18 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
         break;
 
     case hash("effect"): // effect&
-        {
-            const auto effectNumber = opcode.backParameter();
-            if (!effectNumber || *effectNumber < 1 || *effectNumber > config::maxEffectBuses)
-                break;
-            auto value = readOpcode<float>(opcode.value, {0, 100});
-            if (!value)
-                break;
-            if (static_cast<size_t>(*effectNumber + 1) > gainToEffect.size())
-                gainToEffect.resize(*effectNumber + 1);
-            gainToEffect[*effectNumber] = *value / 100;
+    {
+        const auto effectNumber = opcode.backParameter();
+        if (!effectNumber || *effectNumber < 1 || *effectNumber > config::maxEffectBuses)
             break;
-        }
+        auto value = readOpcode<float>(opcode.value, { 0, 100 });
+        if (!value)
+            break;
+        if (static_cast<size_t>(*effectNumber + 1) > gainToEffect.size())
+            gainToEffect.resize(*effectNumber + 1);
+        gainToEffect[*effectNumber] = *value / 100;
+        break;
+    }
 
     // Ignored opcodes
     case hash("hichan"):
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -733,6 +733,20 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
         setCCPairFromOpcode(opcode, amplitudeEG.ccSustain, Default::egOnCCPercentRange);
         break;
 
+    case hash("effect"): // effect&
+        {
+            const auto effectNumber = opcode.backParameter();
+            if (!effectNumber || *effectNumber < 1 || *effectNumber > config::maxEffectBuses)
+                break;
+            auto value = readOpcode<float>(opcode.value, {0, 100});
+            if (!value)
+                break;
+            if (static_cast<size_t>(*effectNumber + 1) > gainToEffect.size())
+                gainToEffect.resize(*effectNumber + 1);
+            gainToEffect[*effectNumber] = *value / 100;
+            break;
+        }
+
     // Ignored opcodes
     case hash("hichan"):
     case hash("lochan"):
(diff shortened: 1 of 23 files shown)


jpcima:
Below some sfz for testing

Bitcrusher on main bus
```
<region>
lokey=0
hikey=127
sample=*sine

<effect>
type=lofi
bitred=90
decim=10
```

Bitcrusher on bus 1, main bus turned off
```
<region>
lokey=0
hikey=127
sample=*sine
effect1=100

<effect>
directtomain=0
fx1tomain=100
type=lofi
bus=fx1
bitred=90
decim=10
```

Bitcrusher on bus 1, main and fx1 mixed 50/50
```
<region>
lokey=0
hikey=127
sample=*sine
effect1=100

<effect>
directtomain=50
fx1tomain=50
type=lofi
bus=fx1
bitred=90
decim=10
```

src/sfizz/effects/Lofi.cpp
```diff
@@ -0,0 +1,213 @@
...
+        {
+            // exponential curve fit
+            float a = 1.289079e+00, b = 1.384141e-01, c = 1.313298e-04;
+            dt = std::pow(a, b * depth) * c - c;
+            dt = fSampleTime / dt;
+        }
+
+        float phase = fPhase;
+        float lastValue = fLastValue;
+        hiir::Downsampler2xFpu<12>& downsampler2x = fDownsampler2x;
```
paulfd: The other ref

src/sfizz/effects/Lofi.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        if (depth == 0) {
+            if (in != out)
+                std::memcpy(out, in, nframes * sizeof(float));
+            clear();
+            return;
+        }
+
+        float lastValue = fLastValue;
+        hiir::Downsampler2xFpu<12>& downsampler2x = fDownsampler2x;
```
paulfd: Is this ref (and the one below) useful somehow?

src/sfizz/effects/Lofi.cpp
```diff
@@ -0,0 +1,213 @@
...
+        for (unsigned c = 0; c < EffectChannels; ++c) {
+            _bitred[c].clear();
+            _decim[c].clear();
+        }
+    }
+
+    void Lofi::process(const float* const inputs[2], float* const outputs[2], unsigned nframes)
+    {
+        for (unsigned c = 0; c < EffectChannels; ++c) {
+            _bitred[c].setDepth(_bitred_depth);
```
paulfd, line 72: For now this is unmodulated but it should be at some point right?

src/sfizz/Region.cpp
```diff
@@ -751,6 +751,20 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
         setCCPairFromOpcode(opcode, amplitudeEG.ccSustain, Default::egOnCCPercentRange);
         break;
 
+    case hash("effect"): // effect&
```
paulfd: When we merge the other one I'll rebase this

src/sfizz/Synth.cpp
```diff
@@ -188,6 +196,74 @@ void sfz::Synth::handleControlOpcodes(const std::vector<Opcode>& members)
...
+
+        // note(jpc): gain opcodes are linear volumes in % units
+
+        case hash("directtomain"):
+            if (auto valueOpt = readOpcode<float>(opcode.value, {0, 100}))
+                getOrCreateBus(0).setGainToMain(*valueOpt / 100);
+            break;
+
+        case hash("fxtomain"): // fx&tomain
+            if (auto numberOpt = opcode.firstParameter()) {
```
paulfd: There too, it'll be cleaner with the new parser

paulfd (approved):

src/sfizz/effects/Lofi.cpp
jpcima: not very useful, it can go, or I don't mind to change to `auto&`

src/sfizz/effects/Lofi.cpp
jpcima: It must have modulation.
Since the focus is the framework of effects, it's limited in implementation.
Let's add this when the modulation matrix is well handled.

paulfd:
All good for me, can merge :) Thanks alot!

jpcima:
On blockSize and sampleRate : I don't really agree with the renaming `setSampleRate`

The entry point `init` is really intended for an initialization role.
- It *should* setup anything which is rate dependent, and clear the state.
- It *should not* modify the values of parameters

I would suggest to pass both of sample rate and buffer size to this `init` method.

For information, I wish this process to be equivalent with a design of future C API which can be used for registering external effects into sfizz.

(in such a way, we can register MDA's effect package externally like Sforzando, or do anything custom)

paulfd:
OK, I'm good with this, I'll update.

PR #88, opened 2020-02-27
Install rules fixed for MinGW

jpcima:
- install the static library
- install the pkgconfig regardless whether shared is build or not
- set staticlib's output name such that it creates "libsfizz.a"
- have staticlib in `lib/` rather than `lib/static/`
- install the dll in `bin/`

That means on minGW we get this layout, which is well according to its standard.
- bin/libsfizz.dll
- lib/libsfizz.dll.a
- lib/libsfizz.a

and we can have a `pkgconfig --static` which works because static and shared are homonymous.

I don't know if these changes are good for MS, it's needed to check it.


paulfd:
You mean installing on windows? Redtide did some innosetup thing for the msvc builds.

jpcima:
Having read appveyor's script, I noticed we don't use the install rules for MS.
(maybe we could?)

It's maybe the case that a hierarchy like `lib/static` would be more appropriate for this compiler.
In MinGW, the compiler knows to distinguish between static and shared in the same directory. (libsfizz.a and libsfizz.dll.a)

paulfd:
I have to admit that I have absolutely no clue how it's supposed to work for Windows. I always assumed people put DLLs in their applications as needed.

Since the license can allow static linking easily, maybe the best way to propose the library for windows users would be to integrate it into vcpkg or conan. The former at least seems to be quite pushed by MS for managing C++ dependencies in their products...

PR #96, opened 2020-03-03
Add curves

jpcima:
This adds a support class for curves.

- [x] parsing `<curve>` opcodes
- [x] explicit and implicit indexing
- [x] linear interpolation
- [x] spline interpolation (not in sfz, can be introduced as extension)


jpcima:
Some test cases for curves (plot)

- Builtin curve 6
    - run: `./build/tests/sfizz_plot_curve 6 curves.sfz > curves.dat`
    - sfz: `/*empty*/`

- Undefined curve 7, fallback to default
    - run: `./build/tests/sfizz_plot_curve 7 curves.sfz > curves.dat`
    - sfz: `/*empty*/`

- Custom curve 7 (not indexed)
    - run: `./build/tests/sfizz_plot_curve 7 curves.sfz > curves.dat`
    - sfz: `<curve> v0=1.0 v64=0.9 v127=0.0`

- Custom curve 7, omitted side points, fallback to 0.0 and 1.0 (not indexed)
    - run: `./build/tests/sfizz_plot_curve 7 curves.sfz > curves.dat`
    - sfz: `<curve> v64=0.9`

- Custom curve 8 (indexed)
    - run: `./build/tests/sfizz_plot_curve 8 curves.sfz > curves.dat`
    - sfz: `<curve> curve_index=8 v0=1.0 v64=0.9 v127=0.0`

- Spline curve 7 (extension)
    - delete relevant `#if 0` in `Curve.cpp`
    - run: `./build/tests/sfizz_plot_curve 7 curves.sfz > curves.dat`
    - sfz: `<curve> sfizz:curve_interpolator=spline v64=0.75`

jpcima:
Rebased on the new parser

src/sfizz/Curve.cpp
```diff
@@ -0,0 +1,250 @@
...
+        slot = &_curves.back();
+    } else {
+        size_t index = static_cast<unsigned>(explicitIndex);
+        if (index >= _curves.size())
+            _curves.resize(index + 1);
+        _useExplicitIndexing = true;
+        slot = &_curves[index];
+    }
+
+    slot->reset(new Curve(curve));
```
paulfd, line 208: Any reason not to just avoid the `slot` variable and call reset on `_curves.back()` or `_curves[index]`?

src/sfizz/Curve.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+Curve Curve::buildPredefinedCurve(int index)
+{
+    Curve curve;
+
+    switch (index) {
+    default:
+        ASSERTFALSE;
+    case 0:
```
paulfd, line 68: Should we break here?

src/sfizz/Curve.cpp
```diff
@@ -0,0 +1,250 @@
...
+namespace sfz
+{
+
+static const Curve defaultCurve = Curve::buildBipolar(0.0, 1.0);
+
+Curve Curve::buildCurveFromHeader(
+    absl::Span<const Opcode> members, Interpolator itp, bool limit)
+{
+    Curve curve;
+    bool fillStatus[NumValues] = {};
```
paulfd, line 23: I know that `std::vector<bool>` has some packing going on to avoid taking up heaps of memory. Would it be important here? Can we expect `bool[]` to have similar optimizations?

src/sfizz/Curve.cpp
jpcima: It shouldn't matter as this case does not happen.
I set the default to the linear curve.

src/sfizz/Curve.cpp
jpcima: On the plus side `vector<bool>` is compact, but access takes more instructions.
Anyway, these temporary computations are not relevant to performance.

src/sfizz/Curve.cpp
jpcima: Because this complies to RAII and avoids repeating a small code fragment in both branches.

jpcima:
What is the bug of `lerpFill` corrected in `e58bf6f`?

paulfd:
Any of the "Custom" test case (I missed your comment when writing them, saw it after :sweat:)
```cpp
TEST_CASE("[Curve] Custom 2")
{
    auto curve = sfz::Curve::buildCurveFromHeader({
        { "v063", "1" }
    }, sfz::Curve::Interpolator::Linear);
    REQUIRE( curve.evalCC7(0) == 0.0f );
    REQUIRE( curve.evalCC7(127) == 1.0f );
    REQUIRE( curve.evalCC7(63) == 1.0f );
    REQUIRE( curve.evalCC7(2) == Approx(0.032).margin(1e-3) );
    REQUIRE( curve.evalCC7(70) == 1.0f );
}
```
At 63 the value obtained is 0.984. I took the opportunity to use the `linearRamp` SIMD thingy although it's not active on x86, but it should be a simple index issue so we can recover the original code if you prefer. 
```
-------------------------------------------------------------------------------
[Curve] Custom 2
-------------------------------------------------------------------------------
/home/paul/source/sfizz-paulfd/tests/CurveT.cpp:119
...............................................................................

/home/paul/source/sfizz-paulfd/tests/CurveT.cpp:126: FAILED:
  REQUIRE( curve.evalCC7(63) == 1.0f )
with expansion:
  0.98437f == 1.0f
```

src/sfizz/Curve.cpp
paulfd: It warns about falling through so I'll silence it.

jpcima:
It overwrited the index 63? oh, my bad.
The current index `iCurr` should have been checked for `fillStatus` =1, and in this case, skip it.

jpcima:
on the `else DBG(...)` thing:

To fix this cleanly in all cases, the macro expansion should be wrapped into the form `do { ... } while (0)`.
There is a discussion: https://stackoverflow.com/questions/257418/do-while-0-what-is-it-good-for

paulfd:
I'll let you revert it and correct it if you want. The loops that fills values are slightly tighter like this but all of this is not so critical :slightly_smiling_face: You can build it against the test suite it should cover most corner cases.

paulfd:
Other than this I'm OK to merge!

jpcima:
No it doesn't interpolate multi points correctly.
Check for instance `<curve> v0=1.0 v64=0.9 v100=0.9 v127=0.0`

jpcima:
Sorry it does work correctly... it's just me who took the wrong plotfile..

paulfd:
It's OK, your punition is another test case :+1:

paulfd (approved):

jpcima:
Phew I thought I was punished to be forced to write it :smile: 
It's good to merge.

PR #101, opened 2020-03-05
Changed the way the parser handles multiple parameters in opcodes

paulfd:
Idea from @jpcima :)

The resulting code is much cleaner and properly handles wrong positioning of parameters, with basically no cost. Also changed the parameters to be `u16` to handle values larger than 256 for the CC number.

### DIFF
--- a/src/sfizz/Opcode.cpp
+++ b/src/sfizz/Opcode.cpp
@@ -24,11 +24,11 @@ sfz::Opcode::Opcode(absl::string_view inputOpcode, absl::string_view inputValue)
         nextCharIndex = opcode.find_first_not_of("1234567890", nextNumIndex);
 
         uint32_t returnedValue;
-        hasBackParameter = (nextCharIndex == opcode.npos);
-        const auto numDigits = hasBackParameter ? opcode.npos : nextCharIndex - nextNumIndex;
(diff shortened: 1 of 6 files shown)


src/sfizz/Opcode.cpp
```diff
@@ -24,11 +24,9 @@ sfz::Opcode::Opcode(absl::string_view inputOpcode, absl::string_view inputValue)
         nextCharIndex = opcode.find_first_not_of("1234567890", nextNumIndex);
 
         uint32_t returnedValue;
-        hasBackParameter = (nextCharIndex == opcode.npos);
-        const auto numDigits = hasBackParameter ? opcode.npos : nextCharIndex - nextNumIndex;
+        const auto numDigits = (nextCharIndex == opcode.npos) ? opcode.npos : nextCharIndex - nextNumIndex;
         if (absl::SimpleAtoi(opcode.substr(nextNumIndex, numDigits), &returnedValue)) {
-            // ASSERT(returnedValue < std::numeric_limits<uint8_t>::max());
-            parameterPositions.push_back(parameterPosition);
+            lettersOnlyHash = hash("&", lettersOnlyHash);
```
jpcima, line 29: The case that the invalid input contains "&" should be handled for safety.
Otherwise you could match a hash which doesn't have the required parameter count.

paulfd: So should we ignore these opcodes or just silently strip out the "&" from the opcode name?

PR #103, opened 2020-03-08
Add a clang-tidy check on CI

paulfd:
If a warning is too aggressive/unneeded, you can deactivate the line by appending `// NOLINT` at the end, or `// NOLINTNEXTLINE` on the line above.


paulfd:
Added in #104

PR #105, opened 2020-03-08
Add the autopan effect

jpcima:
This adds autopan, conformant to cakewalk in functionality.

Adds also a simplified implementation of phase-synced LFO, which is needed to modulate.

Example file
```
<region>
sample=*sine

<effect>
type=apan
apan_waveform=0
apan_freq=1.0
apan_phase=90.0
apan_depth=100
apan_dry=0
apan_wet=100
```

### DIFF
--- a/src/sfizz/effects/Apan.cpp
+++ b/src/sfizz/effects/Apan.cpp
@@ -37,8 +37,8 @@ namespace fx {
 
     void Apan::setSamplesPerBlock(int samplesPerBlock)
     {
-        _lfoOutLeft.reset(new float[samplesPerBlock]);
-        _lfoOutRight.reset(new float[samplesPerBlock]);
+        _lfoOutLeft.resize(samplesPerBlock);
+        _lfoOutRight.resize(samplesPerBlock);
     }
 
     void Apan::clear()
@@ -51,8 +51,8 @@ namespace fx {
         float dry = _dry;
         float wet = _wet;
         float depth = _depth;
-        float* modL = _lfoOutLeft.get();
-        float* modR = _lfoOutRight.get();
+        float* modL = _lfoOutLeft.data();
+        float* modR = _lfoOutRight.data();
 
         computeLfos(modL, modR, nframes);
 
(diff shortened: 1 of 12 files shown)


src/sfizz/effects/Apan.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    void Apan::setSampleRate(double sampleRate)
+    {
+        _samplePeriod = 1.0 / sampleRate;
+    }
+
+    void Apan::setSamplesPerBlock(int samplesPerBlock)
+    {
+        _lfoOutLeft.reset(new float[samplesPerBlock]);
+        _lfoOutRight.reset(new float[samplesPerBlock]);
```
paulfd: Is it ok to use `sfz::Buffer` here? Maybe it can be improved too, but it has memory tracking in debug mode to check if memory is not dangling..

src/sfizz/effects/Apan.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+        return CXX11_MOVE(fx);
+    }
+
+    void Apan::computeLfos(float* left, float* right, unsigned nframes)
+    {
+        switch (_lfoWave) {
+        #define CASE(X) case lfo::X:                          \
+            computeLfos<lfo::X>(left, right, nframes); break;
+        default:
```
paulfd, line 120: triangle by default then?

src/sfizz/effects/Apan.cpp
jpcima: oh absolutely, `Buffer` is fine for here.
I forget about the memory tracking feature. (which is good)

src/sfizz/effects/Apan.cpp
jpcima: Yes

jpcima:
Note: for most of effect ranges, I don't have the info and they are guessed
(except for percent ones)

paulfd:
Yeah I can imagine. That's also why I think that putting everything in the Defaults file is a bit tedious but can be useful if we wish to i) see at a glance the values and ii) possibly generate this file automatically from a spec at some point.

jpcima:
Not only that should be meta, but also the relations between opcodes and their respective `oncc` modulators.

paulfd:
What do you mean?

jpcima:
That there is a method that can get me the modulated buffer of `apan_depth` when there is `oncc` and `curvecc` on the opcode, and there exist knowledge somewhere that these have relation.

Which also may be deduced from the spec document.

PR #108, opened 2020-03-09
Add French translated LV2 strings

trebmuh:
Hope that helps.


jpcima:
It's good but I need it in develop not master. I'll handle it, don't worry about it

PR #112, opened 2020-03-10
Travis build for MOD devices

jpcima:
I add some Travis scripts that run the build for MOD devices.
To do so, I shared some of the docker process with mingw.

Also, I've created a mod-plugin-builder derived docker image. It's based of the moddevices/mod-plugin-builder which is now 2 years old since the last update. It adds
- cmake update in /usr/local (it was 3.4, which is too old)
- a wrapper script in /usr/local/bin/mod-plugin-builder which sets up environment

The derived docker is a bit annoying but I had some trouble to set it up with the base image, so for now that's what it is.

I'm changing the docker to be invoked with `-u <uid>`.
This will fix permission problems that occured with MPB, but also this does not permit root commands. It's the reason for introducing also `buildenv_as_root` that passes `-u 0`, which it needs to install mingw packages.


paulfd (approved):
Ok for me! Let's do the point release first though since it seems things may be broken by all the changes.

jpcima:
>Ok for me! Let's do the point release first though since it seems things may be broken by all the changes.

I wouldn't mind, but first I have to make the build process for VST also.
Should I create yet more build jobs, or make that together with LV2?

I think the latter more reasonable.

paulfd:
Yes it would be a good idea.

paulfd:
I added the C++17 switch and rebased on the gcc 4.9 branch to see if the CI follows !

PR #113, opened 2020-03-10
Fix travis buildenv in prepare_tarball

jpcima:
The goal of this PR is to fix a problem in CI, which is bound to arrive at the moment of deployment.
It's sure to conflict with #112, but I'll fix it later

There exists the function `buildenv` which can prefix a command in order to run it in the container.

When `prepare_tarball.sh` is invoked, and a tag is present, `buildenv` will be called to prefix `make install`. But `buildenv` is only defined when it's docker (mingw).

- I make the docker process more generic to accept the image name by environment (`$CONTAINER`).
- if `$CONTAINER` is set, we can know this target uses docker
- use this information in `prepare_tarball` in order to call it under appropriate `buildenv` (a dummy in case of Linux or Mac)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ jobs:
     stage: "Build"
     env:
     - CROSS_COMPILE=mingw32
-    - CONTAINER=cross
+    - CONTAINER=archlinux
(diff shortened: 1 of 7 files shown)


paulfd (approved):
Ok for me!

PR #116, opened 2020-03-12
Update clang-tidy with VST and filters additions

paulfd:
And remove some linting error along the way. The most would probably be the `~SfizzVstProcessor`. I don't call `setEnable` but only stop the background thread, and within a `try` block to avoid excepting in the destructor. Apparently you cannot call virtual functions in destructors.

I'll rebase on the filter additions when done.

### DIFF
--- a/.travis/download_vst_sdk.sh
+++ b/.travis/download_vst_sdk.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -ex
+
+vst_sdk_archive="vst-sdk_3.6.14_build-24_2019-11-29.zip"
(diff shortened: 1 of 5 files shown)


.travis/download_vst_sdk.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -ex
+
+vst_sdk_archive="vst-sdk_3.6.14_build-24_2019-11-29.zip"
```
jpcima, line 5: This archive is expected to be kept in travis cache. Skip download if it exists.

paulfd: Done :)

jpcima (approved):

PR #122, opened 2020-03-15
Avoid purging and recreating the synth in Reaper

paulfd:
It's less clear in Bitwig because I did not attach the console, but Reaper calls `setEnable(false)` when transport stops, and `setEnable(true)` when it starts. This triggers a destruction and recreation/reloading of the sample files in the current state. I changed it to only adapt the sample rate if needed and start/stop the background worker.

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -579,6 +579,8 @@ void sfz::Synth::noteOn(int delay, int noteNumber, uint8_t velocity) noexcept
     ASSERT(noteNumber < 128);
     ASSERT(noteNumber >= 0);
 
+    DBG("Note on at " << noteNumber << " (" << +velocity << ")");
+
     ScopedTiming logger { dispatchDuration, ScopedTiming::Operation::addToDuration };
     resources.midiState.noteOnEvent(delay, noteNumber, velocity);
 
@@ -594,6 +596,8 @@ void sfz::Synth::noteOff(int delay, int noteNumber, uint8_t velocity) noexcept
     ASSERT(noteNumber < 128);
     ASSERT(noteNumber >= 0);
     UNUSED(velocity);
+
+    DBG("Note off at " << noteNumber << " (" << +velocity << ")");
 
     ScopedTiming logger { dispatchDuration, ScopedTiming::Operation::addToDuration };
     resources.midiState.noteOffEvent(delay, noteNumber, velocity);
@@ -961,4 +965,12 @@ void sfz::Synth::enableLogging() noexcept
 void sfz::Synth::disableLogging() noexcept
 {
     resources.logger.disableLogging();
+}
+
+void sfz::Synth::allSoundOff() noexcept
+{
+    for (auto &voice: voices)
+        voice->reset();
+    for (auto& effectBus: effectBuses)
+        effectBus->clear();
 }
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -965,6 +965,11 @@ void sfz::Synth::disableLogging() noexcept
 
 void sfz::Synth::allSoundOff() noexcept
 {
+    AtomicDisabler callbackDisabler{ canEnterCallback };
+    while (inCallback) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+
     for (auto &voice: voices)
         voice->reset();
     for (auto& effectBus: effectBuses)
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -579,8 +579,6 @@ void sfz::Synth::noteOn(int delay, int noteNumber, uint8_t velocity) noexcept
     ASSERT(noteNumber < 128);
     ASSERT(noteNumber >= 0);
 
-    DBG("Note on at " << noteNumber << " (" << +velocity << ")");
-
     ScopedTiming logger { dispatchDuration, ScopedTiming::Operation::addToDuration };
     resources.midiState.noteOnEvent(delay, noteNumber, velocity);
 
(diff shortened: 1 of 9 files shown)


vst/SfizzVstProcessor.cpp
```diff
@@ -46,6 +46,13 @@ tresult PLUGIN_API SfizzVstProcessor::initialize(FUnknown* context)
 
     _state = SfizzVstState();
 
+    fprintf(stderr, "[sfizz] new synth\n");
+    _synth.reset(new sfz::Sfizz);
+    if (!_synth) {
```
jpcima: `synth` cannot possibly be null after new.

vst/SfizzVstProcessor.cpp
```diff
@@ -105,21 +112,27 @@ tresult PLUGIN_API SfizzVstProcessor::canProcessSampleSize(int32 symbolicSampleS
...
 tresult PLUGIN_API SfizzVstProcessor::setActive(TBool state)
 {
-    stopBackgroundWork();
-    _synth.reset();
+    sfz::Sfizz* synth = _synth.get();
 
-    if (state) {
-        fprintf(stderr, "[Sfizz] new synth\n");
-        sfz::Sfizz* synth = new sfz::Sfizz;
+    if (!synth) {
```
jpcima: Why create the synth here, not just return a kResultFalse?
Does this fix bad host behavior that you observed?

paulfd: No this was more to keep the previous behavior which would create the synth if it doesn't exist---albeit only if state was true. This can go entirely.

vst/SfizzVstProcessor.cpp
paulfd: Right if it fails it's gonna throw.

jpcima:
This passes the checks in Steinberg VST validator.

The synth needs now a state reset, or panic, or whatever to call it.
Shall we add such an API?

paulfd:
Sure! How panicky should we make it? There's a `resetAllControllers()` call in the MidiState, I guess the rest would be to kill all voices and reset them?

jpcima:
Usually, when it's only MIDI, I would think `resetAllControllers` to be a good idea.
But with sfz, we'll be messing actually with parameters, and it's not very good that a DSP modifies its own parameters. It's possible in VST, but LV2 spec is actively against this idea.
(speaking of generalities)


For panic, I would think of the following:
- kill voices
- clear effects
- clear filter histories of modulations (when we make `smoothcc`)
- [...]

paulfd:
- `voice.clear()` should reset the voice OK and free the background loading.
- `EffectBus` has both `clear()` and `clearInputs()`; I think both need to be called.
- The modulation is (sorry) still not there. For now it's in the `Voice` so it will be cleared by the above call.
- Controllers no, but we do need to send `noteOff` events to all regions I guess? And the `MidiState` too since it counts/will count the polyphony.

jpcima:
>clearInputs()

Absolutely unnecessary, this is only to initialize the buffer for mixing the regions into the inputs of buses, and the process function does this by itself.

>Controllers no, but we do need to send noteOff events to all regions I guess?

It's not supposed to pass through note-off, just shutdown the voices directly.
Much like MIDI "all sound off" vs "all notes off".

>And the MidiState too since it counts/will count the polyphony.

Without doubt, yes

paulfd:
It means that for the regions through they could have seen a note on without note off, so the keyswitching (e.g. legato, which is handled at the region level right now) will be off. If you go back in transport and start again it will be wrong. Reaper sends all current notes on transport play so you'll recover the "proper" state if you assumed note-offs before. Come to think about it, it may even send note-offs on transport stop.

jpcima:
On this legato stuff I don't know a lot, so don't mind if I may have some (dumb) questions.
- can we have note killing that happens immediately, but also in keyswitch preserving fashion?
- if the host resends the notes as you observed, doesn't that mean it takes itself the role of restoring note state, and there shouldn't be an issue if we clear it?

paulfd:
Ok so I checked the code again, the legato uses the `MidiState` now which it should. Updating the midi state will ensure the legato behavior.

The note-off is only needed for `sw_down` and `sw_up` correctness, which indicate that a region is active when another key is down or up. This has the following form for information.
```c++
    if (keyswitchRange.containsWithEnd(noteNumber)) {
        if (keyswitchDown && *keyswitchDown == noteNumber)
            keySwitched = false;

        if (keyswitchUp && *keyswitchUp == noteNumber)
            keySwitched = true;
    }
```

I will check the host behavior. It would actually be nice of them to send the note-offs. If I attach to the bitwig bridge thing can I see the console output?

jpcima:
>I will check the host behavior. It would actually be nice of them to send the note-offs. If I attach to the bitwig bridge thing can I see the console output?

By attaching, I've no idea.
I wrote this an an earlier post, one sure method is you replace the plugin bridge with a wrapper script that calls it but redirects outputs to files. Then you obtain plugin output by `tail -f` of that file.

paulfd:
Both of them seem to send note-offs correctly so we should not worry too much about the midi state.

paulfd:
I added the panic to the API, and called it from both the LV2 and VST. Seems to work OK.

src/sfizz/Synth.cpp
```diff
@@ -962,3 +966,11 @@ void sfz::Synth::disableLogging() noexcept
...
     resources.logger.disableLogging();
 }
+
+void sfz::Synth::allSoundOff() noexcept
+{
+    for (auto &voice: voices)
+        voice->reset();
+    for (auto& effectBus: effectBuses)
+        effectBus->clear();
+}
```
jpcima, line 977: Should this method not have the `AtomicDisabler` like others?

paulfd: Good point.

PR #126, opened 2020-03-16
Proper tuning of wavetable generators

jpcima:
The changes allows to take in consideration the pitch ratio.
It fixes the use of `tune`/`pitch` and oscillators together.


paulfd (approved):
👍

PR #127, opened 2020-03-16
Add support of oscillator_phase

jpcima:
This supports oscillator dephasing.

Phase in degrees example
```
<region>
sample=*saw

<region>
sample=*saw
oscillator_phase=90
```

Random phase example
```
<region>
sample=*saw

<region>
sample=*saw
oscillator_phase=-1
```

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -58,6 +58,17 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, uint8_t value
             break;
         }
         waveOscillator.setWavetable(wave);
+
+        float phase;
+        float phaseParam = region->oscillatorPhase;
+        if (phaseParam >= 0) {
+            phase = phaseParam * (1.0f / 360.0f);
+            phase -= static_cast<int>(phase);
+        } else {
+            std::uniform_real_distribution<float> phaseDist { 0.0001f, 0.9999f };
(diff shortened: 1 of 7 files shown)


src/sfizz/Voice.cpp
```diff
@@ -58,6 +58,17 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, uint8_t value
             break;
         }
         waveOscillator.setWavetable(wave);
+
+        float phase;
+        float phaseParam = region->oscillatorPhase;
+        if (phaseParam >= 0) {
+            phase = phaseParam * (1.0f / 360.0f);
+            phase -= static_cast<int>(phase);
```
paulfd, line 66: Could we do this in the region and store the normalized version directly?

src/sfizz/Voice.cpp
```diff
@@ -58,6 +58,17 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, uint8_t value
...
+        } else {
+            std::uniform_real_distribution<float> phaseDist { 0.0001f, 0.9999f };
+            phase = phaseDist(Random::randomGenerator);
```
paulfd, line 69: So the expected behavior is that for negative phase (especially -1) the  phase is random?

jpcima: Yes, this was observed in the software and is confirmed in the details written in the book

PR #128, opened 2020-03-17
Add oscillator=on and wavetables from file

jpcima:
This supports loading wavetables from file.
The file will be loaded directly and mipmapped.

I've added a status enum to `FilePromise` which I hope correct, and adapted.
This lets me know if file loading has failed because of an error of libsndfile.

- Example

```
<region>
sample=ramp.wav
oscillator=on
```

- Demo files

[sinewave.wav.gz](https://github.com/sfztools/sfizz/files/4340849/sinewave.wav.gz)
[ramp.wav.gz](https://github.com/sfztools/sfizz/files/4340848/ramp.wav.gz)
[nimportequoi.wav.gz](https://github.com/sfztools/sfizz/files/4340847/nimportequoi.wav.gz)

### DIFF
--- a/src/sfizz/FilePool.h
+++ b/src/sfizz/FilePool.h
@@ -38,7 +38,6 @@
 #include "Logger.h"
 #include <chrono>
 #include <thread>
-#include <sndfile.hh>
 
 namespace sfz {
 using AudioBufferPtr = std::shared_ptr<AudioBuffer<float>>;
--- a/src/sfizz/FilePool.h
+++ b/src/sfizz/FilePool.h
@@ -55,7 +55,7 @@ struct FilePromise
 {
     AudioSpan<const float> getData()
     {
-        if (dataReady)
+        if (dataStatus == DataStatus::Ready)
             return AudioSpan<const float>(fileData);
         else if (availableFrames > preloadedData->getNumFrames())
             return AudioSpan<const float>(fileData).first(availableFrames);
@@ -69,10 +69,22 @@ struct FilePromise
         preloadedData.reset();
         filename = "";
         availableFrames = 0;
-        dataReady = false;
+        dataStatus = DataStatus::Wait;
         oversamplingFactor = config::defaultOversamplingFactor;
         sampleRate = config::defaultSampleRate;
     }
+
+    void waitCompletion()
+    {
+        while (dataStatus == DataStatus::Wait)
+            std::this_thread::sleep_for(std::chrono::milliseconds(50));
+    }
+
+    enum class DataStatus {
+        Wait = 0,
+        Ready,
+        Error,
+    };
 
     absl::string_view filename {};
     AudioBufferPtr preloadedData {};
@@ -80,7 +92,7 @@ struct FilePromise
     float sampleRate { config::defaultSampleRate };
     Oversampling oversamplingFactor { config::defaultOversamplingFactor };
     std::atomic<size_t> availableFrames { 0 };
-    std::atomic<bool> dataReady { false };
+    std::atomic<DataStatus> dataStatus { DataStatus::Wait };
     std::chrono::time_point<std::chrono::high_resolution_clock> creationTime;
 
     LEAK_DETECTOR(FilePromise);
--- a/src/sfizz/FilePool.h
+++ b/src/sfizz/FilePool.h
@@ -43,11 +43,19 @@ namespace sfz {
 using AudioBufferPtr = std::shared_ptr<AudioBuffer<float>>;
 
 
+struct FileInformation {
+    uint32_t end { Default::sampleEndRange.getEnd() };
+    uint32_t loopBegin { Default::loopRange.getStart() };
+    uint32_t loopEnd { Default::loopRange.getEnd() };
+    double sampleRate { config::defaultSampleRate };
+    int numChannels { 0 };
+};
+
 // Strict C++11 disallows member initialization if aggregate initialization is to be used...
-struct PreloadedFileHandle
+struct FileDataHandle
 {
     std::shared_ptr<AudioBuffer<float>> preloadedData;
-    float sampleRate;
+    FileInformation information;
 };
 
 struct FilePromise
@@ -142,14 +150,6 @@ public:
      */
     size_t getNumPreloadedSamples() const noexcept { return preloadedFiles.size(); }
 
-    struct FileInformation {
-        uint32_t end { Default::sampleEndRange.getEnd() };
-        uint32_t loopBegin { Default::loopRange.getStart() };
-        uint32_t loopEnd { Default::loopRange.getEnd() };
-        double sampleRate { config::defaultSampleRate };
-        int numChannels { 0 };
-    };
-
     /**
      * @brief Get metadata information about a file.
      *
@@ -159,7 +159,7 @@ public:
     absl::optional<FileInformation> getFileInformation(const std::string& filename) noexcept;
 
     /**
-     * @brief Check that a file is preloaded with the proper offset bounds
+     * @brief Preload a file with the proper offset bounds
      *
      * @param filename
      * @param offset the maximum offset to consider for preloading. The total preloaded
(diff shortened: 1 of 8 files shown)


src/sfizz/Wavetables.cpp
```diff
@@ -307,4 +332,65 @@ const WavetableMulti* WavetablePool::getWaveSquare()
...
+{
+    _fileWaves.clear();
+}
+
+const WavetableMulti* WavetablePool::createFileWave(FilePool& filePool, const std::string& filename)
+{
+    if (const WavetableMulti* wave = getFileWave(filename))
+        return wave;
+
+    if (!filePool.preloadFile(filename, 0))
```
paulfd: I'm wondering if we really need the filepool here, since we will read the whole file in memory anyway and preload them upon creation of the synth? You could have `preloadFileWave` to preload the wave oscillator as given by the synth, and `getFileWave` in the voice upon launching. The latter could fail if the voice was not preloaded.

src/sfizz/FilePool.h
```diff
@@ -69,18 +69,30 @@ struct FilePromise
...
-        dataReady = false;
+        dataStatus = DataStatus::Wait;
         oversamplingFactor = config::defaultOversamplingFactor;
         sampleRate = config::defaultSampleRate;
     }
 
+    void waitCompletion()
+    {
+        while (dataStatus == DataStatus::Wait)
+            std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
paulfd, line 87: This might be long for small wave files (although not for compressed files) considering the loading will spinwait on this.

src/sfizz/FilePool.h
```diff
@@ -69,18 +68,30 @@ struct FilePromise
...
+    };
+
     absl::string_view filename {};
     AudioBufferPtr preloadedData {};
     AudioBuffer<float> fileData {};
     float sampleRate { config::defaultSampleRate };
     Oversampling oversamplingFactor { config::defaultOversamplingFactor };
     std::atomic<size_t> availableFrames { 0 };
-    std::atomic<bool> dataReady { false };
+    std::atomic<DataStatus> dataStatus { DataStatus::Wait };
```
paulfd, line 102: I added a check for lock-freeness, but other than this I think we can keep this enum state even if we don't use the Filepool to load wavetables.

src/sfizz/Synth.cpp
```diff
@@ -345,6 +346,9 @@ bool sfz::Synth::loadSfzFile(const fs::path& file)
             if (!resources.filePool.preloadFile(region->sample, maxOffset))
                 removeCurrentRegion();
         }
+        else if (region->oscillator && !region->isGenerator()) {
```
paulfd, line 364: Should we add a "type" enum to Region?

src/sfizz/Wavetables.cpp
jpcima: I had this idea to use the same file loader as used for everything.
This loader can have an concept of caching, and it may interact directly in the future with the idea of global storage.

src/sfizz/FilePool.h
jpcima: The purpose was to have a value here that will be set even in error case, such that I may know when to stop waiting if it failed.

src/sfizz/Synth.cpp
jpcima: Yes, the reason of why I didn't is that I don't know what to call it

src/sfizz/FilePool.h
jpcima: How much?

src/sfizz/Wavetables.cpp
paulfd: I guess the Wavepool is also static now? The Filepool does not store these wavetables, it may store the beginning of files but this should be pretty moot in the end since we're mostly interested in the files and not the table. Most of the Filepool logic is centered on this preloading/streaming/oversampling complexity. Reading a file with libsndfile is maybe 3 lines of code on the other hand. What we may do is not compute a table if the static/global wavepool finds that it already has the proper table.

One thing that we may do is provide an API for file reading but not preloading in the Filepool, possibly with caching?

src/sfizz/FilePool.h
paulfd: Given https://box.ferrand.cc/sfizz-reports/2020-03-27-floating-midi.html from an SSD it can probably be set to 1ms.

src/sfizz/FilePool.h
paulfd: Yes it's good, no problem. The check is there to ensure no hidden mutex, although enum class probably has the same atomic properties as integers.

src/sfizz/Synth.cpp
paulfd: Right... `source` with `sample`, `generator`, `wavetable` maybe is OK? or `dataSource`?

jpcima: Maybe designate `oscillator=on` and `sample=*something` as both oscillators.
And rename `region->oscillator` as something else.
(eg. `runAsOscillator`)

It reminds me: in the Cakewalk book, it's said that if a sample file is short enough, it's taken automatically as an oscillator.

src/sfizz/Wavetables.cpp
jpcima: What I thought of and forgot mentioning, it's also something that will check file changes and reload accordingly. I remember this happen for SFZ files but for samples I was not sure if it's the case or not.

src/sfizz/FilePool.h
jpcima: Should this be changed to int to make sure ?

paulfd: Apparently no, from https://en.cppreference.com/w/cpp/language/enum

> 1) declares a scoped enumeration type whose underlying type is int (the keywords class and struct are exactly equivalent)

so if int is lock-free this should be.

src/sfizz/Synth.cpp
paulfd: So what would be short enough? :D

jpcima: I thought substituting this conditional with something as `region->isOscillator()`, which is true under either these 2 conditions.

src/sfizz/Wavetables.cpp
paulfd: It happens for sfz files but the library users have to call it. Basically there's a "check if sfz files have changed" method and you have to call it and reload in your own time. In this case though, everything is purged and reloaded, at least for now. And this only checks the sfz files, not the sample files. If you change the sample files between preloading and actual loading, there will be pain...

src/sfizz/Wavetables.cpp
```diff
@@ -389,7 +389,7 @@ const WavetableMulti* WavetablePool::createFileWave(FilePool& filePool, const st
     };
 
     auto wave = std::make_shared<WavetableMulti>(
-        WavetableMulti::createForHarmonicProfile(hp, 1.0));
+        WavetableMulti::createForHarmonicProfile(hp, 1.0, config::tableSize, fileHandle->information.sampleRate));
```
jpcima: It's intentional that sample rate is not passed here.
A default 44100Hz is always fine, there is never need of harmonics > 22 kHz.

Beside all synth instances might share wavetables at some point; 44.1k wavetables may be used in a 96k synth for instance, but the reverse will not be true.

src/sfizz/Wavetables.h
```diff
@@ -185,15 +187,42 @@ class WavetableMulti {
...
+    const WavetableMulti* getFileWave(const std::string& filename);
+    /**
+     * @brief Load a file wave from the filepool and use it to create a wavetable.
+     * This function is not real-time safe.
+     *
+     * @param filePool the file pool to use to load the file
+     * @param filename the file name to load
+     * @return the wavetable
+     */
+    const WavetableMulti* createFileWave(FilePool& filePool, const std::string& filename);
```
paulfd: Considering how this is used, a boolean might be enough as a return value.

jpcima: No objection if you prefer it

PR #130, opened 2020-03-17
Initial parser rework

jpcima:
This is a rework of the parser, and its test program.
This parser is fully incremental, it supports reporting errors and warnings, as well as locations where they happen.

It will permit to add embedded samples, support is not yet there.
It's not embedded in the synth either, the previous parser is kept for now.

The test program is a SFZ editor in sfizz_demo_parser, and the test file is loaded as default, which illustrates many cases of errors.

### DIFF
--- a/clients/jack_client.cpp
+++ b/clients/jack_client.cpp
@@ -210,11 +210,11 @@ int main(int argc, char** argv)
     std::cout << "\tPreloadedSamples: " << synth.getNumPreloadedSamples() << '\n';
     std::cout << "==========" << '\n';
     std::cout << "Included files:" << '\n';
-    for (auto& file : synth.getIncludedFiles())
-        std::cout << '\t' << file.string() << '\n';
+    for (auto& file : synth.getParser().getIncludedFiles())
+        std::cout << '\t' << file << '\n';
     std::cout << "==========" << '\n';
     std::cout << "Defines:" << '\n';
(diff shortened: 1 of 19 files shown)


paulfd:
OK to merge, thanks :slightly_smiling_face:

jpcima:
I reverted a few things from the last commit, while adding abstract file parsing.

- if path is empty, it will invoke listener's `onParseBegin` but return without `onParseEnd`
  on the contrary, if it's empty, I want to let the parser emit a file-not-found error.
- I let the parser take its originalDirectory from the first file again, since I provided the alternative solution as discussed
- I let the directory be `.` if the path is relative. I suppose it to make sense.
If users want it otherwise, their responsability to pass absolute paths.

Regarding the abstract files
- `parseVirtualFile` accepts a path and reader. If reader is null, a file will be opened, otherwise the path will be not accessed and taken as virtual.
- the origin directory is deduced from the virtual path
- the virtual path is used for warnings and error locations

jpcima:
The commit 1510c0a permits to emit a better error in case of a file that cannot be included.

Before, if you try including a file "MyFileWhichDoesntExist.sfz", the error location would be `"MyFileWhichDoesntExist.sfz", 0, 0`.

From now, the error location will be the include statement in the parent file.
It's observable in the demo program. The error line now shows 18, it would be 0 in the older revision.

PR #131, opened 2020-03-18
Add effects: limiter, rectifier, gain, width

jpcima:
This adds the limiter effect. This one doesn't have opcodes.

Characteristics of this limiter / differences:
- like Cakewalk, this one limits to 0.0 dBFS
- Cakewalk's is a little faster, this one smoother
- A/R values taken from some documented recommendations (DAFX book, faust doc)
- Faust's stock one does not limit 0.0 dBFS when you give this value as threshold, but less (?)
- Faust's modulates the dB gain, this one modulates linear gain (like DAFX, faster code)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -24,6 +24,7 @@ set (SFIZZ_SOURCES
     sfizz/effects/Eq.cpp
     sfizz/effects/Apan.cpp
     sfizz/effects/Lofi.cpp
+    sfizz/effects/Limiter.cpp
 )
 include (SfizzSIMDSourceFiles)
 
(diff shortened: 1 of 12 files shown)


jpcima:
Add the wave rectifier, a book-only effect, but simple enough that I can add it.
(there are some others in the same case)

To make the opcode `rectify` significant, I turned it into a mix control.

```
<region>
sample=*sine
<effect>
type=rectify
rectify=100
rectify_mode=full
```

```
<region>
sample=*sine
<effect>
type=rectify
rectify=70
rectify_mode=half
```

paulfd (approved):

jpcima:
also I forgot, the limiter might benefit from oversampling 2x, in some cases. (also compressor/gate)
Since SFZ is supposed generic, maybe it will be nice to add it.

jpcima:
Added the limiter oversampling 2x.
It's ok to merge it.

PR #136, opened 2020-03-25
Envelope generator

jpcima:
This rewrites EG in such a way that:
- decay operates more similar to other products
- no exp/log calls in the process function

Fails the tests currently.
It may be the case that the release segment takes longer to reach 0, keeping the voice alive longer.
Release is considered finished when EG < virtuallyZero (-86dB). This threshold could be set higher.

Below are Cakewalk, Sfizz, ARIA (in that order)
Capture du 2020-03-25 19-58-21

### DIFF
--- a/src/sfizz/ADSREnvelope.cpp
+++ b/src/sfizz/ADSREnvelope.cpp
@@ -102,10 +102,22 @@ void ADSREnvelope<Type>::getBlock(absl::Span<Type> output) noexcept
 {
     State currentState = this->currentState;
     Type currentValue = this->currentValue;
+    bool shouldRelease = this->shouldRelease;
+    int releaseDelay = this->releaseDelay;
 
     while (!output.empty()) {
         size_t count = 0;
         size_t size = output.size();
+
+        if (shouldRelease && releaseDelay == 0) {
+            // release takes effect this frame
+            currentState = State::Release;
+            releaseDelay = -1;
+        }
+        else if (shouldRelease && releaseDelay > 0) {
+            // prevent computing the segment further than release point
+            size = std::min<size_t>(size, releaseDelay);
+        }
 
         switch (currentState) {
         case State::Delay:
@@ -139,23 +151,13 @@ void ADSREnvelope<Type>::getBlock(absl::Span<Type> output) noexcept
             }
             break;
         case State::Sustain:
-            if (freeRunning)
+            if (!shouldRelease && freeRunning) {
                 shouldRelease = true;
-            if (!shouldRelease) {
-                count = size;
-                currentValue = sustain;
-                sfz::fill(output.first(count), currentValue);
+                break;
             }
-            else {
-                if (releaseDelay > 0) {
-                    count = clamp<size_t>(releaseDelay, 0, size);
-                    currentValue = sustain;
-                    sfz::fill(output.first(count), currentValue);
-                    releaseDelay -= count;
-                }
-                if (releaseDelay <= 0)
-                    currentState = State::Release;
-            }
+            count = size;
+            currentValue = sustain;
+            sfz::fill(output.first(count), currentValue);
             break;
         case State::Release:
             while (count < size && (currentValue *= releaseRate) > config::virtuallyZero)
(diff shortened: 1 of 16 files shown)


jpcima:
Below is a comparison to Dimension, which has different EG that Dropzone.
It seems to fix the release time.

It's observable that Dimension has a slightly longer smooth duration.
Current sfizz is between the two of these, it is adjustable by the parameter of the exponential rate.

Capture du 2020-03-25 22-02-15

jpcima:
After slight adjusting.
Exponential factor=exp(-8/(t*Fs))

Cakewalk/Sfizz/ARIA
Capture du 2020-03-25 22-12-14

jpcima:
The tool is a Jack program that permits to capture the exact EG curves from Dimension with very little manual work.
This can be used in order to build the test set.

jpcima:
@paulfd Do you know an appropriate technique to compare the similarity of 2 curves overall?
The maximum absolute error does not work well.

---

For instance here are Sfizz vs Dim versions of the "Attack" envelope test.
They appear greatly similar at the first look.

Capture du 2020-03-27 12-58-48

When zooming into the release segment, it's observed there are large absolute errors locally, >0.1 in some cases.
Cakewalk looks like a reverse S-shape, suggests a different kind of smoothing. (maybe 2nd-order?)

Capture du 2020-03-27 12-59-58

paulfd:
Max absolute error will greatly favor outliers I would say. The mean squared/absolute error on the other hand does not tell you much about how the curves actually behave w.r.t. each other. I think we'll need to make it very specific for each test case: find the attack/sustain/release/... ranges, and compute a mean squared/absolute error here.

Between mean-squared and absolute, or any other penalty, I guess there are not really a best choice and it kinda depends on what will sound OK. I'd go with mean-squared or root-mean-squared due to habit but as I said, I'm not super clear on the psycho-acoustic part of what the best penalty would be for things to sound "the same".

paulfd:
You could also penalize the derivative if you want to capture smoothness effects.

jpcima:
Attached: some EG recordings from Dimension to match current unit tests
[eg-recordings.tar.gz](https://github.com/sfztools/sfizz/files/4400012/eg-recordings.tar.gz)

src/sfizz/ADSREnvelope.cpp
```diff
@@ -100,110 +99,92 @@ Type ADSREnvelope<Type>::getNextValue() noexcept
...
+        case State::Decay:
+            while (count < size && (currentValue *= decayRate) > sustain)
+                output[count++] = currentValue;
+            if (currentValue <= sustain) {
+                currentValue = sustain;
+                currentState = State::Sustain;
+            }
+            break;
+        case State::Sustain:
+            if (!shouldRelease && freeRunning) {
```
paulfd, line 158: Is the `!shouldRelease` check necessary here?

src/sfizz/ADSREnvelope.cpp
```diff
@@ -100,110 +99,92 @@ Type ADSREnvelope<Type>::getNextValue() noexcept
...
+                shouldRelease = true;
+                break;
+            }
+            count = size;
+            currentValue = sustain;
+            sfz::fill(output.first(count), currentValue);
+            break;
+        case State::Release:
+            while (count < size && (currentValue *= releaseRate) > config::virtuallyZero)
```
paulfd, line 167: So you're saying this could be done earlier? I'm not clear on the "I don't hear anything" level to be honest :)

paulfd (approved):
I will update the test when #142 is merged. The rebasing seems simpler in this order !

src/sfizz/ADSREnvelope.cpp
jpcima: Usually -60 dB, -80 dB sometimes

paulfd:
Btw I'm tempted to remove completely the `getNextValue()` API, since it's never used, I think your block version is also a bit "per-sample" but has small loops and should always perform better anyway. 1 less code path to maintain.

PR #137, opened 2020-03-25
Floating midi values

paulfd:
This changes the internal handling of MIDI values from unsigned to floats.
- Reduces the need for constant renormalization (`normalizeCC(...)` and associates)
- Transitions nicely into hi res midi

Tests are passing in debug, release and asan-tooled release builds. No apparent problem from my ardour test sessions. I have to check the performance for regressions (I need yet a faster way to do this :D)

### DIFF
--- a/benchmarks/BM_mapVsArray.cpp
+++ b/benchmarks/BM_mapVsArray.cpp
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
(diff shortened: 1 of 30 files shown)


paulfd:
[Report here](https://box.ferrand.cc/sfizz-reports/2020-03-27-floating-midi.html)

There is something very strange going on with the amplitude processing. There is a fixed (almost constant?) performance hit, that appears as soon as the maximum number of CCs is set to some number > 255. I'll try to check it on other platforms to see. It is strange because the amplitude processing does not really "loop" over CCs or relate to the number of CCs at all...

Anyway, this is not entirely related to this PR I guess.

paulfd:
This effect does not appear on intel platforms when I tried.

PR #143, opened 2020-03-30
Strings resonator effect

jpcima:
This adds the basic piano `strings` resonator.

I think, this version is not exactly designed like the original, but it has a more simplistic design that's still interesting, and, more importantly, is highly SIMD-able.

It feeds the signal into a bandpass filterpank, which in turns goes to 88 tuned string resonators.
The strings is a basic form of waveguide resonator, which doesn't involve fractional delay lines.

I add the opcode `string_wet`, a sfizz extension, because effects don't implement the CC yet which permit to control it like a sustain pedal.

Currently the effect is not well adjusted, it's stronger than reference, the original being more subtle.
It lacks a few things, so I left some TODO items where appropriate:
- the damping; all the strings play at equal gain
- the feedbacks; all strings resonate for identical durations

Example
```
<region>
sample=*saw

<effect>
type=strings
strings_number=88
strings_wet=100
```

### DIFF
--- a/src/sfizz/MathHelpers.h
+++ b/src/sfizz/MathHelpers.h
@@ -40,6 +40,18 @@ constexpr T min(T op1, Args... rest)
 }
 
 /**
+ * @brief Compute the square of the value
+ *
+ * @param op
+ * @return T
+ */
+template<class T>
+constexpr T power2(T in)
+{
+    return in * in;
+}
+
+/**
  * @brief Converts db values into power (applies 10**(in/10))
  *
  * @tparam Type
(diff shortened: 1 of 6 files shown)


src/sfizz/MathHelpers.h
```diff
@@ -39,6 +39,18 @@ constexpr T min(T op1, Args... rest)
     return min(op1, min(rest...));
 }
 
+/**
+ * @brief Compute the square of the value
+ *
+ * @param op
+ * @return T
+ */
+template<class T, class... Args>
```
paulfd: Not sure you need the `class... Args` here. However you could make it generic I guess, like  https://godbolt.org/z/gBefk4

paulfd (approved):

src/sfizz/MathHelpers.h
jpcima: Good catch yes, this was a copy&paste error

PR #145, opened 2020-04-01
Simplify the amplitude stage by only applying the final gain on the buffer

paulfd:
By avoiding these needless intermediate gains the amp stage went from 40 to 20 µs per voice on my test cases.


paulfd:
FYI: https://ferrand.cc/sfizz-reports/2020-04-01-ampstage.html

PR #158, opened 2020-04-02
Reorganize VST SDK as submodules

jpcima:
No patch for the moment, just a try

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,3 @@ node_modules/
 *.lock
 *.sublime-*
 *.code-*
-
-/vst/download
-/vst/external/VST_SDK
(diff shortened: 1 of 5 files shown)


jpcima:
Since it works, what do you think to merge it now?
This should allow to make the OBS builds of VST plugins.

paulfd:
Sure. Do you want to proceed with the patching in the build stage ?

paulfd:
If you're confident you can merge it directly no problem 😄

jpcima:
I will fork vstgui later. For now I posted them an issue report with links to info and patch.
I don't expect a quick answer (or even maybe an answer), we'll see.

First I will update OBS builds to produce the VST plugin.

Issue #146, opened 2020-04-01
Unison oscillator


jpcima:
When `oscillator_multi` is set in the range 3-9, this enables unison mode with the given number of oscillators. (0-2 being reserved values for AM/FM/PM)

It instanciates extra oscillators which run together with the first, and applies some additional panning and pitch.

When unison is on, `oscillator_detune` should be set to define their spread in pitch, or they will all pile up at the fundamental.

Example
```
<region>
sample=*sine
oscillator_multi=3
oscillator_detune=1200
```

Algorithm implementation below that determines the spread of oscillators. (dropzone)
https://gist.github.com/jpcima/efbb0fb7062d742375029d4de3210381

PR #161, opened 2020-04-03
Add oscillator unison

jpcima:
#146

This adds unison of oscillators.
The output is in stereo. Since stereo was only enabled for sample generators, I did a trick that enables it also in multi-oscillator mode.

Example
```
<region>
sample=*saw
oscillator=on
oscillator_detune=60
oscillator_multi=5
```


src/sfizz/Voice.h
```diff
@@ -284,7 +288,15 @@ class Voice {
     ADSREnvelope<float> egEnvelope;
     float bendStepFactor { centsFactor(1) };
 
-    WavetableOscillator waveOscillator;
+    static constexpr unsigned maxWaveOscillators = Default::oscillatorMultiRange.getEnd();
```
paulfd: An alternative for this one would be to put it in config and initialized the `Default::oscillatorMultiRange` as `{ 1, config::maxWaveOscillators }`

paulfd (approved):
Other than the small comment about the maxWaveOscillator configuration it looks good :)

jpcima:
It's fixed.

PR #166, opened 2020-04-06
Add support for curvecc and stepcc

### LINKED ISSUE #77: Controller curves
Implement `curvecc` for MIDI controller changes.

I have added an analysis in the [wiki](https://github.com/sfztools/sfizz/wiki/Curves).

### LINKED ISSUE #155: Add support for `stepcc` and `curvecc`
@jpcima already provided the curves in #96.

- [x] Add the curve index and step to the `CCModifier` structure.
- [x] Change the modifier lambdas to take these into account when necessary, with proper references to the curves.
- [x] Boolean check (?) in the modifier loop to handle the quantized case.

If necessary, maybe it's time to revisit having a transparent helper for the modifier loop (see e.g. https://godbolt.org/z/7ocfDj to add static assertions on the lambda parameter). In particular, I guess stepped-envelopes should _not_ be smoothed too?

paulfd:
Missing:

- [x] Tests as much as possible
- [x] Compare step behavior to ARIA and others for e.g. the number of steps and the steps for volume or other multiplicative envelopes.

Closes #155 #77


paulfd:
Ok should be good now! At least from a behavior pov :)

jpcima (approved):

jpcima:
one remark: `round` is certainly desired over `trunc` for our case

The performance of various rounding functions depends wildly on processor, flags, compiler.
`lrint` may compile more efficiently on the x86. It uses the current rounding direction.
(which can be set globally at the start of the processing method and restored)

paulfd:
You mean here:
```c++
    auto quantize = [step](float value) -> float {
        return std::floor(value / step) * step;
    };
```
and here:
```c++
    auto quantize = [logStep](float value) -> float {
        if (value > 1)
            return std::exp(logStep * std::floor(std::log(value) / logStep));
        else
            return std::exp(logStep * std::ceil(std::log(value) / logStep));
    };
```
? 

Can you expand a little bit? I used to have round there but it seemed from the tests I made with sforzando that trunc-type behavior was a bit more precise. I wonder if the first one should be `trunc` actually, so that the behavior is kept for negative values.

The tests I've tried to do with Sforzando showed that when you have steps, there is one last "step" exactly at the edge of the value range (for CC this would be at 1.0/127, for pitch bends at +/- 8192, ...). This is what led me to the trunc.

`lrint` could be nice though, I did not know this function, thanks.

jpcima:
This was a remark regarding 3f54a82 titled "Use round rather than trunc; more stable.."
only on topic of performance, and nothing to do with sforzando or the precision

paulfd:
I completely messed up the steps actually, I don't know why I treated it as a number of steps but it's actually a step size. Now this matches ARIA's behavior on all envelope types:
- `trunc` on the steps on *linear* envelopes such as amplitude
- ` round` on the pitch bend envelopes (which avoids most of the practical weirdness when using the pitch bend)
- `trunc` on all other multiplicative envelopes

I need to update the tests to match all behavior, and maybe add a couple to check the fine print, or assume that larger-scale tests should check for these.

PR #169, opened 2020-04-07
Replace AtomicGuards by mutexes and use RTSemaphore in the background loader

### LINKED ISSUE #151: Replace `AtomicGuard`s by `try_locks`
cf #53

The stress test mentioned in #124 and the performance logs should help checking the impact of this.

paulfd:
Closes #151 
Also solves a bug where AtomicGuards would not disable the callback on reloading.


paulfd:
@jpcima I moved the RTSemaphore to the main library but the vst has to include it from there. Is this acceptable?

paulfd:
The waiting times are much better :)

https://ferrand.cc/sfizz-reports/2020-04-07-rt-semaphore.html

src/sfizz/EQPool.cpp
```diff
@@ -108,9 +108,9 @@ sfz::EQPool::EQPool(const MidiState& state, int numEQs)
 
 sfz::EQHolderPtr sfz::EQPool::getEQ(const EQDescription& description, unsigned numChannels, float velocity)
 {
-    AtomicGuard guard { givingOutEQs };
-    if (!canGiveOutEQs)
+    if (!eqGuard.try_lock())
         return {};
+    DEFER(eqGuard.unlock());
```
jpcima: Try-locks are doable simply with use of `unique_lock`.
```
std::unique_lock<std::mutex> lock(mutex, std::try_to_lock);
if (!lock.owns_lock()) [...]
```

jpcima:
I have written a revision of RTSemaphore which has `noexcept` methods and `std::error_code`.
It's not tried, and I didn't try the compilation on Mac.
https://gist.github.com/jpcima/fab242de95e71d2f352372773f8a0c5e

src/sfizz/EQPool.cpp
paulfd: Man you're killing my defer macro :sob:
Thanks for the tip, I changed the locks.

paulfd:
You want to integrate and test it or should I? If only to track your own changes.

jpcima:
Rebase and force-push, I stored a copy at branch `try-locks.old` just in case

src/sfizz/Synth.cpp
```diff
@@ -1036,11 +1005,12 @@ void sfz::Synth::disableFreeWheeling() noexcept
 
 void sfz::Synth::resetAllControllers(int delay) noexcept
 {
-    AtomicGuard callbackGuard { inCallback };
-    if (!canEnterCallback)
+    resources.midiState.resetAllControllers(delay);
```
jpcima, line 1008: Is this really safe to reset controllers without a lock?
this call will be enqueuing MIDI events, right?

paulfd: I have to admit that this does not feel entirely right. The callback guard is disabled when voices or regions are touched, but I was thinking that even in this case you could still track midi events, such as allnotesoff. The controller reset does not do anything on the midi state but put a zero event on all controllers after a delay.

paulfd: The issue would probably be that the midi time is not advanced in the render method. The try lock failure should still advance the midi time.

jpcima: The validity of this model is questionable maybe.
Also it seems more a problem for a different issue.

It will be a behavior no different of noteOn/Off/etc, and also what's currently of sfizz.
Is that correct?

paulfd: All midi events basically get passed to the midi state now, but if the callback lock cannot be taken they have no impact on the voices, regions, or anything. The current behavior of sfizz should be the same except it used the AtomicGuards.

I agree, this may not be the ideal behavior, but I could not really make up my mind so I went with this solution. The thing is that in practice you would kind of expect that a plugin "knows" the current state of all controllers for example. I am not sure that host resend all controller state upon playback.

Anyway, if you find even one reason to change it we can. Most cases where the lock is taken would be so devastating that we'd be better off clearing the midi state in most cases anyway.

jpcima (approved):

PR #170, opened 2020-04-07
Add support for `offset_cc`

### LINKED ISSUE #159: Add support for `offset_cc`
Should be straightforward.

paulfd:
Closes #159 

It will be cleaner once the changes from #166 are merged

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -59,6 +59,12 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
     case hash("offset_random"):
         setValueFromOpcode(opcode, offsetRandom, Default::offsetRange);
         break;
+    case hash("offset_cc&"):
+        if (opcode.parameters.back() > config::numCCs)
+            return false;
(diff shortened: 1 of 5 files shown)


jpcima:
Rebased in order to make it readable (since the base branch changed after the PR)

src/sfizz/Region.cpp
```diff
@@ -59,6 +59,12 @@ bool sfz::Region::parseOpcode(const Opcode& opcode)
     case hash("offset_random"):
         setValueFromOpcode(opcode, offsetRandom, Default::offsetRange);
         break;
+    case hash("offset_cc&"):
+        if (opcode.parameters.back() > config::numCCs)
```
jpcima, line 63: I note there exist multiple occurrences of `ccNumber > config::numCCs` in code.
There is also `for (int cc = 0; cc < config::numCCs; cc++)`.
One of these has to be wrong. Perhaps instead the checks should use `>=`.

paulfd: You're right it should be `>=` in all these checks.

jpcima (approved):

Issue #153, opened 2020-04-02
Add a small amount of smoothing on the amplitude/crossfade envelope


paulfd:
Right now fast changes in amplitude/volume create discontinuities. This could use the `smoothcc` facilities when ready.

cf. #19

PR #172, opened 2020-04-07
Solve some issues with DSmolken's drumkits

paulfd:
- [x] `sample=*silence` table was not initialized and could output crazy values
- [ ] The free-running envelopes seem to play out the decay in ARIA; now we go directly to the release stage in sfizz. What does Dimension do?

If Dimension does the same as ARIA, we may need to add a counter to the decay stage on top of the rate. There is an example below.
[Example.zip](https://github.com/sfztools/sfizz/files/4446879/Example.zip)

Note also that `loop_mode=one_shot` does not free-run on generator regions in ARIA, but it does on sample regions.

### DIFF
--- a/devtools/CaptureEG.cpp
+++ b/devtools/CaptureEG.cpp
@@ -284,6 +284,7 @@ void Application::performSfzUpdate()
 
     QString code;
     code += "<region>\n";
+    code += "key=69\n";
     code += "sample="; code += QFileInfo(samplePath).fileName(); code += "\n";
     code += _ui->envelopeEdit->toPlainText();
 
--- a/src/sfizz/ADSREnvelope.cpp
+++ b/src/sfizz/ADSREnvelope.cpp
@@ -40,11 +40,9 @@ void ADSREnvelope<Type>::reset(const EGDescription& desc, const Region& region, 
 
     releaseDelay = 0;
     shouldRelease = false;
-    freeRunning = (
-        (region.trigger == SfzTrigger::release)
+    freeRunning = ((region.trigger == SfzTrigger::release)
         || (region.trigger == SfzTrigger::release_key)
-        || region.loopMode == SfzLoopMode::one_shot
-    );
+        || (region.loopMode == SfzLoopMode::one_shot && (region.isGenerator() || region.oscillator)));
     currentValue = this->start;
     currentState = State::Delay;
 }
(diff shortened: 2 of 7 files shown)


jpcima:
Some recordings of above sfz in cakewalk
[eg_decay.tar.gz](https://github.com/sfztools/sfizz/files/4447472/eg_decay.tar.gz)

jpcima:
This is at the generation of the decay stage.
`while (count < size && (currentValue *= decayRate) > sustain)`

The top of the attack stage is 100% level, and so is sustain level in this case.
The condition `currentValue > sustain` never holds, and the decay generator is never entered.

However, when using devtools/CaptureEG, it's observed that both Cakewalk and ARIA have some very long decays. (~18 seconds long)

paulfd:
So it seems that the envelope is not really free running in fact. The "release" happens when the sample has ended. The free running logic we have atm seems flawed, I added the one shot loop mode check for one of Alex's example that seemed reasonable but apparently ARIA does not follow this. What does Cakewalk do if you replace the sample by a generator?

paulfd:
Changing [this check](https://github.com/sfztools/sfizz/blob/7878a5c214c59c01f979550fe5fcbe77901dfb4f/src/sfizz/ADSREnvelope.cpp#L46)
```c++
freeRunning = (
    (region.trigger == SfzTrigger::release)
    || (region.trigger == SfzTrigger::release_key)
    || (region.loopMode == SfzLoopMode::one_shot)
);
```
into
```c++
freeRunning = (
    (region.trigger == SfzTrigger::release)
    || (region.trigger == SfzTrigger::release_key)
    || (region.loopMode == SfzLoopMode::one_shot && (region.isGenerator() || region.oscillator))
);
```
could be a way but this boolean is quite ugly.

Edit: it seems to work out though, tried it quickly.

jpcima:
This is another of ARIA. This has AHDS but no R phase.
~The whole envelope is longer that the audio file which is 11 seconds.~

```
 ampeg_attack=1.0 ampeg_hold=1.0 ampeg_decay=1.0 ampeg_release=30.0
ampeg_sustain=50
 loop_mode=one_shot
```

Capture du 2020-04-08 15-21-11

paulfd:
But there is an R phase in the file. You mean that the release phase is never active on the EG?

This fits with "loop_mode=one_shot and a sample file -> only play the EG up to the sustain and ignore the release" logic.

jpcima:
Correction on what is said above, it's correct.
There was a problem of the capturing program that created the plot above. It was not prepared to support loop mode, so the playing `key` had impact on the length and envelope.

It's good.
Capture du 2020-04-14 14-22-11

jpcima (approved):

paulfd:
Note that working on the smoothers, I suspect ARIA has a tiny "gain smoother" as I'm thinking of implemeting (#153 #181), which can explain the difference in the release and decay envelopes.

jpcima:
I'm not sure to understand immediately what this is, will check.
Given this one is validated, I will merge it.

PR #176, opened 2020-04-12
Add RTSemaphore and tests

jpcima:
This will try the build on macOS that I can't do locally.
It will be also nice if we can run the test.


paulfd:
On the mac you mean? Maybe we can add a travis job on the PR but remove it before merging..

PR #180, opened 2020-04-13
Fix libatomic detection on clang 10 compiler

jpcima:
On clang10, libatomic fails to be detected, and the program does not link.

The reason is as follows:
cmake builds a test program that tries to access the named symbol by address.
In clang, the `__atomic` functions work as builtins, and their address can't be taken.

Solution:
I'm attempting a `try_compile` which just links the library and nothing else.
I'm also changing libatomic into an INTERFACE target.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,7 +55,7 @@ target_sources(sfizz_static PRIVATE ${SFIZZ_SOURCES} sfizz/sfizz_wrapper.cpp sfi
 target_include_directories (sfizz_static PUBLIC .)
 target_include_directories (sfizz_static PUBLIC external)
 target_link_libraries (sfizz_static PUBLIC absl::strings absl::span)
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ configure_file (${PROJECT_SOURCE_DIR}/doxygen/scripts/Doxyfile.in ${PROJECT_SOUR
 
 add_library (sfizz::parser ALIAS sfizz_parser)
 add_library (sfizz::sfizz ALIAS sfizz_static)
-if (LIBATOMIC_FOUND)
+if (SFIZZ_LINK_LIBATOMIC)
     target_link_libraries (sfizz_static PRIVATE atomic)
```
paulfd: Isn't this redundant now?

jpcima: yes my bad, removed it

paulfd (approved):
👍

jpcima:
I wish I can make this compile check a bit cleaner, but for now I'll take it

PR #181, opened 2020-04-13
Smoothers and `smoothcc`

### LINKED ISSUE #22: Implement dynamic crossfade
From issue #19:

> Yes, the crossfade calculation are not dynamic right now and they're only checked when the note start. After reading around, this is indeed not what's expected so I'll look into it!

I'm opening this as a separate issue because the main bugs from that issue have been solved.

### LINKED ISSUE #48: Smooth CC formula

### LINKED ISSUE #153: Add a small amount of smoothing on the amplitude/crossfade envelope

paulfd:
This implements a linear and multiplicative smoother, trying to match a reasonable mix of behaviors between ARIA and consorts. The linear smoother uses a VA one-pole filter. The smoothing cutoff is a function of the sample rate, and it is quantized/tabulated between a minimum (a smoothing value of 1 with the default sample rate at 48 kHz) to a maximum (a smoothing value of 100 with a max sample rate of 192 kHz). We'd have to evaluate if this is enough.

- [x] Implement the smoothers with mapping to the `smoothcc` value and some tests.
- [x] Add a gain smoother on the voice for the overall amplitude (#22); the smooth value is set at 10 for now.
- [x] Parse the `bend_smooth` opcodes and add a bend smoother for the pitch bend in the Voices.
- [x] On parsing in the Synth, check the max number CC parameters there are for each modifier target and adapt the number of smoothers in the Voices.
- [x] Jointly loop over the modifiers and the smoothers to compute the modifier envelopes in the Voices.

Closes #48, #22, #153


jpcima:
Can you elaborate on what is the distinction between log and linear?
and why we need them? I lack the context to understand this.
("reasonable mix of behaviors between ARIA and consorts")

Also why's not simplest form of LPF not adequate, and why it's needed to tabulate the power function? (given smoothcc is supposed to be fixed?)

(sorry for lots of questions)

FrnchFrgg:
As a remember when things are working, xfin_smoothccXX and xfout_smoothccXX would be great, even though they are not supported by LinuxSampler (which arguably is the most complete implementation on Linux to date), and maybe even not by ARIA/Sforzando.

paulfd:
**edited out**

@FrnchFrgg this is linked to #22, and I added a gain smoother here which has this function more or less, but having smooth crossfades is one of the goal of this PR :slightly_smiling_face:

jpcima:
I don't find the filter to match to the RE'd specification.
Have you checked it?

jpcima:
This formula matches it exactly.
It lacked a term `2 pi` in the relation between the cutoff and time constant.
```
    smoothing = (smoothValue > 0);
    if (smoothing) {
        float tau = smoothValue * 3e-3f;
        float f = 1.0f / (twoPi<float>() * tau);
        float s = std::tan(pi<float>() * f / sampleRate);
        filter.setGain(s);
    }
```

paulfd:
No I was planning to check it out now so thanks :)

paulfd:
In the end it should reduce to this:
```
std::tan(1.0f / (2 * smoothValue * 3e-3 * sampleRate));
```

jpcima:
Yes; in addition you can group all constants together.

`std::tan((1.0f / (2 * 3e-3f)) / (smoothValue * sampleRate));`

Do we need to tabulate this? I thought the smoothcc are set once and never changed.

paulfd:
Well the filters belong to the voices, so the have to adapt to the currently played region for all `smoothcc`, and possibly sample rate. I guess we could set the smoothing filters upon loading the sfz file and make copies of the filter with their gain already set withing the voice, was this what you had in mind?

jpcima:
If I understand this correctly..
it's supposed to be changed in case of:
- loading sfz file
- changing sample rate

None of these belong to DSP processing, hence why care to save a small fraction of processing time?

paulfd:
No, it's changed when starting a voice, but I'll remove it and benchmark later if this creates problems.

jpcima:
Feel free to cherry-pick 272b195, a test jack client

paulfd:
You can use this to do further work on modulation if you wish.

src/sfizz/Voice.cpp
```diff
@@ -718,3 +714,59 @@ void sfz::Voice::setupOscillatorUnison()
...
+void sfz::Voice::pitchEnvelope(absl::Span<float> pitchSpan) noexcept
+{
+    const auto numFrames = pitchSpan.size();
+    auto bends = resources.bufferPool.getBuffer(numFrames);
+    if (!bends)
+        return;
+
+    const auto events = resources.midiState.getPitchEvents();
+    const auto bendLambda = [this](float bend) {
+        return region->getBendInCents(bend);
```
jpcima: @paulfd, I'm trying to rebase this.
This function name is misleading. It returns a factor not cents.

jpcima (approved):

jpcima:
I seem to have trouble with the following patch.
It's set to control panning left and right using a midi controller change.

When smoothcc is set and the note is keyed on, pan is going to start at the right and it will take a delay until the pan value establishes itself at center.

```
<control>
set_hdcc16=0.5

<region>
sample=*sine
ampeg_attack=0.02
ampeg_release=0.1
pan=-100
pan_oncc16=200
pan_smoothcc16=100
```

paulfd:
I have no computer to try this until tomorrow, sorry. You mean that the value always start on the left on key presd even though the cc value is not changed?

If so it's a problem, it means the panning value does not properly query the CC state at some point yeah.

Jun 20, 2020 17:25:29 JP Cimalando <notifications@github.com>:

jpcima:
It appears the initialization of the filter which is incorrect. I 'll check it.

jpcima:
The modulated unit has been changed to % after rebase, but the initialization had not followed.
This concerns amplitude, pan, width and position.
Fixed in e69365e

PR #192, opened 2020-04-19
Add the modulation matrix

jpcima:
This is the candidate modulation matrix.
It's documented in the header.

Just the basic code, it's not connected to anything.
It's a basic framework, `ModGenerator` is expected to be able to express EG, CC, and LFO.

It's supposed to be used in the cycle of the render method, in a way similar to this.
```
mm ← theModMatrix
mm.beginCycle(numFrames)
for voiceNum ← [0..numVoices-1]
    mm.beginVoice(voiceNum)
    [...]
```

The modulation matrix should be filled, and the connections created, during the opcode parsing stage.

A modulation is computed in a lazy way, by `getModulation`. There is different processing depending if the modulator is per-voice or not.

If `getModulation` returns `nullptr`, the modulation does not exist, then it means the code can choose to call a faster non-modulating variant (for example the filter cutoff). 

For now it doesn't compile because `Buffer` deletes the move assignment.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,7 @@ set (SFIZZ_SOURCES
     sfizz/SfzFilter.cpp
     sfizz/Curve.cpp
     sfizz/Wavetables.cpp
+    sfizz/ModMatrix.cpp
     sfizz/RTSemaphore.cpp
     sfizz/Effects.cpp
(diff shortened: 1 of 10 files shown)


jpcima:
Note: I consider to add the modulation depths also as part of the matrix, as a multiplier to the sources and targets.

PR #193, opened 2020-04-19
Do some cleanup in the Buffer class

jpcima:
A few improvements in the `Buffer` class.

- add the move operations
- use `unique_ptr` for the allocation
- have throwing and `nothrow` variants (with default to the former)
- do not count the non-allocated instances as "new buffers" (this facilitates move)
- add `const` and `noexcept` everywhere applicable
- take care of "TODO C++17" regarding the static variable

### DIFF
--- a/src/sfizz/AudioBuffer.h
+++ b/src/sfizz/AudioBuffer.h
@@ -64,17 +64,30 @@ public:
      * @return true if the resize worked
      * @return false otherwise
      */
-    bool resize(size_t newSize)
+    bool resize(size_t newSize, std::nothrow_t) noexcept
     {
         bool returnedOK = true;
 
         for (size_t i = 0; i < numChannels; ++i)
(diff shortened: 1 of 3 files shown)


jpcima:
Btw, the `Buffer<float>` and the `Buffer<char>` are going to have different counters.
Is it desired behavior?

paulfd:
Yes it was like this from the beginning. For its purpose of tracking I thought it did not matter too much since everyone must be at zero.

paulfd:
Ok with merging I am on my phone now.

jpcima:
Ok I will merge. I went crazy with tests to make sure not to break.
A static init for the buffer counter revealed a bad idea, especially given that Catch itself runs in the static init apparently.
So the original paradigm is good, but I make the counter singleton such that it works for all Buffer types.

Issue #195, opened 2020-04-20
Block comments in SFZ file fail to load


AlexMitchellMus:
## Issue
Block comments starting with `/*` and ending with `*/` cause Sfizz to not load the SFZ file, and print to Ardour log:
2020-04-20T17:36:34 [ERROR]: [sfizz] Error with EXAMPLE_FILE; no file should be loaded

Removing block comments fixes issue. Comments starting with `//` do not exhibit this issue.

## Versions:
Latest Sfizz Git
Latest Ardour Git

## Example File:
```
/*
*/
<region> sample=*sine
```

jpcima:
It's fixed starting with commit c2767b3

PR #196, opened 2020-04-20
Support block comments

jpcima:
#195 

This adds support of block comments in the parser.
There are some examples in the parser demo.

Maybe not all possibilities are supported.
For example, don't expect to be able to write: `sample=/*blabla*/mycomment.wav`

If the comment is not terminated with `*/`, an error will be raised.

### DIFF
--- a/src/sfizz/parser/Parser.h
+++ b/src/sfizz/parser/Parser.h
@@ -84,7 +84,13 @@ private:
     void flushCurrentHeader();
 
     // helpers
-    static int hasComment(Reader& reader); // 0=None 1=Line 2=Block
+    enum class CommentType {
+        None,
+        Line,
+        Block,
+    };
+
+    static CommentType getCommentType(Reader& reader);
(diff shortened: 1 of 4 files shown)


src/sfizz/parser/Parser.h
```diff
@@ -84,8 +84,8 @@ class Parser {
     void flushCurrentHeader();
 
     // helpers
-    static bool hasComment(Reader& reader);
-    static size_t skipComment(Reader& reader);
+    static int hasComment(Reader& reader); // 0=None 1=Line 2=Block
```
paulfd: Maybe an enum would be slightly clearer here for the future

paulfd (approved):

jpcima:
- changed comment type to an enum
- modified `skipComment` so the character count returned is correct
  (note it doesn't prevent the parser to work correctly, but better to fix it regardless)

PR #200, opened 2020-04-22
Add a sfizz_render client in-tree

paulfd:
This uses the `fmidi` library and its player, which adds a dependency on `libev`. The goal of having this client in-tree is to facilitate benchmarking of different versions.

@jpcima not sure I use your library right so check it :slightly_smiling_face:

### DIFF
--- a/clients/sfizz_render.cpp
+++ b/clients/sfizz_render.cpp
@@ -1,0 +1,208 @@
+
+#include <sndfile.hh>
+#include "sfizz/Synth.h"
+#include "sfizz/MathHelpers.h"
+#include "sfizz/SfzHelpers.h"
+#include "sfizz/SIMDHelpers.h"
+#include "MidiHelpers.h"
+#include "cxxopts.hpp"
+#include <fmidi/fmidi.h>
+#include <iostream>
+
+#define LOG_ERROR(ostream) std::cerr  << ostream << '\n'
+#define LOG_INFO(ostream) if (verbose) { std::cout << ostream << '\n'; }
+#define ERROR_IF(check, ostream) if ((check)) { LOG_ERROR(ostream); std::exit(-1); }
+
+struct CallbackData {
+    sfz::Synth& synth;
+    int trackIdx;
+    double lastCallTime;
+    double blockSizeInSeconds;
+    bool finished;
+};
+
+void midiCallback(const fmidi_event_t * event, void * cbdata)
+{
+    auto data = reinterpret_cast<CallbackData*>(cbdata);
+
+    if (event->type != fmidi_event_type::fmidi_event_message)
+        return;
+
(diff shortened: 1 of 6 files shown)


jpcima:
>This uses the fmidi library and its player, which adds a dependency on libev. 

This is not true, it does not add a dependency on libev.
Only the example programs are dependent of the library, libev was moved out of the main library since a long time.

The library depends on `boost` and on `fmt`, and that's only for the description API which converts SMF to plain text.

I have around an edit of the library that can disable the above feature, such that it does not have external deps. Then it's a ~140K source code without recursive submodules.
https://github.com/jpcima/midiplay/tree/master/thirdparty/fmidi
https://github.com/jpcima/midiplay/blob/21c095fdf6b8326dc00608046e9ccf6f28b734a2/Makefile#L77

paulfd:
It seems I had to install libev to access the `fmidi_player_XX` interfaces? After some more reading it seems to be unneeded. There are these lines in the CMakeLists
```cmake
find_library(ev_LIBRARY ev)
set(fmidi-player_BUILD FALSE)
if(ev_LIBRARY)
  set(fmidi-player_BUILD TRUE)
else()
  message(STATUS "libev is missing, NOT building library fmidi-player.")
endif()
```

clients/sfizz_render.cpp
```diff
@@ -0,0 +1,208 @@
...
+void midiCallback(const fmidi_event_t * event, void * cbdata)
+{
+    auto data = reinterpret_cast<CallbackData*>(cbdata);
+
+    if (event->type != fmidi_event_type::fmidi_event_message)
+        return;
+
+    switch (midi::status(event->data[0])) {
+        case midi::noteOff:
+            data->synth.noteOff(event->delta, event->data[1], event->data[2]);
```
jpcima: The field `event->delta` is not relevant to sequencing based on delta times, it's a raw MIDI delta value (related to tempo).

For now, the way to go is to force delta as 0, and process in small buffer sizes. (in the order of 5ms)

Ideally the library should pass the player a timestamp, which right now is not the case.

jpcima:
>It seems I had to install libev to access the fmidi_player_XX interfaces?

It's the example player that comes with as part of the library. (see `programs/`)

clients/sfizz_render.cpp
jpcima: In fact, a much better idea, after thinking better about it:

Keep the block size large, like it is.
Tick the midi player by 1/sampleRate for each frame, and increment a counter each time in the callback context, that starts at 0 for each block.
Use this counter as MIDI delay to sfizz events.

paulfd:
If we go with #201 can I close this one safely?

jpcima:
Yes, it's no problem to close this one.
It's based on the identical commit ref 6f39d4c.

Issue #202, opened 2020-04-23
Support for embedded WAV loop ranges


agittins:
TL/DR: sfizz doesn't seem to support loop ranges embedded in wav samples. They seem to work in linuxsampler. They are (heavily?) used in VirtualPlayingOrchestra.

A number (lots?) of instruments use loop markers that are embedded in the WAV file itself. 

An example is the VPO French Horn, which uses Mattias Westlund's looped samples. The sfz file has no loop ranges defined in it, but when opened in linuxsampler, it loops the samples indefinitely and smoothly. Using the same sfz in sfizz (0.3.2) it loops the entire sample, which is... not as nice. :-)

As I understand it (and this is only based on this evening's googling!), samplers often use the "smpl" block in a wav file to define cue points for looping.

This seems to be the case with the french horn samples, here's what I get from running sndfile-info against them:

```
$ sndfile-info horns-sus-ff-e4-PB-loop.wav 
========================================
File : horns-sus-ff-e4-PB-loop.wav
Length : 647600
RIFF : 647592
WAVE
fmt  : 16
  Format        : 0x1 => WAVE_FORMAT_PCM
  Channels      : 2
  Sample Rate   : 44100
  Block Align   : 4
  Bit Width     : 16
  Bytes/sec     : 176400
data : 647488
smpl : 60
  Manufacturer : 0
  Product      : 0
  Period       : 22675 nsec
  Midi Note    : 60
  Pitch Fract. : 0
  SMPTE Format : 0
  SMPTE Offset : 00:00:00 00
  Loop Count   : 1
    Cue ID :  0  Type :  0  Start : 33625  End : 127592  Fraction :     0  Count :     0
  Sampler Data : 0
End

----------------------------------------
Sample Rate : 44100
Frames      : 161872
Channels    : 2
Format      : 0x00010002
Sections    : 1
Seekable    : TRUE
Duration    : 00:00:03.671
Signal Max  : 23198 (-3.00 dB)
```

```Type``` seems to map directly to SFZv2's ```loop_type``` opcode.

The smpl chunk seems pretty well documented here: https://sites.google.com/site/musicgapi/technical-documents/wav-file-format#smpl and I guess since sndfile-info is able to read it there's probably some level of support for extracting this in libsndfile.

Happy to provide sample files etc, but I assume you've probably got VPO already :-)

jpcima:
Hello, sfizz is supposed to support this feature already.
[...]

PR #204, opened 2020-05-01
Fix the use of file loop markers

jpcima:
#202 

Consider the loop marker first, otherwise sample end.
Does this appear to be a correct fix?


paulfd:
Yes I think I messed up this commit... 😨

PR #205, opened 2020-05-02
Misc

jpcima:
Eliminate some warnings from headers.
They are emitted using Clang 10.

In C++11, it's not allowed for constexpr functions to have local variables.


paulfd (approved):
We should try to push the atomic_queue changes upstream.

jpcima:
PR is sent.

PR #206, opened 2020-05-03
Some cleanup around cmake and sfizz_render

jpcima:
- it would not build if `SFIZZ_JACK` was not set
- I've edited fmidi such that libfmt can be optional, if we don't use printing APIs (which we don't)
- in fmidi upstream, I added a script that can roll up the whole source into a single file. (71 kB)
I have replaced the submodule with this source code, hence 2 submodules less to deal with.


jpcima:
Since `sfizz_render` was never apparently built on Windows OS (consequence of `SFIZZ_JACK`), there are the additional changes to make it work.

clients/sfizz_render.cpp
```diff
@@ -143,7 +143,7 @@ int main(int argc, char** argv)
     ERROR_IF(!synth.loadSfzFile(sfzPath), "There was an error loading the SFZ file.");
     LOG_INFO(synth.getNumRegions() << " regions in the SFZ.");
 
-    fmidi_smf_u midiFile { fmidi_smf_file_read(midiPath.c_str()) };
+    fmidi_smf_u midiFile { fmidi_smf_file_read(midiPath.u8string().c_str()) };
```
paulfd, line 146: This is under windows? Should we pass `native()`?

paulfd (approved):

clients/sfizz_render.cpp
jpcima: The deal here is that Windows native paths are widechar, and fmidi accepts `char` only.

The other problem of Windows is that `char` file APIs use the local ANSI codepage instead of UTF-8. This is always a problem when you'd like to open a file which has extended characters in the path.

The commonly accepted resolution to this, it's that you convert the wide path to UTF-8 (`u8string`), and at app level handle with `MultiByteToWideString` followed by `_wfopen`.
(also it's why I introduced the update at 1226a59)

PR #209, opened 2020-05-03
Opcode helpers

jpcima:
This adds a few helpers needed later to process opcode names.

- get the opcode category: whether basic, _oncc, _curvecc, etc.
- get an opcode's derived name, within a target category

See below in tests.

### DIFF
--- a/src/sfizz/Opcode.cpp
+++ b/src/sfizz/Opcode.cpp
@@ -6,11 +6,14 @@
 
 #include "Opcode.h"
 #include "StringViewHelpers.h"
+#include "absl/strings/ascii.h"
+#include "absl/strings/match.h"
 #include <cctype>
 
 sfz::Opcode::Opcode(absl::string_view inputOpcode, absl::string_view inputValue)
     : opcode(trim(inputOpcode))
     , value(trim(inputValue))
+    , category(identifyCategory(inputOpcode))
 {
     size_t nextCharIndex { 0 };
     int parameterPosition { 0 };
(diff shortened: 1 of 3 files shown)


paulfd:
OK I have to admit that I do not see the point now but I believe this integrates with further work. What is the use of `getDerivedName`? Do we need to "pay" the cost of categorization for all opcodes if we're not using this function later?

jpcima:
>What is the use of getDerivedName?

Regarding the bigger picture:
If you recall about CC modulations, there is a need to extract one particular mod target's cc/step/smooth/curve all at once, because this tuple will be modulation ID.
This routine comes in handy to scan the opcode list, in search of buddy opcodes that relate with a particular `oncc`.

>Do we need to "pay" the cost of categorization for all opcodes if we're not using this function later?

Not necessarily, if you'd prefer to recall the categorizing function each time instead of member variable inside of `Opcode`.

paulfd:
Hmm right now this tuple exists within each region, since the opcodes are a bit "volatile" in the sense that they kind of disappear after the parsing step and the values of reference are stored within the regions. I guess you're trying to have this tuple live outside of each region? 

Btw I tested the performance hit and this is basically invisible, so I guess there is no real problem either way :)

paulfd (approved):

jpcima:
The goal is that to fill in the mod-matrix as the same time CC opcodes are handled in the region.
The trouble regarding this tuple, it's actually a key, in a map container.
As such, it's not something that is so easily incrementally modified.

If this was desired, then it'd need to find the CC source in of matrix (involving a small O(N)), extract it, change ID, put it back in and reconnect.
On second thought, maybe that it's **not so bad**.
- A positive is this idea doesn't systematically break the unit tests, which are depending on incremental region opcode handling.
- On the other hand, it would be still enjoyable to have the opcode helpers around, so we avoid duplicating every time the manual handling of `step`/`smooth`/`curve` for every cc

PR #212, opened 2020-05-04
Opcode cleanup

jpcima:
Add the normalization of opcode names, and unit test.


jpcima:
Start with 4460114, the normalization is in effect in the synth, all the opcodes in the switch are deduplicated. (aliases, implicit indices)

Regarding `fil_gaincc`, it was a choice to either implement its case in the cleanup function, or remove it; given it doesn't have an implementation in ARIA, it's why I chose to rather remove it.

paulfd (approved):

PR #213, opened 2020-05-05
Cc processing

jpcima:
It's a followup on PR #212. ~~Only the last commit is relevant to the actual change.~~

This is a suggestion to add generic processing of CC opcodes, such that curve/step/smooth is managed without code repetitions.

As example, 6 opcodes are rewritten in this fashion, removing a lot of code from `Region.cpp`.

Some notes:

- I'm keeping the `oncc` values verbatim, eg. no percent normalizing.
  Perhaps this is wise in perspective in the mod matrix, but I'm not sure if necessary.
  (test failure as consequence, but let's discuss it)

- ~~it's assumed `volumeStepRange`, `tuneStepRange` are incorrect, replaced with `stepCCRange` in all cases. Is this OK?~~

  ~~Moreover, sfzformat indicates that `stepcc` default to 127 if not given.~~
  - ~~The struct `Modifier` has default step to 0, should I change it to 127?~~
  - ~~With `step=0` being invalid, should 1 be set as a minimum for stepcc?~~

  cf. https://sfzformat.com/modulations/stepccN

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -15,6 +15,7 @@
 #include "absl/strings/str_cat.h"
 #include "absl/algorithm/container.h"
 #include <random>
+#include <cassert>
 
 template<class T>
 bool extendIfNecessary(std::vector<T>& vec, unsigned size, unsigned defaultCapacity)
@@ -36,6 +37,13 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
     const Opcode opcode = rawOpcode.cleanUp(kOpcodeScopeRegion);
 
     switch (opcode.lettersOnlyHash) {
+    // Helper for ccN processing
+    #define case_any_ccN(x)        \
+        case hash(x "_oncc&"):     \
+        case hash(x "_curvecc&"):  \
+        case hash(x "_stepcc&"):   \
+        case hash(x "_smoothcc&")
+
     // Sound source: sample playback
     case hash("sample"):
         {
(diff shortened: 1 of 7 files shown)


src/sfizz/Region.cpp
```diff
@@ -932,11 +814,49 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
...
     default:
         return false;
+
+    #undef case_any_ccN
     }
 
     return true;
 }
 
+bool sfz::Region::processGenericCc(const Opcode& opcode, Range<float> range, CCMap<Modifier> *ccMap)
```
paulfd, line 824: ccMap could be a reference maybe? Do you expect that this function could be called for a null map?

paulfd: range could also be const ref although it's a small object anyway.

paulfd (approved):

src/sfizz/Region.cpp
jpcima: I don't expect these CCMaps to remain as this long-term. I wish to move it somewhere related with the mod matrix. They stay around only for now to preserve unit tests.

As of the const-ref, Range is a small enough object that I pass it like this.
Range=2*32 bits no indirection, Ref=64 bits and indirection

jpcima:
To recap regarding the stepcc, and before proceeding to merge this.
- in the future, we should not bound cc ranges nor step ranges
- the present commit leaves an automatic detection of stepcc range, which is maybe not the most logical one, but matches most current tests (except just 1, which I've modified accordingly)

Also:
- such things as cutoff/resonance are not touched yet, their logic has a bit more complication.
  All things which keep `CCMap<float>` or `CCMap<int>` should be changed to modifiers.
  For the matrix, it needs to convert it to identical forms as the others.

paulfd:
Agreed with your last comment! I don't know if you were waiting to merge this?

jpcima:
It can be merged now, although no urgency about it.
it's a brach that mod matrix is currently based on.

PR #214, opened 2020-05-05
Improve the note stealing

### LINKED ISSUE #187: Improve the note stealing
Use the power history to kill the quietest note if below some threshold.

paulfd:
So the previous note stealing was very conservative. Now, we make use of the average power computed by all voices after each block to steal the voice that had the lowest mean-squared score in the last 10 blocks, without further checks. The reasoning is that most of the time under such pressure a new note will always trump an old one, and we might as well take the lowest.

Also, I put the default polyphony to 256 and increased the max to 512.

Close #187

### DIFF
--- a/src/sfizz/HistoricalBuffer.h
+++ b/src/sfizz/HistoricalBuffer.h
@@ -37,6 +37,7 @@ public:
 		buffer.resize(size);
 		fill<ValueType>(absl::MakeSpan(buffer), 0.0);
 		index = 0;
+        validMean = false;
 	}
 
     /**
@@ -46,6 +47,7 @@ public:
      */
 	void push(ValueType value)
 	{
+        validMean = false;
 		if (size > 0) {
 			buffer[index] = value;
 			if (++index == size)
@@ -60,11 +62,18 @@ public:
      */
 	ValueType getAverage() const
 	{
-		return mean<ValueType>(buffer);
+        if (!validMean) {
+            mean = sfz::mean<ValueType>(buffer);
+            validMean = true;
+        }
+
+		return mean;
 	}
 private:
 	Buffer<ValueType> buffer;
 	size_t size { 0 };
 	size_t index { 0 };
+    mutable bool validMean { true };
+    mutable ValueType mean { 0.0 };
 };
 }
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -240,7 +240,13 @@ void sfz::Voice::renderBlock(AudioSpan<float> buffer) noexcept
 
     updateChannelPowers(buffer);
 
-    this->triggerDelay = absl::nullopt;
+    age += buffer.getNumFrames();
+    if (triggerDelay) {
+        // Should be OK but just in case;
+        age = min(age - *triggerDelay, 0);
+        triggerDelay = absl::nullopt;
+    }
+
 #if 0
     ASSERT(!hasNanInf(buffer.getConstSpan(0)));
     ASSERT(!hasNanInf(buffer.getConstSpan(1)));
@@ -629,6 +635,7 @@ void sfz::Voice::reset() noexcept
     region = nullptr;
     currentPromise.reset();
     sourcePosition = 0;
+    age = 0;
     floatPositionOffset = 0.0f;
     noteIsOff = false;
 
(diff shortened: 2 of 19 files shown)


src/sfizz/HistoricalBuffer.h
```diff
@@ -60,11 +62,18 @@ class HistoricalBuffer {
      */
 	ValueType getAverage() const
 	{
-		return mean<ValueType>(buffer);
+        if (!validMean) {
+            mean = sfz::mean<ValueType>(buffer);
+            validMean = true;
+        }
```
jpcima, line 68: You can track the power in more efficient ways which don't involve O(n) computation.

- one is always to keep the moving sum so you skip the O(n) completely.
  As a value gets pushed, remove from the moving sum the buffer's oldest value, and add the new one.
  A problem is that this can drift. A solution is to use fixed-point integers instead of floats, which make it exact.
  Then you can easily make the length way >16, which currently isn't really much.

- otherwise to track the signal level, another possibility commonly used:
  take the absolute values of input samples, pass it in lowpass filter that has different response times on rising and falling signal. It should be immediate on rising and slow on falling.
similar to this https://fr.wikipedia.org/wiki/Redresseur#Filtrage_de_la_tension_de_sortie_d'un_pont_de_diode

jpcima: Ok my bad I misunderstood on first read.
Regardless the current solution faces a tiny problem, it's the power computation will be different depending how you render from the library, where in small buffers or large ones.
So the solutions above are relevant in this regard.

paulfd:
I added the envelope/power tracker as a simple `abs` + super-lowpass filtering set at 5 Hz for now. We may further tweak this part, but at least it's independent of the block size now. I also added the voice age as a possible metric.

The heuristic now is to sort voices by age, set a threshold of the envelope (now set at "average envelope value over the voices / 4"), try to find a voice which fits the bill to kill. If we go beyond "max voice age / 2", we just pick the oldest one.

I also kill all "sister voices" now, which are started at the same time and with the same trigger. Writing this down I realize that this can be a problem if one of the sister voice is super strong, so we still need to refine all of this. 

Maybe consider the average envelope over all sister voices, or just do not kill sister voices?

paulfd:
The more I think about this the more I think we need an opcode so instrument makers can set a voice stealing that is adapted to their instrument behavior...

src/sfizz/Voice.cpp
```diff
@@ -718,3 +734,24 @@ void sfz::Voice::setupOscillatorUnison()
...
+
+    auto tempSpan = resources.bufferPool.getBuffer(buffer.getNumFrames());
+    if (!tempSpan)
+        return;
+
+    for (unsigned i = 0; i < channelPowers.size(); ++i) {
+        for (unsigned s = 0; s < buffer.getNumFrames(); ++s)
+            (*tempSpan)[s] = std::abs(buffer.getConstSpan(i)[s]);
+
+        channelPowerFilters[i].processLowpass(*tempSpan, *tempSpan);
```
jpcima: Can the one-pole filter have:
- `inline float tickLowpass(float x)`
- `inline float tickHighpass(float x)`

There isn't benefit to run it in buffers, since it's data-dependent from one cycle to the next.
This eliminates the temp span.

jpcima:
>Maybe consider the average envelope over all sister voices, or just do not kill sister voices?

Perhaps consider the max of all of them?
From the case of a loud voice like you mentioned, I think this to make the most sense.

src/sfizz/Voice.cpp
paulfd: Done :+1:

paulfd:
> > Maybe consider the average envelope over all sister voices, or just do not kill sister voices?
> 
> Perhaps consider the max of all of them?
> From the case of a loud voice like you mentioned, I think this to make the most sense.

Did that, and I also added that killed voices pre-render their block in all relevant effect busses. This data is "ramped out" at the beginning of the block before the rendering of active voices. It's a linear ramp now, it could be an exponential ramp.

The behavior seems good, on hardcore piano stuff (like say Chopin or Liszt) with relatively low polyphony you don't "hear" dropped notes, and if you're way crazy on the polyphony you can sometimes distinguish something going wrong.

I was bitten by the fact that the voice kills itself btw. Maybe flagging ended voices and letting the synth clean them up is a better idea in the end.

~I should polish and split up the stealing function, it ended up quite large and convoluted.~

jpcima (approved):

jpcima:
Looking good. Another thing, you may consider as heuristic to check if there exists already other voices using the same instrument, and increase the killability score if this is the case.

paulfd:
I'll merge the current state and listed possible improvements for the next release based on our discussions. Thanks all!

PR #216, opened 2020-05-06
Add a build helper for DPF

jpcima:
This adds a build helper for DPF plugins with sfizz.
There is an explanation how to use it at the top of the file.

@rghvdberg

### DIFF
--- a/dpf.mk
+++ b/dpf.mk
@@ -34,8 +34,8 @@
 # 4. Build using `make` from the top folder.
 #
 
-SFIZZ_DIR = $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-SFIZZ_BUILD_DIR = $(SFIZZ_DIR)/dpf-build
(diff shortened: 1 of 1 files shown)


jpcima:
Abseil seems to pull more deps under `DEBUG=true`, hence I updated to take it in consideration.

@pdesaulniers hi, possibly it interests you to check this work as well.

paulfd (approved):

paulfd:
Btw, how come we pull in so many unrelated things from abseils? Like the time zone and synchronization primitives. I think we mostly use optional, span, the atoi-type helpers and a couple string functions here and there no? And the string_views..

jpcima:
I've included there the strict necessary set of libraries which is needed to please the linker, not one more.
Their respective source listings are from abseil's subdirectory CMakeLists, listed under `SRCS`.

paulfd:
The question was more aimed at absl than your script :slightly_smiling_face:

PR #217, opened 2020-05-06
Add the string loading API

jpcima:
As previously discussed.
cc @rghvdberg


paulfd (approved):

jpcima:
The header guard is already in, hence this commit redundant.
I'll remove it and merge.

PR #219, opened 2020-05-07
Make the opposing curves equal at MIDI value 64

jpcima:
Make the opposing curves equal at MIDI value 64, defined as center by MIDI specs.
- [0,1] should have center at 0.5
- [-1,1] should have center at 0
- Xfades should have center at 1/√2

TODO: the tests

### DIFF
--- a/src/sfizz/Curve.cpp
+++ b/src/sfizz/Curve.cpp
@@ -14,7 +14,7 @@
 namespace sfz
 {
 
-static const Curve defaultCurve = Curve::buildBipolar(0.0, 1.0);
+static const Curve defaultCurve = Curve::buildBipolar(0.0, 1.0, true);
 
 Curve Curve::buildCurveFromHeader(
     absl::Span<const Opcode> members, Interpolator itp, bool limit)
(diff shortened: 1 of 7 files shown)


paulfd (approved):

paulfd:
I added/updated the tests to match the behavior, and I also changed the `normalizeXXX` functions so that midi 64 is 0.5f (which in return already corrects many tests :slightly_smiling_face:).

I'm ok for the merge!

paulfd (approved):

jpcima:
Nice thanks for updating this. I'll add just a few observations.

- there was a claim on the main channel that ARIA does not center at 64. This is based on Cakewalk.
- also the hdcc thing came later based on some sample set needs, and this raises a new problem: the hdcc 0.5 is supposed to be center, but this becomes a mismatch to cc7bit when (de)normalizing from one to another.

paulfd:
Well I think this would be better practice to center at 64 anyway... for the (de)normalization, I updated the normalizing, you'd want to adapt the denormalizing too? Not super pretty but

```cpp
template<class T, absl::enable_if_t<std::is_integral<T>::value, int> = 0>
inline CXX14_CONSTEXPR constexpr T denormalize7Bits(float value)
{
    if (value < 0.0f)
        return T { 0 };

    if (value > 1.0f)
        return T{ 127 };

    if (value <= 0.5f)
        return static_cast<T>(value * 128.0f);

    return T{ 64 } + static_cast<T>((value - 0.5f) * 126.0f);
}
```
would work I guess?

paulfd:
I added the previous function and simple tests.

paulfd:
This one looks OK to me now, if you want to undraft and merge it

jpcima:
This one, I've come to wonder if really it's a good idea, after dealing with ARIA hdcc.

ARIA has used its own extensions to deal with this problem: hdcc and icurvecc.
This is going to modify this behavior in possibly undesired ways.

This commit permits conforming to some Cakewalk behavior, but then it disturbs normal behavior of these ARIA extensions, it's the problem. (also possibly things as on_lohdcc)
It sits in the middle of both ARIA and CW but in a not very satisfying way.

paulfd:
You mean that since ARIA does not care to center at 64 we maybe should just use the extensions for "precise" centering?

kinwie:
If this is not finalized yet, I suggest to do ARIA, hdcc=0.5. I'm guessing, centering Midi value 64 might disturb the curve?

PR #222, opened 2020-05-12
Expand vars aria

jpcima:
First attempt to fix @redtide's Naked Drums set
Will add some tests later.

### DIFF
--- a/src/sfizz/parser/Parser.cpp
+++ b/src/sfizz/parser/Parser.cpp
@@ -457,21 +457,25 @@ std::string Parser::expandDollarVars(const SourceRange& range, absl::string_view
             std::string name;
             name.reserve(64);
 
-            while (i < n && isIdentifierChar(src[i]))
+            // ARIA: we will accumulate any chars after $, until this is the
+            //       name of a known variable
+            auto def = _currentDefinitions.end();
+            while (i < n && isIdentifierChar(src[i]) && def == _currentDefinitions.end()) {
                 name.push_back(src[i++]);
+                def = _currentDefinitions.find(name);
+            }
 
             if (name.empty()) {
                 emitWarning(range, "Expected variable name after $.");
                 continue;
             }
 
-            auto it = _currentDefinitions.find(name);
-            if (it == _currentDefinitions.end()) {
+            if (def == _currentDefinitions.end()) {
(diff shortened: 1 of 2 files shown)


jpcima:
@paulfd Where does the below test come from?
This fails to load in Sforzando, for a reason which I'm not yet fully sure.

Based on some tests I made on my side:
Sforzando will use the shortest name after '$' which matches the name of a known variable.
This behavior would contradict the test case.

(of course the fact this doesn't load in Sfzorzando means there's more to this, but for us it will be OK to raise it as warning and go on permissively)

```
#define $KeyCrash2Edge     52 // Crash Cymbal 2 (Edge)
#define $KeyCrash2     57 // Crash Cymbal 2 (Bow)
#define $Param     10 // Opcode parameter replacement

<region> key=$KeyCrash2Edge sample=../mono_sample.wav
<region> key=$KeyCrash2 sample=../mono_sample.wav
<region> amplitude_oncc$Param=34 sample=../mono_sample.wav
```

jpcima:
Test case above is confirmed incorrect according to Sforzando.

To make it load one must change:
`#define $KeyCrash2     57 // Crash Cymbal 2 (Bow)`
to
`#define $KeyCrash2 57 // Crash Cymbal 2 (Bow)`

It seems to not like multiple spaces after define (and this one only, most curiously).
After this change, both regions will be on the key 57.

paulfd:
From these files

[GSCW-DrumKit-1.zip](https://github.com/sfztools/sfizz/files/4623912/GSCW-DrumKit-1.zip)

@redtide gave them to me.

The spaces I have to admit I am not sure, I may have seen this somewhere else though... The book says nothing. And it does make sense to strip the whitespace no?

jpcima:
ARIA does not left-trim after `#define`, more in #225

paulfd:
Ok, I think we can be pragmatic and factual. The book does very little to specify the behavior of defines, and ARIA clearly is the target against which most library are built today so for this type of things I think we go their way.

When everything is wrapped up we should consider documenting the behavior on sfzformat to help people building libraries be mindful of this and not attempt to be too crazy with this facility.

jpcima:
Closing, and I will redo this better in new PRs.

PR #223, opened 2020-05-12
Changes to the worker messages

paulfd:
Checking for file changes or loading disables future file change checks
The worker only responds to reactivate the file change checks
Upon parameter change, the new value is stored immediately so only one message is sent to the worker
Removed the changing_state bool and let the sfizz internal lock protect the state

### DIFF
--- a/lv2/sfizz.c
+++ b/lv2/sfizz.c
@@ -145,8 +145,7 @@ typedef struct
     int num_voices;
     unsigned int preload_size;
     sfizz_oversampling_factor_t oversampling;
-    // TODO: use atomic flags
-    volatile bool changing_state;
+    volatile bool check_modification;
     int max_block_size;
     int sample_counter;
     float sample_rate;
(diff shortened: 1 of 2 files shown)


paulfd:
I have to go now but it seems to work more or less. Less pingponging...

One thing that could make sense is moving to lv2 patch messages for all parameters. Before, the internal value for e.g. `num_voices` was only updated after it was changed, so if you had multiple concurrent changes the run method would send many messages to the worker until the voices were indeed changed. This could have created an overflow.

jpcima (approved):

jpcima:
To observe just a thing, not in relation with this PR.

The enumerated value (1,2,4,8) does not work well in a variety of LV2 hosts.
I recall Qtractor to have problems controlling these, and also in Carla, the knob will try to send intermediates 3,5,6,7 and this logs errors in sfizz.

Perhaps it can be taken as a linear integer range?

paulfd:
And rounded? The issue is that internally this oversampling uses x2 stages so it has to be 1 2 4 8. Maybe we don't need to log errors on intermediate values though is that what you mean?

jpcima:
Rounded is fine. Then it should be rounded up to the next power,
such that the obtained oversampling amount is greater or equal to the amount requested.

PR #226, opened 2020-05-13
Residual tiny quality of life changes in the LV2 plugin

paulfd:
- Oversampling parameter changes in LV2 takes the next power of 2 but does not upper bound the value
- Added the "fast path" check within sfizz (i.e. don't change if the current value is the same as the new one) for preload size, oversampling and polyphony. Easier and more encompassing to do it there rather than in the LV2, and it benefits everyone.
- Cosmetic changes for C-style struct initialization on atoms.
- Updated the docs.

### DIFF
--- a/lv2/sfizz.c
+++ b/lv2/sfizz.c
@@ -553,13 +553,11 @@ static int next_pow_2(int v)
 static void
 sfizz_lv2_check_oversampling(sfizz_plugin_t* self)
 {
-    if (*self->oversampling_port < 0)
-        return;
-
-    int port_value = (int)*self->oversampling_port;
+    int port_value = next_pow_2((int)*self->oversampling_port);
     if (port_value == (int)self->oversampling)
         return;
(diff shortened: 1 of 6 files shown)


lv2/sfizz.c
```diff
@@ -534,23 +534,43 @@ sfizz_lv2_status_log(sfizz_plugin_t *self)
...
+    v++;
+    return v;
+}
+
 static void
 sfizz_lv2_check_oversampling(sfizz_plugin_t* self)
 {
-    sfizz_oversampling_factor_t oversampling = (sfizz_oversampling_factor_t)*self->oversampling_port;
-    if (oversampling != self->oversampling)
+    if (*self->oversampling_port < 0)
```
jpcima: Maybe apply next-pow-2 first on the port value, then compare?
and this makes this condition superfluous, given the negative cases is already handled in next-pow-2?

paulfd: Yeah I was a bit wary since well, it looks like more operations but I guess it's not a big problem.

PR #227, opened 2020-05-14
Add some status messages regarding the AU build

jpcima:
Attempt to figure out what went wrong with macOS CI


jpcima:
Seems it was a cached archive in travis causing the problem.
Hopefully this doesn't reproduce in future builds, otherwise remove the cache entry `vst/download`.

PR #230, opened 2020-05-14
parser: dollar expansions and multiple #define on the same line

jpcima:
This fixes some parsing problems, preventing some real files being read correctly.
Tests included.

- in case of a possibly ambiguous `$` expansion, the shorter existing variable name wins (ARIA)
- support of `#define $foo 1 #define $bar 2` all on the same line
  ~~this is done by factorizing the code which extracts the opcode value with look-ahead, and this does the same work in presence of `#define`~~
  did the same as ARIA, which consists in extracting only the first word

### DIFF
--- a/src/sfizz/parser/Parser.cpp
+++ b/src/sfizz/parser/Parser.cpp
@@ -295,7 +295,7 @@ void Parser::processOpcode()
             bool stop = false;
 
             // consume space characters following
-            while (i < valueSize && isSpaceChar(valueRaw[endPosition + 1]))
+            while (i < valueSize && isSpaceChar(valueRaw[i]))
                 ++i;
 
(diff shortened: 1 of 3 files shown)


jpcima:
- Identified a regression
- fixed, and added the matching test case

jpcima:
The last commit is an addition that fixes some cases found in Salander Grand Piano where directives and opcode values are all on the same line.

Example
```
<group> #define $VEL v1 lovel=1 hivel=26 #include "Data/region.txt"
```

jpcima:
Fixed a small mistake. apologies for spam
Parsing sfz is delicate and always new annoyances are found, but that's no secret at this point.

Salamander and Naked Drums should both work (together with the other small side PRs)

Issue #232, opened 2020-05-14
Implement set_hdccN


jpcima:
Required for Naked Drums to initialize the volumes

A minimal example
```
<control>
set_hdcc16=0.5
<master>
amplitude_oncc16=200
<region>
sample=*sine
```

PR #233, opened 2020-05-14
Support the opcode set_hdccN

jpcima:
#232

Support `set_hdccN` and the equivalent `set_realccN` + test


jpcima:
Ok so this PR went slightly overboard, I'll sum it up.
This is a set of additions to improve ARIA compatibility.

- case sensitivity #234 handled in the cleanup
- support `set_lohdccN`/`set_hihdccN` and aliases
- support `lohdccN`/`hihdccN`
- support `start_lohdccN`/`start_hihdccN` and aliases
- applied cleanups to all opcodes everywhere (was `<region>` only)
- all relevant tests

jpcima:
These new opcodes are tested against ARIA, working as expected.
I merge it. This adds support of Naked Drums which needs `set_hdcc` to initialize its volumes.

PR #238, opened 2020-05-16
Polynomial resampler

jpcima:
This implements 2 polynomial interpolators, adds  `sample_quality` and sets 2 as default.
1=linear, 2=polynomial.

The B-spline variant is used since it seems to make the alias decay more and faster.
With SSE active, the degradation of performance is about 4x compared to linear.
(scalar takes approximately 8x)

```
2020-05-16 16:44:44
Running ./build/benchmarks/bm_interpolators
Run on (8 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 64K (x4)
  L2 Unified 512K (x4)
  L3 Unified 4096K (x1)
Load Average: 2.25, 1.89, 1.44
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
Interpolators/Linear/4            9.14 ns         9.05 ns     64813831
Interpolators/Linear/16           36.2 ns         35.9 ns     19443079
Interpolators/Linear/64            142 ns          141 ns      4922180
Interpolators/Linear/256           576 ns          569 ns      1249803
Interpolators/Linear/1024         2243 ns         2220 ns       313092
Interpolators/Linear/4096         9165 ns         9067 ns        78192
Interpolators/Hermite3/4          40.9 ns         40.5 ns     17008264
Interpolators/Hermite3/16          169 ns          168 ns      4165492
Interpolators/Hermite3/64          660 ns          653 ns      1087134
Interpolators/Hermite3/256        2587 ns         2557 ns       274573
Interpolators/Hermite3/1024      10356 ns        10255 ns        64380
Interpolators/Hermite3/4096      41167 ns        40765 ns        17159
Interpolators/Bspline3/4          39.2 ns         38.8 ns     18303277
Interpolators/Bspline3/16          156 ns          154 ns      4534226
Interpolators/Bspline3/64          641 ns          634 ns      1118829
Interpolators/Bspline3/256        2440 ns         2415 ns       258462
Interpolators/Bspline3/1024       9969 ns         9864 ns        72287
Interpolators/Bspline3/4096      40142 ns        39676 ns        18327
```

### DIFF
--- a/benchmarks/BM_interpolators.cpp
+++ b/benchmarks/BM_interpolators.cpp
@@ -4,6 +4,7 @@
 // license. You should have receive a LICENSE.md file along with the code.
 // If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
 
+#include "Config.h"
 #include "Interpolators.h"
 #include "ScopedFTZ.h"
 #include "absl/types/span.h"
@@ -46,7 +47,8 @@ static void doInterpolation(absl::Span<const float> input, absl::Span<float> out
 {
     const float kOutToIn = static_cast<float>(input.size()) / output.size();
 
-    for (size_t iOut = 0; iOut < output.size(); ++iOut) {
+    for (size_t iOut = sfz::config::excessFileFrames;
+            iOut < output.size() - sfz::config::excessFileFrames; ++iOut) {
         float posIn = iOut * kOutToIn;
         unsigned dec = static_cast<int>(posIn);
         float frac = posIn - dec;
@@ -81,6 +83,6 @@ BENCHMARK_DEFINE_F(Interpolators, Bspline3)(benchmark::State& state)
     }
 }
 
-BENCHMARK_REGISTER_F(Interpolators, Linear)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
-BENCHMARK_REGISTER_F(Interpolators, Hermite3)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
-BENCHMARK_REGISTER_F(Interpolators, Bspline3)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
+BENCHMARK_REGISTER_F(Interpolators, Linear)->RangeMultiplier(4)->Range(1 << 4, 1 << 12);
+BENCHMARK_REGISTER_F(Interpolators, Hermite3)->RangeMultiplier(4)->Range(1 << 4, 1 << 12);
+BENCHMARK_REGISTER_F(Interpolators, Bspline3)->RangeMultiplier(4)->Range(1 << 4, 1 << 12);
--- a/benchmarks/BM_interpolators.cpp
+++ b/benchmarks/BM_interpolators.cpp
@@ -20,13 +20,13 @@ public:
         std::uniform_real_distribution<float> dist { -1.0f, 1.0f };
 
         const size_t numFramesIn = state.range(0);
-        inputBuffer = std::vector<float>(numFramesIn + excessFrames);
+        inputBuffer = std::vector<float>(numFramesIn + 2 * sfz::config::excessFileFrames);
 
         // any ratio will do, compute time will be proportional
         static constexpr float ratio = 1.234;
 
         const size_t numFramesOut = static_cast<size_t>(std::ceil(numFramesIn * ratio));
-        input = absl::MakeSpan(inputBuffer.data(), numFramesIn);
+        input = absl::MakeSpan(inputBuffer).subspan(sfz::config::excessFileFrames, numFramesIn);
         output = std::vector<float>(numFramesOut);
         std::generate(input.begin(), input.end(), [&]() { return dist(gen); });
     }
@@ -47,8 +47,7 @@ static void doInterpolation(absl::Span<const float> input, absl::Span<float> out
 {
     const float kOutToIn = static_cast<float>(input.size()) / output.size();
 
-    for (size_t iOut = sfz::config::excessFileFrames;
-            iOut < output.size() - sfz::config::excessFileFrames; ++iOut) {
+    for (size_t iOut = 0; iOut < output.size(); ++iOut) {
         float posIn = iOut * kOutToIn;
         unsigned dec = static_cast<int>(posIn);
         float frac = posIn - dec;
--- a/benchmarks/BM_interpolators.cpp
+++ b/benchmarks/BM_interpolators.cpp
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
+// If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
+
+#include "Interpolators.h"
+#include "ScopedFTZ.h"
+#include "absl/types/span.h"
+#include <benchmark/benchmark.h>
+#include <random>
+
+class Interpolators : public benchmark::Fixture {
+public:
+    void SetUp(const ::benchmark::State& state)
+    {
+        std::random_device rd { };
+        std::mt19937 gen { rd() };
+        std::uniform_real_distribution<float> dist { -1.0f, 1.0f };
+
+        const size_t numFramesIn = state.range(0);
+        inputBuffer = std::vector<float>(numFramesIn + excessFrames);
+
+        // any ratio will do, compute time will be proportional
+        static constexpr float ratio = 1.234;
+
+        const size_t numFramesOut = static_cast<size_t>(std::ceil(numFramesIn * ratio));
+        input = absl::MakeSpan(inputBuffer.data(), numFramesIn);
+        output = std::vector<float>(numFramesOut);
+        std::generate(input.begin(), input.end(), [&]() { return dist(gen); });
+    }
+
+    void TearDown(const ::benchmark::State& /* state */)
+    {
+    }
+
+    std::vector<float> inputBuffer;
+    absl::Span<float> input;
+    std::vector<float> output;
+
+    enum { excessFrames = 8 };
+};
+
+template <sfz::InterpolatorModel M>
+static void doInterpolation(absl::Span<const float> input, absl::Span<float> output)
+{
+    const float kOutToIn = static_cast<float>(input.size()) / output.size();
+
+    for (size_t iOut = 0; iOut < output.size(); ++iOut) {
+        float posIn = iOut * kOutToIn;
+        unsigned dec = static_cast<int>(posIn);
+        float frac = posIn - dec;
+        output[iOut] = sfz::interpolate<M>(&input[dec], frac);
+    }
+}
+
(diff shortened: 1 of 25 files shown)


FrnchFrgg:
Why not use something like https://kokkinizita.linuxaudio.org/linuxaudio/zita-resampler/resampler.html ? This is a very good resampling library, AFAICT. Good enough that it is used by Ardour for internal resampling when the internal engine doesn't run at the same rate as the hardware (currently only used when varispeeding, but in the future it will probably be used when the hardware cannot run at the session rate).

AlexMitchellMus:
IIRC Zita is GPL3, and sfizz is BSDc2. Works one way not the other.

jpcima:
This method is taken for the time being, because of these reasons
- current code is designed for interpolation between neighbour samples, it would take some extra work to redesign it for convolution/polyphase based resamplers
- the interpolator needs to consider only N neighbouring samples per jump; the traditional resampler wants to be fed the entire stream of input samples
- Zita VResampler is variable-rate, but has some documented operational limits "which can variable within a range of 0.95 to 16.0 w.r.t. the originally configured one"
- It lerps a convolution kernel from a large table, and this is not vectorizable at least the current way that it's written
- The FIR convolution is 32 elements recommended minimum for quality result; at lower, I don't know if this will yield any better results than the poly (using the polynomial order 4). The poly is a compromise that tries to preserve good performance wrt. linear.

FrnchFrgg:
> IIRC Zita is GPL3, and sfizz is BSDc2. Works one way not the other.

Oh, too bad.



> This method is taken for the time being, because of these reasons
> 
>     * current code is designed for interpolation between neighbour samples, it would take some extra work to redesign it for convolution/polyphase based resamplers
> 
>     * the interpolator needs to consider only N neighbouring samples per jump; the traditional resampler wants to be fed the entire stream of input samples

I probably misunderstood then. I thought that you were writing about the resampler needed when samples are 44.1kHz and sfizz runs in a 96kHz session.

>     * Zita VResampler is variable-rate, but has some documented operational limits "which can variable within a range of 0.95 to 16.0 w.r.t. the originally configured one"

Ardour uses Zita resampler with factors going from `0.02` to `16`, so it seems that the documentation is overly conservative in the slowdown case. I found that it is a limitation of the stereo resampling. Ardour uses mono (decorrelated) resampling for varispeed. To resample the samples (at loading time ?) it wouldn't matter, because you'd use the non-variable resampler, but if SFZ has some on_cc "playback speed" opcode that's another story.

>     * The FIR convolution is 32 elements recommended minimum for quality result; at lower, I don't know if this will yield any better results than the poly (using the polynomial order 4). The poly is a compromise that tries to preserve good performance wrt. linear.

Ardour uses hlen=12 AFAICT and this already has a very good result. Note that it uses a special mono version, and thus might get away with lower quality too.

None of this matters of course, since zita's license is not compatible.

paulfd:
@FrnchFrgg yes you need on-the-fly resampling to support things like pitch bends and CC-based pitch variations.

src/sfizz/Voice.cpp
```diff
@@ -496,35 +496,35 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
...
-            *left = linearInterpolation(leftSource[*ind], leftSource[*ind + 1], *leftCoeff, *rightCoeff);
-            *right = linearInterpolation(rightSource[*ind], rightSource[*ind + 1], *leftCoeff, *rightCoeff);
-            incrementAll(ind, left, right, leftCoeff, rightCoeff);
-        }
+    const int quality = region->sampleQuality;
+
+    switch (quality) {
+    default:
+        if (quality > 2)
+            goto high; // TODO sinc, not implemented
```
paulfd, line 514: Quick Q as I'm more the communication guy: in this context we would rather that the impulse decays rapidly, so using something like a raised cosine instead of a sinc is something people do in audio?

src/sfizz/Voice.cpp
```diff
@@ -496,35 +496,35 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
...
+        // fall through
+    case 1:
```
paulfd, line 516: I'm not super fond of numbers here but as the sfz spec defines it this way maybe it's better to stick to it. If we decide that we only have a single sinc/whatever in the end, I think it would be clearer and more "sfz-agnostic" to have enums here.

src/sfizz/Voice.h
```diff
@@ -242,6 +243,20 @@ class Voice {
...
+
+    /**
+     * @brief Fill a destination with an interpolated source.
+     *
+     * @param source the source sample
+     * @param dest the destination buffer
+     * @param indices the integral parts of the source positions
+     * @param coeffs the fractional parts of the source positions
+     */
+    template <sfz::InterpolatorModel M>
```
paulfd: we're in the namespace here I think the prefixes can go

paulfd:
I can correct the comments I made btw. If you agree with the previous commits it looks OK to me. As I said on the chat, I will try to think about simplifying some part of the code to avoid ugliness like
```c++
if (dataStatus == DataStatus::Ready)
        return AudioSpan<const float>(fileData)
            .subspan(sfz::config::excessFileFrames,
                fileData.getNumFrames() - 2 * sfz::config::excessFileFrames);
```
which seems error prone for the future.

src/sfizz/FilePool.cpp
```diff
@@ -79,17 +83,19 @@ std::unique_ptr<sfz::AudioBuffer<T>> readFromFile(SndfileHandle& sndFile, uint32
...
-    oversampler.stream(*baseBuffer, *outputBuffer);
+    oversampler.stream(baseSpan, outputSpan);
     return outputBuffer;
 }
 
 template <class T>
 void streamFromFile(SndfileHandle& sndFile, uint32_t numFrames, sfz::Oversampling factor, bool reverse, sfz::AudioBuffer<float>& output, std::atomic<size_t>* filledFrames = nullptr)
 {
-    if (factor == sfz::Oversampling::x1) {
+    if (factor == sfz::Oversampling::x1 && sndFile.format() == SF_FORMAT_WAV) {
```
jpcima: Why introduce this check of format=WAV in this commit?

src/sfizz/Interpolators.hpp
```diff
@@ -68,8 +68,8 @@ class Interpolator<kInterpolatorHermite3, R>
     static inline R process(const R* values, R coeff)
     {
         R y = 0;
-        for (int i = 0; i < 4; ++i) {
-            R h = hermite3<R>(i - 1 - coeff);
+        for (int i = -1; i < 3; ++i) {
```
jpcima, line 71: To offset it by -1, why not, at it looks correct overall. On the other hand, it needs a new prerequisite for use.

BM_interpolators was not updated to work in this new model. It's going to read out-of-bounds at the start of input.

src/sfizz/Voice.cpp
jpcima: I've never seen it before in practice, but a priori, it could work viably as an option.
It's like sinc you would have to tabulate and interpolate it, because of trigonotry that will be involved to compute it.

src/sfizz/FilePool.cpp
paulfd: Oops. I tested something, should not have been in the commit, sorry.

src/sfizz/Interpolators.hpp
paulfd: I added a docstring for the padding and updated the benchmark to heed the excessFileFrames

paulfd:
I'm thinking about the whole handling of padding and bounds and stuff. What do you think about adding padding natively to all AudioBuffers? As a template parameter possibly.

benchmarks/BM_interpolators.cpp
```diff
@@ -46,7 +47,8 @@ static void doInterpolation(absl::Span<const float> input, absl::Span<float> out
 {
     const float kOutToIn = static_cast<float>(input.size()) / output.size();
 
-    for (size_t iOut = 0; iOut < output.size(); ++iOut) {
+    for (size_t iOut = sfz::config::excessFileFrames;
+            iOut < output.size() - sfz::config::excessFileFrames; ++iOut) {
```
jpcima: The fix seems inadequate. The excess count is not supposed to apply to output indices.
Plus would it fix anything in case we'd be upsampling by some higher factor?

The better way seems to over-allocate the input buffer as part of `SetUp`.

src/sfizz/FilePool.h
```diff
@@ -68,12 +68,13 @@ struct FilePromise
             return AudioSpan<const float>(fileData)
                 .subspan(sfz::config::excessFileFrames,
                     fileData.getNumFrames() - 2 * sfz::config::excessFileFrames);
-        else if (availableFrames > preloadedData->getNumFrames())
-            return AudioSpan<const float>(fileData).subspan(sfz::config::excessFileFrames, availableFrames);
+        else if (availableFrames > preloadedData->getNumFrames() + sfz::config::excessFileFrames)
+            return AudioSpan<const float>(fileData)
+                .subspan(sfz::config::excessFileFrames, availableFrames - sfz::config::excessFileFrames);
```
jpcima: Is the use of `availableFrames` correct here, considering that this value does not account for any padding?

I would think this 1st and 3rd line to be correct instead:
```
        else if (availableFrames > preloadedData->getNumFrames() - 2 * sfz::config::excessFileFrames)
```

```
                .subspan(sfz::config::excessFileFrames, availableFrames);
```

paulfd: Actually it does because the streaming function is passed a span, and not the raw buffer, so it should be the real number of available frames. 

The check actually should probably be `availableFrames > preloadedData->getNumFrames() - sfz::config::excessFileFrames`. The reasoning is that you want that the reported number of available frames is `excessFileFrames` less than the actual number of available frames, so that the interpolator will not read beyond the actual end. This was the reasoning of the change I made, but I did not take into accout that `preloadedData->getNumFrames()` has an excess of `2 * sfz::config::excessFileFrames` frames already...

benchmarks/BM_interpolators.cpp
paulfd: Sorry, corrected

paulfd (approved):

jpcima:
I've thought we also need an API to set default quality for both processing modes.

eg. `setDefaultQuality(int realtimeQuality, int freewheelQuality)`

The quality is 1-10 like SFZ, or -1 to use a default setting of sfizz.
The SFZ file can override this with opcode `sample_quality`, or -1 for sfizz current internal setting.

So it selects in this order: 1) sample quality from SFZ file 2) from API 3) from Sfizz defaults

paulfd:
Fine for me!

src/sfizz/FilePool.h
jpcima: I resolved the padding problem at AudioBuffer level instead of FilePool.
This should remove all `excessFileFrames` uglyness everywhere, at keep it as simple as before.

paulfd:
Looks much better indeed, good job!

Issue #229, opened 2020-05-14
Feature request: loops from FLAC tags


jpcima:
Currently it's only WAV and a handful others which can handle loops directly in the sound file.
Flac does not have this ability.

A suggestion is to use tags to store this information.
There isn't an established specification, but there are some relatively established conventions used by a few video game engines.

https://wohlsoft.ru/pgewiki/How_To:_Looping_music_files
https://doomwiki.org/wiki/Audio_loop

Desired also: a tool for converting wav to flac, and preserving the loop using one of these specifications.

[loop.tar.gz](https://github.com/sfztools/sfizz/files/4629659/loop.tar.gz)

kinwie:
Tested the included flac file. The loop seems not function yet

PR #242, opened 2020-05-17
Add the support of FLAC loops

jpcima:
#229
This adds the support of FLAC loops, the same as ARIA.
This is not yet supported in libsndfile: https://github.com/erikd/libsndfile/issues/59.
Hence this uses an ad-hoc implementation based on direct file parsing.

This is NOT the tag-based implementation. Tag-based support may be added as extra in the future.

This uses a feature known as foreign RIFF blocks.
These blocks are obtained by encoding a WAV file by the FLAC command-line encoder, using the flag `--keep-foreign-metadata`.

Example
[FLAC_loop_test.zip](https://github.com/sfztools/sfizz/files/4640300/FLAC_loop_test.zip)

Reference
- https://xiph.org/flac/format.html
- https://sites.google.com/site/musicgapi/technical-documents/wav-file-format#smpl

### DIFF
--- a/src/sfizz/FileInstrument.cpp
+++ b/src/sfizz/FileInstrument.cpp
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
+// If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
+
+#include "FileInstrument.h"
+#include "absl/types/span.h"
+#include <memory>
+#include <cstdio>
+#include <cstring>
+
+namespace sfz {
+
+// Utility: file cleanup
+
+struct FILE_deleter {
+    void operator()(FILE* x) const noexcept { fclose(x); }
+};
+typedef std::unique_ptr<FILE, FILE_deleter> FILE_u;
+
+// Utility: binary file IO
+
+static bool fread_u32le(FILE* stream, uint32_t& value)
+{
+    uint8_t bytes[4];
+    if (fread(bytes, 4, 1, stream) != 1)
+        return false;
+    value = bytes[0] | (bytes[1] << 8) | (bytes[2] << 16) | (bytes[3] << 24);
+    return true;
+}
+
+static bool fread_u32be(FILE* stream, uint32_t& value)
+{
+    uint8_t bytes[4];
+    if (fread(bytes, 4, 1, stream) != 1)
+        return false;
+    value = (bytes[0] << 24) | (bytes[1] << 16) | (bytes[2] << 8) | bytes[3];
+    return true;
+}
+
+/**
+ * @brief Extract the instrument data from the RIFF sampler block
+ *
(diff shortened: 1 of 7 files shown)


src/sfizz/FileInstrument.cpp
```diff
@@ -0,0 +1,143 @@
...
+            char riffId[4];
+            uint32_t riffChunkSize;
+            if (fread(blockId, 4, 1, stream.get()) == 1 && memcmp(blockId, "riff", 4) == 0 &&
+                fread(riffId, 4, 1, stream.get()) == 1 && memcmp(riffId, "smpl", 4) == 0 &&
+                fread_u32le(stream.get(), riffChunkSize) && riffChunkSize <= block_size - 12)
+            {
+                std::unique_ptr<uint8_t[]> chunk { new uint8_t[riffChunkSize] };
+                if (fread(chunk.get(), riffChunkSize, 1, stream.get()) == 1)
+                    return extractSamplerChunkInstrument(
+                        absl::MakeConstSpan(chunk.get(), riffChunkSize), ins);
```
paulfd: You could be fancy here I think with `{ chunk.get(), riffChunkSize }`

src/sfizz/FileInstrument.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        if (block_type == 2) { // APPLICATION block
+            char blockId[4];
+            char riffId[4];
+            uint32_t riffChunkSize;
+            if (fread(blockId, 4, 1, stream.get()) == 1 && memcmp(blockId, "riff", 4) == 0 &&
+                fread(riffId, 4, 1, stream.get()) == 1 && memcmp(riffId, "smpl", 4) == 0 &&
+                fread_u32le(stream.get(), riffChunkSize) && riffChunkSize <= block_size - 12)
+            {
+                std::unique_ptr<uint8_t[]> chunk { new uint8_t[riffChunkSize] };
```
paulfd, line 129: How big is riffChunkSize here? If we ignore values that are too crazy we could stack-allocate this with `std:array` (and get an implicit cast to const span later too).

src/sfizz/FileInstrument.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    return true;
+}
+
+bool FileInstruments::extractFromFlac(const fs::path& path, SF_INSTRUMENT& ins)
+{
+    memset(&ins, 0, sizeof(SF_INSTRUMENT));
+
+#if !defined(_WIN32)
+    FILE_u stream(fopen(path.native().c_str(), "rb"));
```
paulfd: Do we need the `native()` here?

paulfd (approved):

src/sfizz/FileInstrument.cpp
jpcima: The chunk size has a fixed header, followed by arbitrary number of loops, followed by the sampler-specific chunk.

Here I'm fine with alloc'ing this chunk, because this is checked to be no larger than the FLAC metadata block. The block has a size encoded in 24 bits, which makes it 16MB in the worst case.

PR #244, opened 2020-05-17
Add the API to receive hdcc values

jpcima:
This is so synth UIs can manipulate sfizz controllers with full precision.
New API: `sfizz_send_hdcc`

@rghvdberg

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -845,12 +845,17 @@ void sfz::Synth::noteOnDispatch(int delay, int noteNumber, float velocity) noexc
 
 void sfz::Synth::cc(int delay, int ccNumber, uint8_t ccValue) noexcept
 {
+    const auto normalizedCC = normalizeCC(ccValue);
+    hdcc(delay, ccNumber, normalizedCC);
+}
+
+void sfz::Synth::hdcc(int delay, int ccNumber, float normValue) noexcept
+{
     ASSERT(ccNumber < config::numCCs);
     ASSERT(ccNumber >= 0);
-    const auto normalizedCC = normalizeCC(ccValue);
 
     ScopedTiming logger { dispatchDuration, ScopedTiming::Operation::addToDuration };
-    resources.midiState.ccEvent(delay, ccNumber, normalizedCC);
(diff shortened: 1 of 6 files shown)


src/sfizz/Synth.cpp
```diff
@@ -844,13 +844,18 @@ void sfz::Synth::noteOnDispatch(int delay, int noteNumber, float velocity) noexc
...
+
+void sfz::Synth::hdcc(int delay, int ccNumber, float normValue) noexcept
 {
     ASSERT(ccNumber < config::numCCs);
     ASSERT(ccNumber >= 0);
-    const auto normalizedCC = normalizeCC(ccValue);
 
     ScopedTiming logger { dispatchDuration, ScopedTiming::Operation::addToDuration };
-    resources.midiState.ccEvent(delay, ccNumber, normalizedCC);
+    resources.midiState.ccEvent(delay, ccNumber, normValue);
```
paulfd, line 858: The `normalizeCC` function clamps the value. Should we do so here too? Or ignore oob values?

paulfd:
We could argue whether a new name is necessary rather than an overload. I think that it's better to have a different name, since it could be maybe confusing for the value `1`. I believe overload resolution would prefer the integer version in this case?

I have to admit I had in mind to deprecate the `uint8_t` version of the API at some point after the change to float in the internal engine, and let front-ends do their normalization if necessary. VST3 sends normalized CC already IIRC. Keeping both APIs does not make it very clear that we're basically converting the `uint8_t` to float internally.

jpcima:
>We could argue whether a new name is necessary rather than an overload

Overloading is preferably avoided, in general. It introduces selection and masking problems.
This makes the compiler's job harder, and the programmer's job as well, just as you have noted.

src/sfizz/Synth.cpp
jpcima: We should do like ARIA. I think we need to look at behavior and replicate the same.
`set_hdcc200=2.0`

paulfd: It doesn't clamp, so let's not clamp either.

paulfd (approved):

PR #245, opened 2020-05-17
Numbering of regions and voices

jpcima:
The goal is to track the respective index of each voice and region.

In comind modulation work, the indices will be used to map voices and regions to their respective resources. It will also need to track the lifetimes of voices, and allocate or release the Nth voice resources from a modulators pool.

The region index serves to track which modulation sources are to process for a given voice.

### DIFF
--- a/src/sfizz/Region.h
+++ b/src/sfizz/Region.h
@@ -34,8 +34,8 @@ namespace sfz {
  *
  */
 struct Region {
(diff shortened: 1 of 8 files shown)


paulfd:
Sorry I should have commented this one...

paulfd (approved):

PR #246, opened 2020-05-18
Voice lifecycle

jpcima:
Follow up on the numbering PR #245
It's a suggestion to be able to track the lifecycle of voices globally.

The purpose is to allocate and free voice-specific resources in relation with the mod matrix.

### DIFF
--- a/src/sfizz/Region.h
+++ b/src/sfizz/Region.h
@@ -16,6 +16,7 @@
 #include "AudioBuffer.h"
 #include "MidiState.h"
 #include "FileId.h"
+#include "NumericId.h"
 #include "absl/types/optional.h"
 #include <bitset>
 #include <string>
@@ -36,7 +37,7 @@ namespace sfz {
  */
 struct Region {
     Region(int regionNumber, const MidiState& midiState, absl::string_view defaultPath = "")
-    : regionNumber(regionNumber), midiState(midiState), defaultPath(std::move(defaultPath))
+    : id{regionNumber}, midiState(midiState), defaultPath(std::move(defaultPath))
     {
         ccSwitched.set();
 
@@ -49,9 +50,9 @@ struct Region {
     /**
      * @brief Get the number which identifies this region
      */
-    int getIdNumber() const noexcept
+    NumericId<Region> getId() const noexcept
     {
-        return regionNumber;
+        return id;
     }
 
     /**
@@ -246,7 +247,7 @@ struct Region {
      */
     float getGainToEffectBus(unsigned number) const noexcept;
 
-    const int regionNumber { -1 };
+    const NumericId<Region> id;
 
     // Sound source: sample playback
     FileId sampleId {}; // Sample
--- a/src/sfizz/Region.h
+++ b/src/sfizz/Region.h
@@ -35,8 +35,8 @@ namespace sfz {
  *
  */
 struct Region {
-    Region(const MidiState& midiState, absl::string_view defaultPath = "")
-    : midiState(midiState), defaultPath(std::move(defaultPath))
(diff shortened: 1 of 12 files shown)


src/sfizz/Region.h
```diff
@@ -237,6 +245,8 @@ struct Region {
      */
     float getGainToEffectBus(unsigned number) const noexcept;
 
+    const int regionNumber {};
```
paulfd: Since the region still has a default constructor, should we initialize to -1, put an optional, or just disable the default constructor entirely?

src/sfizz/Synth.cpp
```diff
@@ -95,7 +102,8 @@ void sfz::Synth::onParseWarning(const SourceRange& range, const std::string& mes
 
 void sfz::Synth::buildRegion(const std::vector<Opcode>& regionOpcodes)
 {
-    auto lastRegion = absl::make_unique<Region>(resources.midiState, defaultPath);
+    int regionNumber = static_cast<int>(regions.size());
```
paulfd, line 105: Regarding the above comment, the region number could also be an unsigned then

src/sfizz/Voice.h
```diff
@@ -255,12 +279,16 @@ class Voice {
     void setupOscillatorUnison();
     void updateChannelPowers(AudioSpan<float> buffer);
 
+    /**
+     * @brief Modify the voice state and notify any listeners.
+     */
+    void switchState(State s);
+
+    const int voiceNumber {};
```
paulfd: Same comments

src/sfizz/Voice.h
```diff
@@ -31,14 +31,38 @@ class Voice {
...
+    }
+
+    enum class State {
+        idle,
+        playing
+    };
+
+    class StateListener {
+    public:
+        virtual void onVoiceStateChanged(int /*idNumber*/, State /*state*/) {}
```
paulfd: On the same idea, the region stores an active/inactive state. Would it be useful to have listeners for this in your applications or a further one?

paulfd:
In the current process and if I remember correctly, the set of regions is not "immutable" per-say, since we build all of them and *then* we go through all the sample to i) check that they exist, and ii) preload them. This will disrupt the region index thing. If the default constructor of the regions is disabled I guess we could maybe check that the sample exists on creating the region? Or maybe disable the region rather than removing it from the set?

src/sfizz/Voice.h
paulfd: Also thinking some more about it: wouldn't a pointer to `this` be more useful than the id number? or maybe both?

paulfd:
Just a small "up". I would use something like this for the polyphony groups too probably.

jpcima:
On this indexing remark, the idea is to retract this invariant that id=index.

Instead of direct indexing I consider to make it linear or binary search, where the search is initiated at index `id`.
In the very vast majority of times the id and index are identical and it's a direct hit, otherwise the id will be located somewhere in the near neighborhood.

paulfd:
I see! Ok then we can try it like this.

src/sfizz/Voice.h
jpcima: The id can be deduced from the Region/Voice, it's true, but on the other hand voice id number creates less coupling, and I don't think I necessarily require my future work to look into details of Voice or Region. (but that remains a thing to be confirmed, I'm not fully sure)

paulfd (approved):

jpcima:
Rebased, tested, implemented lookup by ID.
This lead into the problem that removed Regions are out of order, due to the iter_swap way of doing.
Substituting with O(n) removal, the order is kept, and search works correctly in the presence of gaps.

PR #253, opened 2020-05-21
Tuning

### LINKED ISSUE #82: Micro-tuning
We can implement the tuning file support thanks to the surge tuning library.
https://github.com/surge-synthesizer/tuning-library

This library is able to accept one or a pair of .scl and .kbm, and it can output and array of MIDI note frequencies to match these settings. The advantage of this particular library is the MIT license.

As LV2 is concerned, an example of manifest can be taken from rncbc's synthv1.lv2.
Parameters are as follows:
- tuning : on/off
- reference note : 69
- reference pitch : 440.0
- scale file
- keymap file

jpcima:
This adds API support and LV2 controls for tuning.

- scala tuning file
- scala root key
- reference 440 frequency

It's similar to Sforzando but with a small difference.
The Sforzando root key designates the top note of the Scala, whereas Scala itself designates the lower note. (and so does sfizz)

It's not perfect because the parameters *root key* and *tuning frequency* are not currently RT-safe.
It's because the Surge tuning library implements the mapping change by std::string and IO streams.
It can be resolved by making some light changes into the library.

Other note: the implementation into sfizz is by retuning a note as it starts.
It's the same as taking the frequency, retuning, and remapping it into MIDI keys.
Then it proceeds with this fractional key (a real), used in relation to the pitch keycenter.

For example: under 432 Hz tuning
A4 note 69  -> fractional note 68.68

### DIFF
--- a/src/sfizz/Tuning.cpp
+++ b/src/sfizz/Tuning.cpp
@@ -161,8 +161,6 @@ Tuning::~Tuning()
 
 bool Tuning::loadScalaFile(const fs::path& path)
 {
-    const Tunings::KeyboardMapping kbm = impl_->tuning().keyboardMapping;
-
     fs::ifstream stream(path);
     if (stream.bad()) {
         DBG("Cannot open scale file: " << path);
@@ -184,8 +182,6 @@ bool Tuning::loadScalaFile(const fs::path& path)
 
 bool Tuning::loadScalaString(const std::string& text)
 {
-    const Tunings::KeyboardMapping kbm = impl_->tuning().keyboardMapping;
-
     std::istringstream stream(text);
 
     Tunings::Scale scl;
--- a/src/sfizz/Tuning.cpp
+++ b/src/sfizz/Tuning.cpp
@@ -1,0 +1,202 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
+// If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
+
+#include "Tuning.h"
+#include "Debug.h"
+#include "Tunings.h" // Surge tuning library
+#include <fstream>
+#include <sstream>
+#include <array>
+#include <cmath>
+
+namespace sfz {
+
+struct Tuning::Impl {
+public:
+    Impl() { updateKeysFractional12TET(); }
+
+    const Tunings::Tuning& tuning() const { return tuning_; }
+
+    float getKeyFractional12TET(int midiKey) const;
+
+    int rootKey() const { return rootKey_; }
+    float tuningFrequency() const { return tuningFrequency_; }
+
+    void updateScale(const Tunings::Scale& scale);
+    void updateRootKey(int rootKey);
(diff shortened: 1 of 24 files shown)


jpcima:
RT safety problems resolved, except some functions we don't use. (eg. `evenDivisionOfSpanByM`)

paulfd:
Unlike the SFZ file, we could use the LV2 abstract_path extension to save the scala file in a portable way. I can add this if you wish.

I wonder if it would be also nice to have a "reload check" also on the scala tuning? If people want to tweak it on the fly. In this case, using the abstract_path may not be a good idea, as it would copy the original file in the session data.

jpcima:
This could mean auto-reloading will be only valid for the current session where the file was picked, and then no longer after reloading, unless the user would to re-select it.

A thing: isn't the  file with virtual path that Ardour keeps just a symlink to the real file?
In this case it should not cause any problems.

paulfd:
Actually it seems like it, but then I don't understand the point? If you move the session on disk or to another computer, the symlink will break?

paulfd:
I'm more in favor of auto-reloading and not using the abstract path thing.

jpcima:
>I'm more in favor of auto-reloading and not using the abstract path thing.

Fine for me

paulfd:
I added the auto-reloading and corrected a couple of small bugs in the LV2 in the process! Looks OK to me now, except I have these warnings:
```c++
/home/paul/source/sfizz-jpcima/src/sfizz/Tuning.cpp: In member function ‘bool sfz::Tuning::loadScalaFile(const std::filesystem::__cxx11::path&)’:
/home/paul/source/sfizz-jpcima/src/sfizz/Tuning.cpp:164:36: warning: variable ‘kbm’ set but not used [-Wunused-but-set-variable]
  164 |     const Tunings::KeyboardMapping kbm = impl_->tuning().keyboardMapping;
```
They're indeed unused, just want to check with you that they're not supposed to be used ;)

jpcima:
Yes it's a leftofver of how I updated the scale which keeping the previoud kbm.
This has been moved into `Impl` when I figured I needed a more complex logic.
OK to remove it.

paulfd (approved):

src/sfizz/Tuning.cpp
```diff
@@ -44,25 +46,59 @@ struct Tuning::Impl {
...
-void Tuning::Impl::updateScale(const Tunings::Scale& scale)
+void Tuning::Impl::updateScale(const Tunings::Scale& scale, absl::optional<fs::path> sourceFile)
 {
     tuning_ = Tunings::Tuning(scale, tuning_.keyboardMapping);
     updateKeysFractional12TET();
+
+    if (sourceFile) {
+        std::error_code ec;
+        modificationTime_ = fs::last_write_time(*sourceFile, ec);
+        scalaFile_ = sourceFile;
```
jpcima: It should assign `scaleFile_` regardless if the condition holds true or not.
It should stop file checking in the case that a file-based scala is replaced with another non-file-based.

paulfd: true

jpcima:
It appears to be good overall.
@alcomposer it's done, feel free to give this a test

I don't have yet a determined a parametric formula for the stretch tuning that seems really adequate, it will be an extra for another PR.

PR #261, opened 2020-05-26
Basis of a common UI for all plugins

jpcima:
This is a basis for creating portable UI.
It's based on Pugl and Elements libraries, it drops VSTGUI entirely, and it's implemented in LV2 and VST.

No proper UI at the moment, only the SliderAndKnobs demo.

EDIT a note to self
- [x] make building the LV2 UI optional, for 2 reasons
    - MOD devices require ui-less plugins
    - the Elements lib requires C++14, the MOD toolchain is still at C++11


paulfd:
This looks nice! I can debug the windows build if it's a mess for you. I have a windows install.

jpcima:
I leave you a link to the current problem. I guess a cairo custom port could fix this.
https://github.com/microsoft/vcpkg/issues/11640

On the status of this thing, I allowed it to work in Windows and Mac in all VST/LV2/AU.
The elements library, it's back ported to C++14 so it can build this in most of configurations.

Pugl has some crashes which I resolved but they're not in upstream.
For reference,
- a Mac crash when closing editor https://github.com/lv2/pugl/pull/33
- a Windows crash after drawing https://github.com/lv2/pugl/pull/34

@redtide works on designing a real UI on the basis of this PR, he makes it in a separate branch.

Next this will need some sort of adapter in between plugin and UI, to fit the behavior with each of plugin formats. This layer should be allowed to communicate in both directions.

lv2/sfizz_ui.cpp
```diff
@@ -0,0 +1,176 @@
+/*
+  SPDX-License-Identifier: ISC
+
+  Sfizz LV2 plugin
+
+  Copyright 2019-2020, Paul Ferrand <paul@ferrand.cc>
```
paulfd, line 6: Hem :D

jpcima:
@paulfd when you're able, you can check if de51bfa fixes the X11 error.
It reproduces uncertainly because of being an apparent race condition involving the display server, but it's easier by slowing down the program. (valgrind comes in handy for this purpose)

jpcima:
A basic UI is done, connected to plugin controls.
The tuning fonctionality is in LV2, but remains to be implemented in VST plugin side.

PR #262, opened 2020-05-29
Asserts

paulfd:
Separate hard and soft assertions, mostly in simd functions.

### DIFF
--- a/src/sfizz/Debug.h
+++ b/src/sfizz/Debug.h
@@ -8,6 +8,7 @@
 
 #if !defined(NDEBUG) || defined(SFIZZ_ENABLE_RELEASE_ASSERT)
 #include <iostream>
+#include <iomanip>
 
 // Break in source code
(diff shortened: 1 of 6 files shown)


src/sfizz/Debug.h
```diff
@@ -39,17 +39,31 @@
...
 
+#define CHECKFALSE                                                              \
+    do {                                                                        \
+        std::cerr << "Check failed at " << __FILE__ << ":" << __LINE__ << '\n'; \
+    } while (0)
+
+#define CHECK(expression)  \
+    do {                   \
+        if (!(expression)) \
+            CHECKFALSE;    \
```
jpcima: Here if I may suggest, the expression may be stringified as `#expression` and displayed in the error.
The same applies to assert.

paulfd: :+1:

jpcima (approved):

PR #263, opened 2020-05-29
GCSW drumkit bugs

paulfd:
- Corrected a bug with sfz v1 velcurve. I made it use the Curve facilities now.
- When you encounter something like `key=64Garbage`, it seems that ARIA actually reads `key=64` and ignores the `Garbage`.

### DIFF
--- a/src/sfizz/Curve.h
+++ b/src/sfizz/Curve.h
@@ -68,6 +68,17 @@ public:
         Interpolator itp = Interpolator::Linear, bool limit = false);
 
     /**
+     * @brief Build a curve based on sfz v1 "amp_velcurve_&" points
+     *
+     * @param points the points vector
+     * @param itp kind of interpolator to fill between values
+     * @param invert whether to invert the curve
+     */
+    static Curve buildFromVelcurvePoints(
+        const std::vector<std::pair<uint8_t, float>>& points,
+        Interpolator itp = Interpolator::Linear, bool invert = false);
+
+    /**
      * @brief Number of predefined curves
      */
     enum { NumPredefinedCurves = 7 };
--- a/src/sfizz/Curve.h
+++ b/src/sfizz/Curve.h
@@ -75,7 +75,7 @@ public:
      * @param invert whether to invert the curve
      */
     static Curve buildFromVelcurvePoints(
-        const std::vector<std::pair<uint8_t, float>>& points,
+        absl::Span<const std::pair<uint8_t, float>> points,
         Interpolator itp = Interpolator::Linear, bool invert = false);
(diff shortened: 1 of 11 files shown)


jpcima:
>When you encounter something like key=64Garbage, it seems that ARIA actually reads key=64 and ignores the Garbage.

Confirmed yes, I was able to make similar observations and it was a reason for disabling the following unit test cases.
https://github.com/sfztools/sfizz/blob/60436277ebc067a1cf6ed679798c2b5011320010/tests/FilesT.cpp#L349-L356

It's quite possible that ARIA uses stdlib.h conversions internally, but I suspect there is more than that.
I recall this from experimentation with CW, when setting an invalid value for a numeric opcode, it's still going to override the value with something (a default?). That's contrary to sfizz which would just ignore the value.
In relation to this, and #235, I'll like to verify, for instance, what happens if value is not only "-1" but also funny things like "-0.1", "-3.14", "foo".

src/sfizz/Curve.h
```diff
@@ -67,6 +67,17 @@ class Curve {
...
 
+    /**
+     * @brief Build a curve based on sfz v1 "amp_velcurve_&" points
+     *
+     * @param points the points vector
+     * @param itp kind of interpolator to fill between values
+     * @param invert whether to invert the curve
+     */
+    static Curve buildFromVelcurvePoints(
+        const std::vector<std::pair<uint8_t, float>>& points,
```
jpcima: Make this parameter a Span?

jpcima (approved):

paulfd:
In this particular case Sforzando read it as 57 doesn't it? At least it seemed so in my test. The issue is that in such cases sfizz does ignore it. The proper test should thus be
```c++
    //       It is the shorter matching $-variable which matches among both.
    //       The rest of the variable name creates some trailing junk text
    //       which Sforzando accepts without warning. (eg. `key=57Edge`)
    REQUIRE( synth.getRegionView(0)->keyRange.getStart() == 57 );
    REQUIRE( synth.getRegionView(0)->keyRange.getEnd() == 57 );
```
This was a crash cymbal that was suppose to be on note 52. The `key` opcode being ignored by sfizz, by default, the sound will play on all the keyboard and be pitch shifted accordingly. Coupled with the above velocity bug that output `0.0f` gain in some cases, you had a veeeeeeeeeeeeery long silent sample.

jpcima:
On the test yes, I recall it's something like that.
The reason that I disabled it, instead of correcting, it's because this changed what the test what meant to illustrate and made it confusing. (the test case was erroneous to start)

PR #264, opened 2020-05-30
Revamp the SIMD helpers

paulfd:
Goals:
- Collapse all the spans at runtime in release builds into a single size parameter, while keeping the bound checks as is in debug builds
- Allow choosing the SIMD level at runtime
- Allow for arbitrary large SIMD operations, making use of `cpuinfo`
- Drop unused/verbose code as much as possible, such as the interpolation helpers and the math functions

For now it seems to work fine, and there are many tests in this area of the code so it's good. There are some cases where it already helps: the `readInterleaved` and `writeInterleaved` scalar code was naturally optimized by the compiler, whereas before it was not for some reason.


jpcima:
If I may also suggest, there is a single header ARM compatibility for SSE intrinsics
https://github.com/jratcliff63367/sse2neon

paulfd:
@jpcima yes I saw it, but I kinda want to dabble a bit in the intrinsics to learn and see :slightly_smiling_face: If I get bored I'll use it ahah.

I know some intrinsics don't really have a straightforward equivalent apparently?

paulfd:
OK, looks good now, and it's kinda ready if you want AVX and NEON operations :slightly_smiling_face:  I'm thinking we could maybe add a way to benchmark and auto-activate the SIMD operations in some API?

jpcima:
If it's desired to add AVX for instance, how is this supposed to be implemented under this model?

paulfd:
I added an example in 8cc9dc3aa3036d890ed6ca37e18eb71dde32bcc7

It's slower than SSE though...

jpcima:
Hmm but you couldn't possibly write direct AVX code in generic compile units, right?
Unless `-march=native` or `-mavx`, this compilation would fail.

paulfd:
I added the flag to the cmake file property, or do you mean something else?

jpcima:
>I added the flag to the cmake file property, or do you mean something else?

Oh that's right, I missed that. Don't do this.
This makes the compiler free to select AVX insts for any ordinary code, and this can produce failure on ordinary processors.

On AVX slowness, there are some hints here and there regarding `vzeroupper`
https://stackoverflow.com/questions/41303780/why-is-this-sse-code-6-times-slower-without-vzeroupper-on-skylake

paulfd:
OK so this need additional preprocessor to disable AVX code? This might get a bit unreadable...

jpcima:
It needs a strict separation AVX code into distinct compile units. (ie. `SIMDHelpersAVX.cpp`)

paulfd:
OK, I'm slightly sad because there's less code removed this way but I separated the routines in different units:
- The `simd` directory has all specializations. The `SIMDHelpers.h` files handles aliases, the `SIMDHelpers.cpp` does the dynamic dispatching.
- The `simd/HelpersScalar.h` has the templated versions
- The `simd/HelpersAVX.*` and `simd/HelpersSSE.*` have the simd versions. This naming should make the DPF makefile set their flags properly.

I also added preprocessor guards in the SIMD versions. If you compile the file without SIMD support and call the function it should give you the proper result.

jpcima:
- what about slowness? should we not disable AVX until resolving the performance issue?
- also on SIMD wrappers, there seems to be too much unnecessary branching going in.
  Imo, it will be better to keep these functions as global pointers and reduce the wrapper into a jump instruction.

paulfd:
- Yes it was more to answer your other question.
- Can you elaborate? Which wrappers are you talking about? The content of SIMDHelpers.cpp?

Jun 2, 2020 15:01:09 JP Cimalando <notifications@github.com>:

jpcima:
it's at this branch, the idea of a minimal CPU dispatch, in the same kind as compiler-generated ones
https://github.com/jpcima/sfizz/tree/simd-dispatch

But it fails in static initializers.. with origin unknown.
This PR also has static initializers and it's possibly a problem as well.

paulfd:
Functionally it works on my machine and I tried some debug messages at the beginning to check for codepaths. I did not have a mutex as you did though which was probably missing. I'm never too sure of anything with static initializations.

On the dispatch topic I see your point now. I can try to work on it maybe.

paulfd:
So the default (static) bipolar Curve uses the simd helper to build a ramp, and this in turns calls the uninitialized (at this point) static variable `simdDispatch`. I guess that's what people mean by pitfalls in static initialization.

If you change the `simdDispatch` from a free static variable to a function
```c++
static SIMDDispatch& simdDispatch()
{
    static SIMDDispatch dispatch;
    return dispatch;
}
```
it seems to work out when using e.g.
```c++
template <>
void add<float>(const float* input, float* output, unsigned size) noexcept
{
    return simdDispatch().add(input, output, size);
}
```
If I understand the problem properly, since the original simdDispatch is a free variable, it will get initialized at some point but not necessarily before the curve. On the other hand, when you put it in a function, it's lazily built the first time the function is called. Is this what happens?

Note: in testing I detemplatized the SIMDDispatch class but that might not be necessary, although for now we only have floats :) I tried with
```c++
template<class T>
static SIMDDispatch<T>& simdDispatch()
{
    static SIMDDispatch<T> dispatch;
    return dispatch;
}
```
and replacing all `simdDispatch` variable calls by `simdDispatch<float>()`. A bit uglier, but seems to work too.

paulfd:
I cherry picked your branch, and put the templates on calls to the `SIMDOpStatus` functions. If we every decide to do doubles we could, otherwise it's only slightly more verbose. Hopefully the tests should pass now!

src/sfizz/SIMDHelpers.cpp
```diff
@@ -0,0 +1,374 @@
...
+{
+    return simdDispatch<float>().diff(input, output, size);
+}
+
+///
+
+static volatile bool simdInitialized = false;
+static std::mutex simdMutex;
+
+SIMDInitializer::SIMDInitializer()
```
jpcima: There's no more need for SIMDInitializer, this can go

src/sfizz/SIMDHelpers.cpp
```diff
@@ -0,0 +1,374 @@
...
+{
+    const unsigned index = static_cast<unsigned>(op);
+    ASSERT(index < simdStatus.size());
+
+    simdStatus[index] = enable;
+
+    const cpuid::cpuinfo& info = cpuInfo();
+    bool useSSE = enable && info.has_sse();
+    // bool useAVX = enable && info.has_avx();
+
```
jpcima: There should be `#if` checks around usage of `useSSE` and `useAVX`,

paulfd:
Ok so I added this, which does (only slightly) reduced the if branches in the `setStatus` function.
```c++
template<class F>
F chainedIfWithDefault(F value)
{
    return value;
}

template<class F, class ...Args>
F chainedIfWithDefault(bool condition, F value, Args... args)
{
    return condition ? value : chainedIfWithDefault(args...);
}
```
It's doing a chain ternary operator in a way.

The issue is that with clang, `F` does not seem to be deduced as `noexcept` before C++17 which enforced this rule, whereas gcc was more permissive but probably wrong in their standard interpretation. I tried a couple things but it gets quite deep into template hell. Unless you have a better solution, I guess we could:
1. Fall back to the original `if (useSSE) ...` blocks.
2. Compile all clang targets (MacOS and tidy mostly) with C++17 enabled.
3. Remove the `noexcept` specification from all helpers.
4. Use some preprocessor macro locally, e.g.
```c++
#define USE_AVX_IF_ENABLED(name) if (useAVX) { name = & name ## AVX; break; }
#define USE_SSE_IF_ENABLED(name) if (useSSE) { name = & name ## SSE; break; }
#define ELSE_USE_SCALAR(name) name = & name ## Scalar<float>;
```
but if 1 setting activates multiple accelerators them this does not work either.

I guess the first solution is not super pretty, but has the merit of being clear and working...

jpcima:
>It's doing a chain ternary operator in a way.

It's just like a chain of ternary so I don't see the point for it.
The function is a more verbose `a ? b : c ? d : e` that forces evaluation of its arguments.

The macro solution, this one can be nice, especially that you can define these macros to empty in ARM case.

Another tiny advice, it's to ensure calling `simdDispatch` on synth init.
The dispatcher instantiates cpuid, and cpuid allocates memory.

paulfd:
Ok tried with macros (my git tree was super messed up). Like this I'm not even sure the macros help so much.

src/sfizz/SIMDHelpers.cpp
```diff
@@ -0,0 +1,307 @@
...
+            SIMD_OP(copy)
+            SIMD_OP(cumsum)
+            SIMD_OP(diff)
+            SIMD_OP(mean)
+            SIMD_OP(meanSquared)
+        }
+#undef SIMD_OP
+    }
+
+#ifdef SFIZZ_HAVE_AVX
```
jpcima: It's wrong. The above checks for -mavx flag to the compiler, which will not be present.
The conditionals should check just the processor type.

paulfd: But then aren't these checks already in cpuinfo? This method will probably never be called in some hot path so the boolean check on `cpuinfo::has_avx()` should be pretty insignificant no?

jpcima: Change to `#if SFIZZ_CPU_FAMILY_X86_64 || SFIZZ_CPU_FAMILY_I386`
There's no need of `SFIZZ_HAVE_AVX` whatsoever, since that ISA is checked at runtime.

paulfd: ok

PR #265, opened 2020-06-03
Prep work for finer polyphony control

paulfd:
I'm thinking I'll break this down a little bit into pieces. The final goal is to handle voice stealing under different policies and at different levels. Ideally in the spec, you should be able to steal because of
- `note_polyphony` at the region level
- `polyphony` limits for regions, groups, (masters) and global
- `group=` in conjunction with `polyphony` at the group level, but also apparently at the region level.
- Global engine limit

This PR adds the following facilities to support stealing and allow for the above:
- The "hierarchy" of groups/masters/... is now built as a view onto the still flat vector of regions. Each region has a parent `RegionSet`, that can contain regions and other region sets. This should make at least the structure amenable to possibly arbitrary nesting of groups within groups. The sets should also track their active voices, using the voice lifecycle addition or the synth updating it upon cleaning up voices that have ended.
- Polyphony groups (for `group=` opcodes) also follow a similar idea in a separate structure.
- A small linked list is added to the voices and set when they're started so that you can iterate over the "sister" voices that were started at the same time. This is needed now and later for flexible voice stealing. The current voice stealing algorithm has been updated to make use of this. When a voice dies, it removes itself from the ring of sister voices.
- I added also a set of "overflow" voices to gather stolen voices that do not have `off_mode=fast` set. The `off_mode=time` and `off_mode=normal` need that the normal release of the voices play. For now I set the number of overflow voices at the same number of normal voices but we can try to see how it behaves in practice. We can also kill some overflow voices if they get out of control the same way we do normal voices now.
- I reduced the default number of voices to 64. Hopefully with the coming improvements it will be a sane baseline and as @FrnchFrgg said it's better to be conservative so that standard settings look good from a CPU usage PoV.

What needs to happen here:
- [x] Add some tests for the polyphony groups (on groups, on regions, etc)

What I plan to do next is:
- [ ] Hook this with the voice lifecycle somehow to update the polyphony properly over all the needed elements.
- [ ] Try to have templated `Voice* stealVoice<VoiceStealingPolicy>(span<Voice*> activeVoices)` functions that provide you with a stolen voice (or null) within a set of active voices. 

Probably some more stuff too. This would be in a further PR.

### DIFF
--- a/lv2/sfizz.ttl.in
+++ b/lv2/sfizz.ttl.in
@@ -129,7 +129,7 @@ midnam:update a lv2:Feature .
     lv2:portProperty pprop:expensive ;
     lv2:portProperty lv2:integer ;
     lv2:portProperty lv2:enumeration ;
-    lv2:default 256 ;
+    lv2:default 64 ;
     lv2:minimum 8 ;
     lv2:maximum 256 ;
     lv2:scalePoint [ rdfs:label "8 voices",
@@ -161,11 +161,6 @@ midnam:update a lv2:Feature .
       "256 voix"@fr ,
       "256 Voci"@it;
       rdf:value 256
-    ] ;
-    lv2:scalePoint [ rdfs:label "512 voices",
-      "512 voix"@fr ,
-      "512 Voci"@it;
-      rdf:value 512
     ] ;
   ] , [
     a lv2:InputPort, lv2:ControlPort ;
--- a/src/sfizz/Config.h
+++ b/src/sfizz/Config.h
@@ -37,8 +37,8 @@ namespace config {
     constexpr bool loggingEnabled { false };
     constexpr size_t numChannels { 2 };
     constexpr int numBackgroundThreads { 4 };
-    constexpr int numVoices { 256 };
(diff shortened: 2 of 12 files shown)


jpcima:
>The "hierarchy" of groups/masters/... is now built as a view onto the still flat vector of regions.
>Polyphony groups (for group= opcodes) also follow a similar idea in a separate structure.
>A small linked list is added to the voices and set when they're started so that you can iterate over the "sister" voices that were started at the same time. 

Seems good

>I added also a set of "overflow" voices to gather stolen voices that do not have off_mode=fast set.

I don't understand yet how this one works. Is it finished implementing?

>I reduced the default number of voices to 64.

Yes, this can be a big deal on small power machines like ARM, and when envelopes have long release.

paulfd:
re: the overflow voices. Not finished now, I just added the structure and rendering. The idea is that stealing algorithms can put voices on "end-of-life" in here beyond the selected polyphony. I plan to refine and use it in the follow-up pull requests.

PR #269, opened 2020-06-12
Fix LV2 problems with the atom sequence

jpcima:
- pop the frame when finished writing
- add safety checks in case write space is short


jpcima:
In the last commit:
- avoid casting a `char[]` into `LV2_Atom*`, which might result in a misalignement
- the null-termination was removed for simplicity. atom forge is already null-terminating buffers, so this one is probably redundant.

Issue #270, opened 2020-06-13
macos sierra 10.12.6 build/link error around align_val_t


hexdump0815:
i just tried to build sfizz myself on macos sierra 10.12.6 (compiler found is AppleClang 9.0.0.9000039) and get the following build/linker errors:
```
[ 49%] Linking CXX shared module ../sfizz.lv2/sfizz.so
undef: __ZnwmSt11align_val_t
undef: __ZdlPvSt11align_val_t
Undefined symbols for architecture x86_64:
  "operator new(unsigned long, std::align_val_t)", referenced from:
      _sfizz_create_synth in libsfizz.a(sfizz_wrapper.cpp.o)
  "operator delete(void*, std::align_val_t)", referenced from:
      _sfizz_create_synth in libsfizz.a(sfizz_wrapper.cpp.o)
      _sfizz_free in libsfizz.a(sfizz_wrapper.cpp.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
any odea how to get this resolved? any compiler options to maybe not add extra alignement stuff?

a lot of thanks in advance and best wishes - hexdump

jpcima:
If I believe the reference document, aligned operator new is available only on c++ version 17 and up.
https://en.cppreference.com/w/cpp/memory/new/operator_new

Perhaps `-DCMAKE_CXX_STANDARD=17` helps to resolve this?

hexdump0815:
i did already use that option - i basically used the cmdline from https://github.com/sfztools/sfizz/blob/develop/.travis/script_osx.sh ... looks like -DCMAKE_CXX_STANDARD=14 helps :)

jpcima:
Most likely it's the effect of a system c++ runtime too old to support this feature.

I don't recall where exactly this alignment consideration comes from, but I recall it is seen in some warning logs.

hexdump0815:
at least the resulting binaries work and this way i'm able to compile sfizz on my system too ...

jpcima:
I downgraded the Travis build to 14, hopefully this lets next release builds to work on older macOS.

PR #271, opened 2020-06-13
Build macOS in C++14, can fix runtime link issues

jpcima:
In macOS Sierra, it's reported that some C++17 features don't have standard library support.
To prevent possible incompatibilities, downgrade the C++ requirement to 14.


hexdump0815:
just to have it here as a reference: for more info see https://github.com/sfztools/sfizz/issues/270

PR #272, opened 2020-06-14
Sister voices

paulfd:
Breaking down the previous PR; sister voice part.

A small linked list is added to the voices and set when they're started so that you can iterate over the "sister" voices that were started at the same time. This is needed now and later for flexible voice stealing. The current voice stealing algorithm has been updated to make use of this. When a voice dies, it removes itself from the ring of sister voices.

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -551,49 +551,38 @@ sfz::Voice* sfz::Synth::findFreeVoice() noexcept
         / static_cast<float>(voiceViewArray.size()) * config::stealingEnvelopeCoeff;
     const auto ageThreshold = voiceViewArray.front()->getAge() * config::stealingAgeCoeff;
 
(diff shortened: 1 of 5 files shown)


src/sfizz/Synth.cpp
```diff
@@ -550,49 +551,39 @@ sfz::Voice* sfz::Synth::findFreeVoice() noexcept
...
+
+        // Jump over the sister voices in the set
+        do { idx++; }
+        while (idx < voiceViewArray.size() && sisterVoices(ref, voiceViewArray[idx]));
     }
-    assert(returnedVoice->isFree());
+
+    auto tempSpan = resources.bufferPool.getStereoBuffer(samplesPerBlock);
+    SisterVoiceRing::applyToRing(returnedVoice, [&] (Voice* v) {
+        std::cout << "Killing a voice" << '\n';
```
jpcima: Remove std::cout

paulfd: :+1:

jpcima (approved):

PR #274, opened 2020-06-14
The synth now resets the voices

paulfd:
The voice does not reset itself anymore, but rather enters a zombie state.
This intermediate state will be used by the synth to update the polyphony counters if needed.

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -708,6 +708,9 @@ void sfz::Synth::renderBlock(AudioSpan<float> buffer) noexcept
             callbackBreakdown.amplitude += voice->getLastAmplitudeDuration();
             callbackBreakdown.filters += voice->getLastFilterDuration();
             callbackBreakdown.panning += voice->getLastPanningDuration();
+
+            if (voice->toBeCleanedUp())
+                    voice->reset();
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/sfizz/Synth.cpp
```diff
@@ -892,6 +895,12 @@ void sfz::Synth::hdcc(int delay, int ccNumber, float normValue) noexcept
         return;
     }
 
+    if (ccNumber == config::allNotesOffCC || ccNumber == config::allSoundOffCC) {
```
jpcima, line 898: Why? registerCC already does this below.
(also `all sound off` gets itself processed out of order with the midi stream, but that's something else)

paulfd: The point was to not let the voice reset itself.

src/sfizz/Voice.cpp
```diff
@@ -139,7 +139,7 @@ void sfz::Voice::release(int delay, bool fastRelease) noexcept
         return;
 
     if (egEnvelope.getRemainingDelay() > delay) {
-        reset();
+        switchState(State::cleanMeUp);;
```
jpcima: double semicolon on this line

jpcima (approved):

PR #275, opened 2020-06-15
Region sets

### LINKED ISSUE #231: Group-level opcodes considered "unknown"
The opcodes handled at group-level, such as `polyphony`, are put in the unknown opcode list, despite that they have been handled at `handleGroupOpcodes`.

~~The opcode "polyphony" appears at <master> level in the Naked Drums set.
Sfizz has a problem with it because it seems only handled at the <group> level.~~

### LINKED ISSUE #243: Drumset voices not playing after the Nth one
This happens with the ARIA mapping of the Naked Drums set.

The following happens: when you play a note 4 times in moderately fast succession, the voice plays, but the next ones don't. One has to wait some seconds until the key is playable again.

This may have to do with a particularity of this set: all samples are 2 seconds long wave files, ending with a silence period. Often, the silence part occupies half the sound length or more.

### LINKED ISSUE #259: Polyphony opcode within <region>
With Sforzando, I can use the following line without issue:

`<region> sample=/Lower8/release/31.wav lokey=g1 hikey=g#1 pitch_keycenter=g1 offset=9594 group=1 polyphony=1`

The `group=x polyphony=1` is useful in particular for preventing build-up of release samples during long trills in baroque keyboard music. As I now understand it, SFZ format specifies polyphony as a group code but clearly Sforzando can also interpret it just fine within a region and how I've seen it used in other instruments.

I now have a potential workaround from Paul by using "master" to create hierarchy enabling use of "group" per release note but it would be great for sfizz to be able to read existing patches that work flawlessly with Sforzando.

paulfd:
- Add the region sets (hierarchy)
- Link the voice lifecycle into region sets and polyphony groups
- Respect crudely the polyphony limits without stealing for now

### DIFF
--- a/src/sfizz/RegionSet.h
+++ b/src/sfizz/RegionSet.h
@@ -16,53 +16,93 @@ namespace sfz
 
 class RegionSet {
 public:
-    void setPolyphonyLimit(unsigned limit)
-    {
-        polyphonyLimit = limit;
-        voices.reserve(limit);
-    }
-    unsigned getPolyphonyLimit() const { return polyphonyLimit; }
-    void addRegion(Region* region)
-    {
-        if (absl::c_find(regions, region) == regions.end())
-            regions.push_back(region);
-    }
-    void addSubset(RegionSet* group)
-    {
-        if (absl::c_find(subsets, group) == subsets.end())
-            subsets.push_back(group);
-    }
-    void registerVoice(Voice* voice)
-    {
-        if (absl::c_find(voices, voice) == voices.end())
-            voices.push_back(voice);
-    }
-    void removeVoice(const Voice* voice)
-    {
-        swapAndPopFirst(voices, [voice](const Voice* v) { return v == voice; });
-    }
-    static void registerVoiceInHierarchy(const Region* region, Voice* voice)
-    {
-        auto* parent = region->parent;
-        while (parent != nullptr) {
-            parent->registerVoice(voice);
-            parent = parent->getParent();
-        }
-    }
-    static void removeVoiceFromHierarchy(const Region* region, const Voice* voice)
-    {
-        auto* parent = region->parent;
-        while (parent != nullptr) {
-            parent->removeVoice(voice);
-            parent = parent->getParent();
-        }
-    }
-    RegionSet* getParent() const { return parent; }
-    void setParent(RegionSet* parent) { this->parent = parent; }
-    const std::vector<Voice*>& getActiveVoices() const { return voices; }
-    std::vector<Voice*>& getActiveVoices() { return voices; }
-    const std::vector<Region*>& getRegions() const { return regions; }
-    const std::vector<RegionSet*>& getSubsets() const { return subsets; }
+    /**
+     * @brief Set the polyphony limit for the set
+     *
+     * @param limit
+     */
+    void setPolyphonyLimit(unsigned limit) noexcept;
+    /**
+     * @brief Add a region to the set
+     *
+     * @param region
+     */
+    void addRegion(Region* region) noexcept;
+    /**
+     * @brief Add a subset to the set
+     *
+     * @param group
+     */
+    void addSubset(RegionSet* group) noexcept;
+    /**
+     * @brief Register a voice as active in this set
+     *
+     * @param voice
+     */
+    void registerVoice(Voice* voice) noexcept;
+    /**
+     * @brief Remove an active voice for this set.
+     * If the voice was not registered this has no effect.
+     *
+     * @param voice
+     */
+    void removeVoice(const Voice* voice) noexcept;
+    /**
+     * @brief Register a voice in the whole parent hierarchy of the region
+     *
+     * @param region
+     * @param voice
+     */
+    static void registerVoiceInHierarchy(const Region* region, Voice* voice) noexcept;
+    /**
+     * @brief Remove an active voice from the whole parent hierarchy of the region.
+     *
+     * @param region
+     * @param voice
+     */
+    static void removeVoiceFromHierarchy(const Region* region, const Voice* voice) noexcept;
+    /**
+     * @brief Get the polyphony limit
+     *
+     * @return unsigned
+     */
+    unsigned getPolyphonyLimit() const noexcept { return polyphonyLimit; }
+    /**
+     * @brief Get the parent set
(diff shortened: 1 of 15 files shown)


src/sfizz/Synth.h
```diff
@@ -668,15 +695,25 @@ class Synth final : public Voice::StateListener, public Parser::Listener {
...
-    std::vector<std::unique_ptr<Voice>> voices;
+    using RegionViewVector = std::vector<Region*>;
+    using VoiceViewVector = std::vector<Voice*>;
+    using VoicePtr = std::unique_ptr<Voice>;
+    using RegionPtr = std::unique_ptr<Region>;
+    using RegionSetPtr = std::unique_ptr<RegionSet>;
+    std::vector<RegionPtr> regions;
+    std::vector<VoicePtr> voices;
+    // These are more general "groups" than sfz and encapsulates the full hierarchy
+    enum class Header { Global, Master, Group };
```
jpcima: This exists already as OpcodeScope found in `Opcode.h`.

src/sfizz/RegionSet.h
```diff
@@ -0,0 +1,66 @@
...
+        if (absl::c_find(voices, voice) == voices.end())
+            voices.push_back(voice);
+    }
+    void removeVoice(const Voice* voice)
+    {
+        swapAndPopFirst(voices, [voice](const Voice* v) { return v == voice; });
+    }
+    static void registerVoiceInHierarchy(const Region* region, Voice* voice)
+    {
+        auto parent = region->parent;
```
jpcima: let's maybe write these as `auto*`?

src/sfizz/Synth.h
paulfd: OK. I added the `master` level. It seems there's no need to handle this anywhere else in the regex generation but can you check?

jpcima: It's not affected, it would be always called with region scope, after all of master/group/region are coalesced into one big array.

paulfd:
Maybe let me put the voice stealing logic within this PR to make it more self contained and I'll undraft it when done!

src/sfizz/VoiceStealing.h
```diff
@@ -0,0 +1,90 @@
...
+    {
+        // Start of the voice stealing algorithm
+        absl::c_sort(voices, voiceOrdering);
+
+        const auto sumEnvelope = absl::c_accumulate(voices, 0.0f, [](float sum, const Voice* v) {
+            return sum + v->getAverageEnvelope();
+        });
+        const auto envThreshold = sumEnvelope
+            / static_cast<float>(voices.size()) * config::stealingEnvelopeCoeff;
+        const auto ageThreshold = voices.front()->getAge() * config::stealingAgeCoeff;
```
jpcima: Is the span of voices assumed to be sorted by age?
Isn't the `swapAndPop` idion at risk to mess with this order?

Btw unrelated, regarding `swapAndPopAll`, since this thing is O(N) anyway, you may as well replace it with order-preserving removal for identical complexity.

paulfd: Yes it would but I sort it at the beginning of the method, so it's fine here.

Not sure I get your point about swapAndPopAll?

paulfd:
I was a bit too fast on the previous review request as some tests were not passing. Sorry.

src/sfizz/VoiceStealing.h
paulfd: I don't swap in here.

jpcima: This statement makes an assumption on the ordering of voices as they are traversed, no ?
`if (ref->getAge() < ageThreshold)`
I don't really understand what this is trying to express.

>Not sure I get your point about swapAndPopAll?

I'll make it clear with an example.

Here below, the `vector::erase` provides a stable sort, but swap-and-pop does not. (1-2-3 vs 1-3-2)
The runtime complexity is exactly the same.

```
    std::vector<int> v;
    auto predicate = [](int x) { return x > 3; };

    // using swap & pop method
    v = { 1, 4, 2, 5, 3 };
    sfz::swapAndPopAll(v, predicate);
    for (int x : v) std::cout << " " << x; std::cout << "\n";

    // using vector::erase method
    v = { 1, 4, 2, 5, 3 };
    auto it = std::remove_if(v.begin(), v.end(), predicate);
    v.resize(std::distance(v.begin(), it));
    for (int x : v) std::cout << " " << x; std::cout << "\n";
```

paulfd: > This statement makes an assumption on the ordering of voices as they are traversed, no ?
> `if (ref->getAge() < ageThreshold)`
> I don't really understand what this is trying to express.

Yes, but the voices are sorted at the beginning of the method (`absl::c_sort(voices, voiceOrdering);`). The `voiceOrdering` predicate sorts by age. So it's ok to assume this order here. 

This condition `ref->getAge() < ageThreshold` avoids killing voices that are too young compared to the rest. I'll check if it's entirely needed in the newer versions, because this was in early in my attempts to improve this algorithm.

> I'll make it clear with an example.
>
> Here below, the vector::erase provides a stable sort, but swap-and-pop does not. (1-2-3 vs 1-3-2)
The runtime complexity is exactly the same.

Just to be sure, this discussion is indeed unrelated to the actual stealing method?

The complexity in terms of comparisons is the same but the number of swaps would be higher with remove_if, unless I'm missing something? The remove_if version *could* have the same number of swaps if it goes through the vector and then resolves the swaps to do after, but i) this is an effect of having only 5 elements in the example, and ii) can we count on remove_if implementations to operate like this?

jpcima: >Yes, but the voices are sorted at the beginning of the method (absl::c_sort(voices, voiceOrdering);). The voiceOrdering predicate sorts by age. So it's ok to assume this order here.

Ok I see. Maybe it makes sense to make use of a policy pattern for these?
I recall this pattern to be present in some places in boost libraries.

I don't find where this is, but something along the lines of
`void search(T x, already_sorted_t) // the fast implementation`
`void search(T x, not_already_sorted_t = {}) // the slower one`

>Just to be sure, this discussion is indeed unrelated to the actual stealing method?

Yes it's unrelated

jpcima (approved):

Issue #67, opened 2020-02-18
Optimize random generator


jpcima:
Below is a number generator, fast and branchless, in few instructions.
It has a comparison with standard library generators.
https://godbolt.org/z/r_TccA

As a note, the std generator is particularly bad when using clang as the compiler, for some reason.

paulfd:
MSVC is pretty bad too...

I'm good with this, I'd like to bench it against the std::normal after the box-muller transform or do you have a faster method for this? 

On a side note, visually, the white noise I generate now looks quite different from sforzando's. I need to plot some histograms to understand what's going on.

jpcima:
I think the question to ask as: does it sound the same? if so, fine enough.

Visually, I have run a plot as I implemented and the result was like this.
![random](https://user-images.githubusercontent.com/17614485/74747058-a06f1e00-5266-11ea-9163-13139fec8568.png)

paulfd:
I'm moving this up to the next milestone for the SIMD version and the Gaussian one.

jpcima:
It's funny that you comment this now, I was researching gauss and I have a decent one at hand.
I'm going to attach the plots.

jpcima:
There is a distribution which matches c++ `normal_distribution` with various values of variance.
It's nothing original, it's `no.gnoise` of faust, I just make a plot from it.
It accepts a quality settings that ranges from 2 (looks triangle-ish) to any integer values.

![gauss_0 25](https://user-images.githubusercontent.com/17614485/85234774-1f8c8280-b410-11ea-9433-19c671ac949e.png)
![gauss_0 5](https://user-images.githubusercontent.com/17614485/85234783-287d5400-b410-11ea-84ee-54214fc524cd.png)
![gauss_0_75](https://user-images.githubusercontent.com/17614485/85234795-36cb7000-b410-11ea-9d7a-8eeca51808c0.png)

PR #278, opened 2020-06-16
Add the fast random number generator

jpcima:
#67
It's 4x faster in the benchmark.

```
2020-06-16 12:41:32
Running ./build/benchmarks/bm_random
Run on (8 X 3600 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 64K (x4)
  L2 Unified 512K (x4)
  L3 Unified 4096K (x1)
Load Average: 3.20, 1.92, 1.13
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
RandomFill/StdRandom/4                  16.0 ns         16.0 ns     41844831
RandomFill/StdRandom/16                 63.8 ns         63.6 ns     10990813
RandomFill/StdRandom/64                  256 ns          255 ns      2748246
RandomFill/StdRandom/256                1029 ns         1023 ns       685498
RandomFill/StdRandom/1024               4094 ns         4079 ns       171171
RandomFill/StdRandom/4096              16361 ns        16299 ns        42905
RandomFill/StdRandomBipolar/4           16.4 ns         16.3 ns     42879739
RandomFill/StdRandomBipolar/16          65.4 ns         65.2 ns     10710480
RandomFill/StdRandomBipolar/64           261 ns          260 ns      2689280
RandomFill/StdRandomBipolar/256         1045 ns         1042 ns       671687
RandomFill/StdRandomBipolar/1024        4200 ns         4176 ns       168138
RandomFill/StdRandomBipolar/4096       17089 ns        16937 ns        41836
RandomFill/FastRandom/4                 4.35 ns         4.33 ns    160934278
RandomFill/FastRandom/16                17.4 ns         17.3 ns     40407215
RandomFill/FastRandom/64                70.4 ns         70.0 ns     10030776
RandomFill/FastRandom/256                278 ns          277 ns      2524776
RandomFill/FastRandom/1024              1115 ns         1111 ns       630578
RandomFill/FastRandom/4096              4463 ns         4447 ns       157879
RandomFill/FastRandomBipolar/4          4.37 ns         4.36 ns    160553950
RandomFill/FastRandomBipolar/16         17.5 ns         17.4 ns     40205255
RandomFill/FastRandomBipolar/64         69.8 ns         69.6 ns     10021214
RandomFill/FastRandomBipolar/256         279 ns          278 ns      2505383
RandomFill/FastRandomBipolar/1024       1115 ns         1111 ns       627295
RandomFill/FastRandomBipolar/4096       4447 ns         4431 ns       157962
```

### DIFF
--- a/benchmarks/BM_random.cpp
+++ b/benchmarks/BM_random.cpp
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
+// If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
+
+#include "MathHelpers.h"
+#include <benchmark/benchmark.h>
+#include <vector>
+#include <random>
+
+class RandomFill : public benchmark::Fixture {
+public:
+  void SetUp(const ::benchmark::State& state) {
+    output = std::vector<float>(state.range(0));
+  }
+
+  void TearDown(const ::benchmark::State& state) {
+      UNUSED(state);
+  }
+
+  std::vector<float> output;
+};
+
+BENCHMARK_DEFINE_F(RandomFill, StdRandom)(benchmark::State& state) {
+    std::minstd_rand prng;
+    std::uniform_real_distribution<float> dist(0.0f, 1.0f);
+
+    for (auto _ : state)
+    {
+        for (float &out : output)
+            out = dist(prng);
+    }
+}
+
+BENCHMARK_DEFINE_F(RandomFill, FastRandom)(benchmark::State& state) {
+    fast_rand prng;
+    fast_real_distribution<float> dist(0.0f, 1.0f);
+
+    for (auto _ : state)
+    {
+        for (float &out : output)
+            out = dist(prng);
+    }
+}
+BENCHMARK_DEFINE_F(RandomFill, StdRandomBipolar)(benchmark::State& state) {
+    std::minstd_rand prng;
+    std::uniform_real_distribution<float> dist(-1.0f, 1.0f);
+
+    for (auto _ : state)
+    {
+        for (float &out : output)
+            out = dist(prng);
+    }
+}
+
+BENCHMARK_DEFINE_F(RandomFill, FastRandomBipolar)(benchmark::State& state) {
+    fast_rand prng;
+    fast_real_distribution<float> dist(-1.0f, 1.0f);
+
+    for (auto _ : state)
+    {
+        for (float &out : output)
+            out = dist(prng);
+    }
+}
+
+BENCHMARK_REGISTER_F(RandomFill, StdRandom)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
+BENCHMARK_REGISTER_F(RandomFill, StdRandomBipolar)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
+BENCHMARK_REGISTER_F(RandomFill, FastRandom)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
+BENCHMARK_REGISTER_F(RandomFill, FastRandomBipolar)->RangeMultiplier(4)->Range(1 << 2, 1 << 12);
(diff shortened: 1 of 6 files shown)


paulfd:
OK cool I was doing the same thing xD I was trying to get an sse version going too. Found https://software.intel.com/content/www/us/en/develop/articles/fast-random-number-generator-on-the-intel-pentiumr-4-processor.html but I don't really understand why they have all the masking and shuffling going on? It seems to me that the shuffling and masking is essentially a no-op.

Can't we just go like
```c++
const auto sentinel = output + size;
        const auto* lastAligned = prevAligned<4>(sentinel);
        while (unaligned<4>(output) && output < lastAligned) {
            seed *= 1664525u; 
            seed += 1013904223u;
            *output++ = seed;
        }

        const uint32_t _seed [4] { seed, seed + 1, seed + 2, seed + 3 };
        static const uint32_t _mult[4] { 214013, 17405, 214013, 69069 };
        static const uint32_t _add[4] { 2531011, 10395331, 13737667, 1 };
        auto mmAdd = _mm_load_si128( (__m128i*) _add);
        auto mmMult = _mm_load_si128( (__m128i*) _mult);
        auto mmSeed = _mm_load_si128( (__m128i*) _seed);

        while (output < lastAligned) {
            mmSeed = _mm_mul_epu32( mmSeed, mmMult );
            mmSeed = _mm_add_epi32( mmSeed, mmAdd);
            _mm_store_si128((__m128i*) output, mmSeed);
            incrementAll<4>(output);
        }
        // Replace _mm_store_si32 which doesn't exist in all *mmintrin.h headers apparently
        _mm_store_ss((float*)&seed, _mm_castsi128_ps(mmSeed));

        while (output < sentinel) {
            seed *= 1664525u; 
            seed += 1013904223u;
            *output++ = seed;
        }
```

jpcima:
I thought the SIMD could implemented in a way that is rather straightforward.
Instantiate 4 of the generators and run them in parallel.

Now if you wish the generator to not disturb the sequence after misalignment, shuffle the seeds as necessary.
For instance you want to get 2 out of 4 slots (a b c d): compute slots (a b) and reorder as (c d a b)

---
As another note: the generator is static in the header, so each source file which refers it may get its own instance. Perhaps it's wiser to make an instance of PRNG for every site that needs it?

jpcima:
Also be wary of constants of the intel article, because this is for a random of 16-bit resolution (like windows `rand()`). The random of this PR is intended for 32bit full range.

paulfd:
Ok I understand why they had to do the shuffling like this. If you check out `_mm_mul_epu32(...)` it operates on the lower 32 bit part of the 2 64 bits elements in the 128 bit vector, and store the multiplication result in 64 bits. The masking/shuffling is thus necessary since you operate on 2 x 64 bits rather than 4 x 32 bits. 

I think the constants are OK for 32 bits, no? What else is there in these linear congruential generators? It seems that different constants might not mix the bit very well but for our target this may not be such an issue.

The SSE solution is basically doing what you say, running 4 generators in parallel over different constants of a common corrupted seed. Since we're not running crypto it is probably quite ok to just proceed like this. As long as it looks uniform, and the white noise seems white, I'd say we're good 🙂 

I think the key operation will be going from the uniform to gaussian. From a cursory glance, rejection methods might be faster overall but non-deterministic in nature since you might reject and branch alot to get proper results. Tabulating the operations needed for a Box-Muller transform or an inverse CDF might be the cleanest option. We probably can even clamp the distribution to limit the size of the table.

jpcima:
>why they had to do the shuffling like this.

ok it makes sense.

>I think the constants are OK for 32 bits, no?

This was just a remark out of the blue, I haven't given it much thought.
The benefit of the particular LCG generator used, it's modulo = 2^32, which permits omitting the operation on the 32 bit integer.

>I think the key operation will be going from the uniform to gaussian.

That I didn't think of it but, why not.
Also a common audio thing it's to pass white noise in a LPF to create another kind, such as pink.

paulfd:
Regarding the static variable in header and overall organization, what would you think about putting the random generators in the resources? The resource set is already passed and shared in the Voice and Regions who are the primary users of random elements. This way we can seed and initialize what we need along with the other resources upon instantiating a Synth.

jpcima:
>what would you think about putting the random generators in the resources?

It's fine.

PR #283, opened 2020-06-20
Load a default SFZ if path empty

jpcima:
This loads the sine by default, or when the sfz file parameter is empty.

### DIFF
--- a/lv2/sfizz.c
+++ b/lv2/sfizz.c
@@ -88,6 +88,10 @@
 #define LV2_DEBUG(...)
 #endif
 
+static const char default_sfz_text[] =
+    "<region>sample=*sine" "\n"
+    "ampeg_attack=0.02 ampeg_release=0.1" "\n";
+
 typedef struct
 {
     // Features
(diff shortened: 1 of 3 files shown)


paulfd:
What about putting a small envelope on the default? e.g. `ampeg_attack=0.02 ampeg_release=0.1` or something reasonable?

jpcima:
I'd rather keep it the simplest.
For later, I thought we can collect some example synth-based patches and present them under the new ui.
It's as a way to exhibit some of sfizz sfz features.

paulfd:
My concern is that it might click on release mostly, but we have some default envelopr values so that may not be a problem. The smoother can also help for this.

Jun 20, 2020 16:10:29 JP Cimalando <notifications@github.com>:

jpcima:
Resolved. In preparation of a release, I've prepared this PR on the suggestion of @alcomposer.
Also I did this well needed vst update regarding tuning.
The smoothers are looking ok except that I didn't give this a lot of tests.
Do you wish to add the polyphony improvements? regarding #275 and what may come next

paulfd:
I think I can *at least* port the current voice stealing to the new structure of #275 but I'm gonna have a busy couple of weeks I realize. One goal of #275 is to have common voice stealing algorithms regardless of why you need to voice steal (i.e. which polyphony limit you hit), and this should be a small PR on top of #275 . Further goals included finer stealing algorithms and the ability to select the algorithm at runtime, overflow voices and the like but this can probably wait for a further release :slightly_smiling_face:

paulfd (approved):

PR #286, opened 2020-06-22
Fast gauss distribution

### LINKED ISSUE #67: Optimize random generator

jpcima:
This adds the gaussian generator.

A method to implement is documented in faust libraries.
The idea is according to the [central limit theorem](https://en.wikipedia.org/wiki/Central_limit_theorem).
When summing multiple uniform random generators, the output tends to become increasingly normally distributed.

Decent results happen from 3 and up. The code uses 4, because under this magic number, a smart compiler is able to produce SSE introductions by itself. ie. not MS it seems, anyway: benchmarks are below

The unit test is pretty dumb. It compare against the normal distribution from STL.
Preferably it should compare against an actual Gauss PDF, but couldn't get the math straight at this hour, anyway it should be working as such.

```
RandomFill/StdNormal/4                   256 ns          254 ns      2757457
RandomFill/StdNormal/16                 1026 ns         1020 ns       687949
RandomFill/StdNormal/64                 4089 ns         4064 ns       172202
RandomFill/StdNormal/256               16354 ns        16253 ns        43106
RandomFill/StdNormal/1024              65636 ns        65208 ns        10600
RandomFill/StdNormal/4096             261903 ns       260315 ns         2694
RandomFill/FastNormal/4                 9.90 ns         9.84 ns     71188996
RandomFill/FastNormal/16                38.3 ns         38.1 ns     18389591
RandomFill/FastNormal/64                 155 ns          154 ns      4595975
RandomFill/FastNormal/256                603 ns          600 ns      1160145
RandomFill/FastNormal/1024              2397 ns         2383 ns       293549
RandomFill/FastNormal/4096              9576 ns         9515 ns        73787
```

### DIFF
--- a/tests/RandomHelpersT.cpp
+++ b/tests/RandomHelpersT.cpp
@@ -8,6 +8,12 @@
 #include "sfizz/MathHelpers.h"
 #include <vector>
 
+template <class T>
+static inline T squared(T x)
+{
+    return x * x;
+}
+
 /**
  * @brief Check the behavior of a uniform real random generator
  *
@@ -15,7 +21,7 @@
  * - ensure to have at least one element in every division of the result range
  */
 template <class Rand, class Dist>
-static bool randomTest(float min, float max, unsigned gen, unsigned div)
+static bool uniformRandomTest(float min, float max, unsigned gen, unsigned div)
 {
     fast_rand prng;
     fast_real_distribution<float> dist(min, max);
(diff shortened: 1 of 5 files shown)


paulfd:
I would never have expected this method to be a contender ahah, it's very nice.

tests/RandomHelpersT.cpp
```diff
@@ -41,10 +47,98 @@ TEST_CASE("[Random] Fast random generation")
...
+
+    // the tested bounds
+    const double min = -1.0 + mean;
+    const double max = +1.0 + mean;
+
+    // generate, quantify, count occurrences
+    std::vector<size_t> counts(histSize);
+    for (size_t i = 0; i < numGen; ++i) {
+        double value = gen();
+        double normalized = (value - min) / (max - min);
```
paulfd, line 80: Since the normal distribution has unbounded support I am not sure that this normalization makes alot of sense. This value could get negative (with low probability depending on the variance and mean, but still).

tests/RandomHelpersT.cpp
```diff
@@ -41,10 +47,98 @@ TEST_CASE("[Random] Fast random generation")
...
+        long bin = std::lround(histSize * normalized);
+        if (bin >= 0 && static_cast<size_t>(bin) < histSize)
```
paulfd, line 82: Ah ok you reject the negative bins here. Nvm.

tests/RandomHelpersT.cpp
```diff
@@ -41,10 +47,98 @@ TEST_CASE("[Random] Fast random generation")
...
+
+    return true;
+}
+
+TEST_CASE("[Random] Gaussian random generation")
+{
+    unsigned numGenerations = 16384;
+    unsigned numDivisions = 128;
+
+    const double maxAbsErr = 0.05; // PDF ±5%
```
paulfd, line 139: This is one of my issues with probability tests. This could always fail in some cases, especially since we have both the observed and expected frequencies as random. I wonder if we should make this test more "human based", e.g. plotting histograms and deeming the results "good enough"?

jpcima: The generator is performed always with the same seeds, so this does not run at risk that successive runs of the same tests will fail. If it happens by some chance, increase the number of random picks.
It's already chosen to be some magnitudes above the number of bins.

paulfd: OK!

paulfd (approved):

PR #287, opened 2020-06-22
Wavetables hq

jpcima:
Support various qualities of oscillator playback.

- 0: nearest
- 1: linear
- 2: high
- 3: dual-high

example: `<region> sample=*saw oscillator_quality=3`

The quality dual-high interpolates from a pair of adjacent tables.
In this case, a fractional table position is looked up from a table.

The wavetables will allocate extra elements before.
This is to accomodate with the polynomial interpolator.

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -65,19 +65,21 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, float value, 
             wave = resources.wavePool.getWaveSaw();
             break;
         }
-        const int quality = region->oscillatorQuality ? *region->oscillatorQuality : Default::oscillatorQuality;
+        const float phase = region->getPhase();
+        const int quality = region->oscillatorQuality.value_or(Default::oscillatorQuality);
         for (WavetableOscillator& osc : waveOscillators) {
             osc.setWavetable(wave);
-            osc.setPhase(region->getPhase());
+            osc.setPhase(phase);
             osc.setQuality(quality);
         }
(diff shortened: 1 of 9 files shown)


src/sfizz/Voice.cpp
```diff
@@ -65,16 +65,20 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, float value,
             wave = resources.wavePool.getWaveSaw();
             break;
         }
+        const int quality = region->oscillatorQuality ? *region->oscillatorQuality : Default::oscillatorQuality;
```
paulfd: I think you can use the `optional` function `region->oscillatorQuality.value_or(Default::oscillatorQuality)` here. Does the same thing, and slightly less verbose.

src/sfizz/Voice.cpp
```diff
@@ -65,16 +65,20 @@ void sfz::Voice::startVoice(Region* region, int delay, int number, float value,
...
         for (WavetableOscillator& osc : waveOscillators) {
             osc.setWavetable(wave);
             osc.setPhase(region->getPhase());
+            osc.setQuality(quality);
         }
         setupOscillatorUnison();
     } else if (region->oscillator) {
         const WavetableMulti* wave = resources.wavePool.getFileWave(region->sampleId.filename());
+        const int quality = region->oscillatorQuality ? *region->oscillatorQuality : Default::oscillatorQuality;
```
paulfd: id

paulfd (approved):

PR #288, opened 2020-06-22
Fix unused variables warnings

redtide:


src/sfizz/OpcodeCleanup.re
```diff
@@ -30,7 +30,7 @@ static std::string cleanUpOpcodeName(absl::string_view rawOpcode, OpcodeScope sc
     const char* YYCURSOR;
     const char* YYMARKER;
     const char* yypmatch[2 * YYMAXNMATCH];
-    size_t yynmatch;
+    //size_t yynmatch;
```
jpcima: This one, it gets used or not depending on parser rules, which I don't remember it exactly.
I'll rather have it as `UNUSED()` instead of commented.
Also an observation, `./scripts/generate_re2c.sh` should be called after editing this one, and post the cpp file changes as well.

redtide: IDK much how this works, if it has some sort of automation, just simply based on code that it was unused anywhere, so I simply commented it.

jpcima: `OpcodeCleanup.cpp` is generated source based on this file.
In this case, it's fine to include `Macros.h` in the RE file, put `UNUSED(yynmath)` in there, generate cpp, and commit the whole.

jpcima:
On the LV2 one, I prefer to keep it because it reminds we should preferably implement these size checks.
If atom events are arriving for foreign sources, they are not guaranteed to be well formed.

PR #292, opened 2020-06-22
Add the sfz flattener in-tree

paulfd:
Previous version (with the old parser) at https://github.com/sfztools/sfz-flat.


jpcima:
May this program be a duplicate of devtools/Preprocessor, or it does something else ?

paulfd:
It is a bit of a duplicate, except for the option to flatten the file. I forgot about these devtools...

jpcima:
If this offers better functionality, it's fine to replace the existing with this, and rename if desired.

paulfd:
The functionality is mostly the same and this one has the option to add defines so it's nice. No worries.

PR #293, opened 2020-06-23
Scheduling

jpcima:
This increases the running priority of background threads, so it would react faster on note playback.
Some parts are rewritten to use semaphore, to ensure these high-prio threads are not going to spin.

One thing to make sure I get it right:
If a promise enters the list `toClear`, is it correct that it can't be anything else than `Ready` state?
(hence the condition below is redundant and safe to remove)
```
void sfz::FilePool::tryToClearPromises()
{
    const std::lock_guard<std::mutex> promiseLock { promiseGuard };

    for (auto& promise: promisesToClear) {
        if (promise->dataStatus != FilePromise::DataStatus::Wait)
            promise->reset();
    }
}
```

### DIFF
--- a/src/sfizz/Config.h
+++ b/src/sfizz/Config.h
@@ -99,6 +99,10 @@ namespace config {
     static constexpr double amplitudeTriangle = 0.625;
     static constexpr double amplitudeSaw = 0.515;
     static constexpr double amplitudeSquare = 0.515;
+    /**
+       Background file loading
+     */
+    static constexpr int backgroundLoaderPthreadPriority = 50; // expressed in %
 } // namespace config
 
 } // namespace sfz
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -37,8 +37,14 @@
 #include "absl/memory/memory.h"
 #include <algorithm>
 #include <memory>
-#include <sndfile.hh>
 #include <thread>
+#include <system_error>
+#include <sndfile.hh>
+#if defined(_WIN32)
+#include <windows.h>
+#else
+#include <pthread.h>
+#endif
 
 void readBaseFile(SndfileHandle& sndFile, sfz::FileAudioBuffer& output, uint32_t numFrames, bool reverse)
 {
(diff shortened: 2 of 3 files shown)


paulfd:
I think this condition is there to not "reset" the promises twice. Maybe with the semaphore barrier there is no way the background thread would wake up twice before the high priority thread posts in the barrier in your setup?

But yes you're correct the promises going in this vector should not be waiting.

jpcima:
>there is no way the background thread would wake up twice before the high priority thread posts in the barrier in your setup?

A priori I don't think so, but this will be fine.
I wished to make sure that a request which gets into `promisesToClear` will be good to reset right away, and not after some later change of its `dataStatus`. Otherwise this means the `semClearingRequest.post()` is insufficient and should also be posted from other locations which modify `dataStatus`.

If the purpose of the condition is as you said, the change should be OK.

paulfd:
No this should be fine. And there's really no drawbacks to resetting twice except inefficiency in the worst case -- and the inefficiency is in a background thread.

Jun 23, 2020 13:04:39 JP Cimalando <notifications@github.com>:

jpcima:
Ok. For context, this is the start of a work to improve the ogg slowness situation.
Yesterday I received a report of an ogg sample library which is too slow to load.

I hope this will accelerate in case there would be CPU consuming side tasks.

As the file streaming is concerned, perhaps we should read the sound file in chunks relative to the filesystem block size?
and atomic-add the chunk size repeatedly into `availableFrames`. (in case of ogg or large file)

paulfd:
Yes the oversampling case does this already. It should be extended to the ogg files! These are very very slow...

Jun 23, 2020 13:29:04 JP Cimalando <notifications@github.com>:

jpcima:
The oversampling case does nothing to help because it still decodes the entire file in one go.
I'll see about fixing that, but I'll make that job a separate PR.

paulfd:
Yes I mostly meant that the oversampling code updated the sample counter chunk per chunk, as would a chunked reader. For FLAC and wav I seem to remember some cost to reading a file chunked, there's a benchmark somewhere. However in the interest of simpler code maybe it's ok to take a small hit for wav files but read all files chunk by chunk.

Jun 23, 2020 16:36:53 JP Cimalando <notifications@github.com>:

paulfd (approved):

PR #298, opened 2020-07-01
Add support for "sustain_cc=" and "trigger=release_key"

### LINKED ISSUE #68: Differentiate `release` and `release_key`
We need some research and behavior check but definitely now we're not properly handling the difference between these 2 opcodes since they're always checked together.

paulfd:
This should implement the proper behavior regarding the sustain pedal and the `trigger=release/release_key` opcode value.

Closes #68


paulfd:
This will require further architectural work, so I'm delegating the finer behavior to the next release for now.

jpcima:
I must admit to be a bit confused at the `noteIsOff` state.
this flag is suppsed to indicate a note which is on sustain ?
it's normal that `noteIsOff` is present in both Voice and Region?

paulfd:
Yes both flags are set at the same time normally but with different effects let's say. The one in the region is for notes that did not start yet but will start on pedal down. The one in the voice is the opposite. I guess they could use the same flag, with the Voice calling the Region one. 

Overall I'm not an entire fan of the "region state" handling but I'm failing to come up with an overall better plan. In many cases it's kind of hard to separate what should be stored into the region object, the midi state, or the voice itself.

PR #304, opened 2020-07-04
lv2: default files for sfz and scala

### LINKED ISSUE #296: Sine default instrument inconveniences
When a music project was designed for an older sfizz, and it contains unused instances without a SFZ file assigned, they are now going to play as sine wave. This has been reported as an inconvenience.

In this case, let's not make the parameter empty value equivalent of the default patch, but use something else.

jpcima:
This implements a different strategy for file parameters.
It introduces files `DefaultInstrument.sfz` and `DefaultScale.scl`.

By this strategy,
- these 2 files will be loaded initially, and it will store their paths in the instance
- the state loading will not consider the empty string as special case
- as consequence, empty states for old versions are not going to load as sine
- an empty string will not be stored, this should prevent the `state.ttl` issue from happening in Ardour

Closes #296


paulfd:
Closes #296

paulfd (approved):

lv2/sfizz.c
```diff
@@ -975,8 +968,8 @@ restore(LV2_Handle instance,
 
         lv2_log_note(&self->logger, "[sfizz] Restoring the file %s\n", path);
         // Load an empty file to remove the default sine
-        sfizz_lv2_get_empty_sfz_path(instance, self->sfz_file_path, MAX_PATH_SIZE);
-        sfizz_lv2_load_file(instance, self->sfz_file_path);
+        strcpy(self->sfz_file_path, "");
```
jpcima: There seems to be an inconsistency about this decision to reset here.
The sfz gets reset, but not the scala down below.

Moreover it's precisely what this implementation would attempt to avoid before; always keep a non-empty string in there, to work around the problem with Ardour.

The string remains empty if the following load fails.
Perhaps we could store the paths, regardless if the file loading succeeds or not? (+ it's ensuring plugin and host parameters side to be in sync)

paulfd: We can reset the scala too.

Well, this keeps the "behavior" of the previous version. Ardour will through state.ttl at us but we won't make sounds, so previously empty instances remain empty.

You're suggesting that we use the `state.ttl` file as the plugin path value in this case? It might be an indication for the user that something is wrong, so it's not a bad solution. In this case we need to change the `sfz_file_path` variable regardless of whether loading works.

jpcima: We can reset absolutely everything. `restore()` could be as
- set all values to defaults
- extract the parameters present in the object into values
- sync all values into the synth

jpcima:
In the commit a29b04e, the following changes are made
- `bundle_path` must be copied if it's supposed to be used outside of `instantiate`
- `restore` is going to reset missing properties to defaults. This means a future version can retain compatibility with an older state. In the present case, older states would miss the scala file, so this should reset it to 12TET scale.
- the sfz and scala file parameters will be stored, regardless whether loading succeeds or fails

paulfd:
Nice.
- I proposed a correction for the `snprintf` warning in 74e53ca.
- For the default sine and for failed restored files, I disabled the periodic check in 8e4c3df

jpcima:
>For the default sine and for failed restored files, I disabled the periodic check

What about other call sites of `sfizz_lv2_load_file`, should it not be disabled there also?

Also I just recall, it's no longer necessary to reset with `empty.sfz` after this change set, is it? sfizz `loadSfzFile` already does `clean` by itself.

>I proposed a correction for the snprintf warning

Ok, do you mind that I change this later into dynamically alloced buffers? (in other PR)

paulfd:
The other call sites seems to be in the worker thread. The process in this case is:
1. Disable the check in the RT thread
2. Ask for a file loading in the background
3. If the file loading works, the check is reenabled by the RT thread by sending an atom back and switching it on in `work_response`.

At the time this was done to avoid multiple threads jumping on this value (even if it's volatile). Can we assume that `restore` happens before calls on the RT thread?

paulfd:
> Ok, do you mind that I change this later into dynamically alloced buffers? (in other PR)

Sure :smiley:

jpcima:
There is something said in LV2 regarding `restore` and threads.
http://lv2plug.in/ns/ext/state#threadSafeRestore

Thread-safe restore is an opt-in, by default the host does not use `restore` and audio processing concurrently.
From the description, I'm not sure if the constraint will apply to worker threads.

However, it tells that it's possibly to send the worker a message, and do the job of restoring in the work thread.

paulfd:
In this case we can send an atom and let the Worker do it's thing 😄

Jul 5, 2020 14:30:19 JP Cimalando <notifications@github.com>:

paulfd:
Ah wait I read it the wrong way. So if we say nothing the restore function will not be called concurrently with RT functions?

If so, then it's pretty fine for us no? Is there a big interest in implementing `threadSafeRestore`?

jpcima:
`threadSafeRestore` means the RT thread runs, default means it doesn't.
Is there a big interest? I don't know. I wanted to just make a point that this permits to concentrate the loading code into the worker.

The documentation specificies that the RT should be active or not depending the option.
But AFAIK it was not explicit regarding the worker. I don't know if `restore` and worker can be concurrent actors.

Anyway, I don't have a problem that we merge this PR in current state.

paulfd:
Technically all these calls should be thread-safe because they'll disable the callback and they're mutex-protected. In this sense, I believe we could activate the `threadSafeRestore` as-is and still be OK. I do like to concentrate the code, but in this case I am not sure. Except for this `check_modification` boolean, it's pretty clear what's happening now, whereas using a worker in another thread obfuscates the process a little bit. In terms of code complexity you'd have to build Atoms, which takes space.

I am a bit torn but I'd be tempted to merge it this way.

PR #312, opened 2020-07-05
Added include files in CMakeLists as support for QtCreator IDE

### LINKED ISSUE #310: List the header files in cmake target sources
This is based on a remark by @redtide from some time ago.
The header files don't appear in the file hierarchy of Qt creator. It's needed to list them in cmake sources.

redtide:
See #310.

Edit by Paul: Closes #310


devtools/CMakeLists.txt
```diff
@@ -8,11 +8,11 @@ endif()
...
 if(JACK_FOUND AND TARGET Qt5::Widgets)
-    add_executable(sfizz_capture_eg CaptureEG.cpp)
+    add_executable(sfizz_capture_eg CaptureEG.h CaptureEG.cpp)
     target_include_directories(sfizz_capture_eg PRIVATE . ${JACK_INCLUDE_DIRS})
     target_link_libraries(sfizz_capture_eg PRIVATE sfizz-sndfile Qt5::Widgets ${JACK_LIBRARIES})
     set_target_properties(sfizz_capture_eg PROPERTIES AUTOUIC ON)
 endif()
 
-add_executable(sfizz_preprocessor Preprocessor.cpp)
+add_executable(sfizz_preprocessor ../tests/cxxopts.hpp Preprocessor.cpp)
```
paulfd: Maybe this one is not so relevant, since it's basically an external library?

lv2/CMakeLists.txt
```diff
@@ -11,7 +11,13 @@ set (LV2PLUGIN_TTL_SRC_FILES
     manifest.ttl.in
     ${PROJECT_NAME}.ttl.in
 )
-add_library (${LV2PLUGIN_PRJ_NAME} MODULE ${PROJECT_NAME}.c ${LV2PLUGIN_TTL_SRC_FILES})
+source_group("Turtle Files" FILES
```
paulfd, line 14: This creates groups in the Qt project?

paulfd (approved):

paulfd:
I think in a developer doc somewhere we need to document where to add the sources when we add new files, since there's now:
- The various CMakeFiles lying around
- The DPF-oriented Makefile
- The `clang-tidy` static checks

For example I realized I missed some of them recently :D

jpcima:
There's at least `SfzFilterImpls.cxx` missing which is user-written, not generated.
This file is particular in this sense that it's not among the compiled ones but included.
It can't be put direct into the sources array, so this must be managed as special case. (not sure how best to do this, it may be put in a different source group)

jpcima:
@paulfd clang-tidy is possible to interoperate directly with cmake, have you checked that?

devtools/CMakeLists.txt
redtide: Mh, that's right.

lv2/CMakeLists.txt
redtide: Yes, by default QtCreator creates a "Source Files" and a "Header Files" groups by default based on the files listed in the CMakeLists.txt, that one would be a custom group to keep other file types separated.

paulfd:
@jpcima no not really. We could also have a clang-format check somewhere in there btw.

devtools/CMakeLists.txt
redtide: Fixed now.

jpcima:
Also the faust files are good to have.
There are some in
- ./src/sfizz/dsp/filters/
- ./src/sfizz/effects/dsp/

jpcima (approved):

vst/CMakeLists.txt
```diff
@@ -42,8 +51,14 @@ plugin_add_vst3sdk(${VSTPLUGIN_PRJ_NAME})
 plugin_add_vstgui(${VSTPLUGIN_PRJ_NAME})
 
 # Add the ring buffer
+set (HEADER_FILES
```
jpcima: can this be renamed as `RINGBUFFER_HEADERS`?

Issue #19, opened 2019-12-28
Virtual Playing Orchestra bugs


michaelwillis:
Yesterday I compiled the sfizz lv2 from the `develop` branch and tried using [Virtual Playing Orchestra](http://virtualplaying.com/virtual-playing-orchestra/). There are bugs; I'm not sure if they are due to VPO using non-standard sfZ opcodes, or if sfizz is missing features that VPO depends on.

Any of the VPO sfz files with `PERF` in the file name causes sfizz to crash immediately upon playing a single note.

The pizzicato articulations aren't played correctly, they make a very brief sound and then go silent.

There are possibly other issues, these are just the ones I noticed while doing a cursory test.

paulfd:
Thanks for the report I will try these!

michaelwillis:
To clarify, when I say "crash", I mean it caused Ardour to quit with a segfault. Thanks for taking a look!

paulfd:
Hi Michael, I looked into it a bit and pushed in some changes (try e.g. 3608b3145aa6d3a411994a45622743156c6010e8). Not guaranteeing that this whole sfz bank is perfect but it should be a bit better. I'll leave this open, don't hesitate to post back all the weird behavior you see so I can track it down :smiley:

michaelwillis:
Hey @paulfd, thanks for taking a look at this. I'll build the latest and let you know how it goes. Admittedly, VPO is a large and very elaborate collection of virtual instruments, so I'll likely find some other weirdness here and there. I'll report more thorough results as I work on overhauling my orchestra template to use sfizz.

michaelwillis:
I just gave sfizz+vpo a more thorough test. From what I can tell, SFizz now performs the VPO instruments as well as LinuxSampler does, which is great news.

There is one minor bug that I also found in LinuxSampler: The instruments with cross-faded dynamics don't work. I think the problem is that VPO has those instruments set up such that the mod wheel (CC1) controls both volume and cross-fading multiple samples. See this page for more details about the cross-fading feature: http://virtualplaying.com/dynamic-cross-faded-brass/

Edit: To clarify, when I tried to test cross-fading, I adjusted the mod-wheel controller while playing a sustained note. The volume adjusted accordingly, but the cross fade didn't happen.

paulfd:
Thanks, this is indeed great news :)

Yes, the crossfade calculation are not dynamic right now and they're only checked when the note start. After reading around, this is indeed not what's expected so I'll look into it!

[...]

PR #313, opened 2020-07-05
Crossfade smoothers

paulfd:
This adds a crossfade smoother, regarding previous reports in #153 #22 #19 . There was an improvement with the global gain smoothing in #181 which was reverted in #309 because applying it globally on the gain had undesirable side-effects on e.g. drums with big transients.

This PR:
- Adds a "shortcut" in all smoothers. A smoother will just copy the span if i) the input span to smooth seems to be basically constant, and ii) the first value of the span is the same as the current state of the smoother.
- Add a separate process for the crossfades.

I'll evaluate the performance loss/gain.

@jpcima When doing this modulation matrix I think it'd be good to find a clean way to shortcut as much as possible. Let's discuss this further 🙂

### DIFF
--- a/src/sfizz/Smoothers.cpp
+++ b/src/sfizz/Smoothers.cpp
@@ -27,10 +27,25 @@ void Smoother::reset(float value)
 
 void Smoother::process(absl::Span<const float> input, absl::Span<float> output)
 {
-    if (smoothing)
+    CHECK_SPAN_SIZES(input, output);
+    if (input.size() == 0)
+        return;
+
+    const auto midValue = input[input.size() / 2];
+    const bool shortcut = (
+        input.front() == input.back()
+        && input.front() == midValue
+        && input.front() == current()
+    );
+
+    if (smoothing && !shortcut) {
         filter.processLowpass(input, output);
-    else
+    }
+    else if (input.data() != output.data()) {
         copy<float>(input, output);
+    } else {
+        // Nothing to do
+    }
 }
 
 }
--- a/src/sfizz/Smoothers.cpp
+++ b/src/sfizz/Smoothers.cpp
@@ -25,26 +25,21 @@ void Smoother::reset(float value)
     filter.reset(value);
 }
 
-void Smoother::process(absl::Span<const float> input, absl::Span<float> output)
+void Smoother::process(absl::Span<const float> input, absl::Span<float> output, bool canShortcut)
 {
     CHECK_SPAN_SIZES(input, output);
     if (input.size() == 0)
         return;
 
-    const auto midValue = input[input.size() / 2];
-    const bool shortcut = (
-        input.front() == input.back()
-        && input.front() == midValue
-        && input.front() == current()
-    );
+    if (canShortcut && std::abs(input.front() - current()) < config::virtuallyZero) {
+        if (input.data() != output.data())
+            copy<float>(input, output);
 
-    if (smoothing && !shortcut) {
+        filter.reset(input.back());
+    } else if (smoothing) {
         filter.processLowpass(input, output);
-    }
-    else if (input.data() != output.data()) {
+    } else if (input.data() != output.data()) {
         copy<float>(input, output);
-    } else {
-        // Nothing to do
     }
 }
 
--- a/.travis/install_osx.sh
+++ b/.travis/install_osx.sh
@@ -4,6 +4,7 @@ set -ex
 
 sudo ln -s /usr/local /opt/local
 brew update
+brew upgrade python ||  brew link --overwrite python
 brew upgrade cmake
 brew install jack
 brew install dylibbundler
--- a/src/sfizz/Config.h
+++ b/src/sfizz/Config.h
@@ -41,6 +41,7 @@ namespace config {
     constexpr int numVoices { 64 };
     constexpr unsigned maxVoices { 256 };
     constexpr unsigned smoothingSteps { 512 };
+    constexpr uint8_t xfadeSmoothing { 5 };
     constexpr uint8_t gainSmoothing { 0 };
     constexpr unsigned powerTableSizeExponent { 11 };
     constexpr int maxFilePromises { maxVoices };
--- a/src/sfizz/OnePoleFilter.h
+++ b/src/sfizz/OnePoleFilter.h
@@ -25,6 +25,8 @@ class OnePoleFilter {
 public:
     OnePoleFilter() = default;
 
+    Type current() const { return state; }
+
     void setGain(Type gain)
     {
         G = gain / (1 + gain);
--- a/src/sfizz/Smoothers.h
+++ b/src/sfizz/Smoothers.h
@@ -45,6 +45,7 @@ public:
      */
     void process(absl::Span<const float> input, absl::Span<float> output);
 
+    float current() const { return filter.current(); }
 private:
     bool smoothing { false };
     OnePoleFilter<float> filter {};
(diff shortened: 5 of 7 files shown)


paulfd:
It seems to work OK, and does not impact performance too much.

https://box.ferrand.cc/sfizz-reports/2020-07-06-xfade-history.html

Showcasing the perf-tracking work in progress :) The 3 big hits in performance in recent history were due to:
1. The voice stealing algorithm that uses an envelope follower on every voice.
2. The resampler going from linear to polynomial by default
3. The global smoother

Item 2. is not really negotiable and can be user-deactivated, item 3. was improved by deactivating the global smoother in #309 and the current PR recovers the functionality of the global gain smoother specifically on crossfades for a low overhead. For item 1. I'm using a lowpass filter as an envelope follower, but @jpcima mentioned that Faust had envelope followers so we might test those to try and reduce the perf hit at some point.

src/sfizz/Smoothers.cpp
```diff
@@ -27,10 +27,25 @@ void Smoother::reset(float value)
...
 {
-    if (smoothing)
+    CHECK_SPAN_SIZES(input, output);
+    if (input.size() == 0)
+        return;
+
+    const auto midValue = input[input.size() / 2];
+    const bool shortcut = (
+        input.front() == input.back()
+        && input.front() == midValue
```
jpcima: how valid is this as a guarantee that signal may be constant?
if we are smoothing a CC, it may well happen as a coincidence that 3 points of the CC stream are equal but the rest of signal isn't.

- beside even if the condition for shortcut is valid, what if we leave a state of filter smoothing which has not completed? do we take a discontinuity?
- same after we leave shortcut state to go back to the filter, we accept the discontinuity there also?

paulfd: > how valid is this as a guarantee that signal may be constant?
> if we are smoothing a CC, it may well happen as a coincidence that 3 points of the CC stream are equal but the rest of signal isn't.

For sure we may miss these cases.

> 
>     * beside even if the condition for shortcut is valid, what if we leave a state of filter smoothing which has not completed? do we take a discontinuity?
> 
>     * same after we leave shortcut state to go back to the filter, we accept the discontinuity there also?

I think for these the condition
```c++
&& input.front() == current()
```
will be false.

paulfd: Maybe another way could be to check if there was an event in the block. If no event AND if the filter has finished smoothing to the expected value, then we can shortcut.

jpcima: Does this mean
- if there is no event ever received,the code is fast
- the moment the first cc arrives, the code is slow?

what is going to happen with floating point IIR? is a filter going to converge moderately "fast" to equal the target value? will it be guaranteed to converge ever?

jpcima: Anyway, I think shortcut can be deduced according to these conditions
- the cc stream is stationary ; defined as having no events in the current cycle
- the filter output is within some reasonable epsilon of the target

paulfd: > Anyway, I think shortcut can be deduced according to these conditions
> 
>     * the cc stream is stationary ; defined as having no events in the current cycle
> 
>     * the filter output is within some reasonable epsilon of the target

This is basically what I meant in the last reply. Let me try to provide an implementation.

paulfd: New tentative in dda4a94

jpcima (approved):

paulfd:
Something weird is happening to the macOS build.

```
==> Upgrading python 3.7.5 -> 3.7.8 

==> Downloading https://homebrew.bintray.com/bottles/python-3.7.8.high_sierra.bottle.tar.gz

==> Downloading from https://akamai.bintray.com/d2/d2315dbe4f3266b7c6efe8bfda3ae0bedc60792cfff00f4bcdf87e5923274ef3?__gda__=exp=1594117301~hmac=5454a7cc1c5f8a97d7c615f34bd1fb566b63bff9eaf0b91d914715babab3137f&response-content-disposition=attachment%3Bfilename%3D%22python-3.7.8.high_sierra.bottle.tar.gz%22&response-content-type=application%2Fgzip&requestInfo=U2FsdGVkX18jMNrNRc2Amfoydar178dCFXVjHn4PCs5tb-5Zi6BHFBG-4_v6BG8-GEJxLJ9S3JRf8jd-bUyN2HpHqMo3ZZOxMML8Uo7DB1JSNVSbZVwZClASKk0qetNczsybzSSDOcx6JZk95NA9rQ&response-X-Checksum-Sha1=22e6d82b35f46437795c84e371b7b20c772a3fe6&response-X-Checksum-Sha2=d2315dbe4f3266b7c6efe8bfda3ae0bedc60792cfff00f4bcdf87e5923274ef3

==> Pouring python-3.7.8.high_sierra.bottle.tar.gz

Error: The `brew link` step did not complete successfully

The formula built, but is not symlinked into /usr/local

Could not symlink Frameworks/Python.framework/Headers

Target /usr/local/Frameworks/Python.framework/Headers

is a symlink belonging to python@2. You can unlink it:

  brew unlink python@2

To force the link and overwrite all conflicting files:

  brew link --overwrite python

To list all files that would be deleted:

  brew link --overwrite --dry-run python

Possible conflicting files are:

/usr/local/Frameworks/Python.framework/Headers -> /usr/local/Cellar/python@2/2.7.17/Frameworks/Python.framework/Headers

/usr/local/Frameworks/Python.framework/Python -> /usr/local/Cellar/python@2/2.7.17/Frameworks/Python.framework/Python

/usr/local/Frameworks/Python.framework/Resources -> /usr/local/Cellar/python@2/2.7.17/Frameworks/Python.framework/Resources

/usr/local/Frameworks/Python.framework/Versions/Current -> /usr/local/Cellar/python@2/2.7.17/Frameworks/Python.framework/Versions/Current

==> /usr/local/Cellar/python/3.7.8/bin/python3 -s setup.py --no-user-cfg install --force --verbose --install-scripts=/usr/local/Cellar/python/3.7.8/bin --install-lib=/usr/local/lib/python3.7/site-packages --single-version-externally-managed --record=installed.txt

==> /usr/local/Cellar/python/3.7.8/bin/python3 -s setup.py --no-user-cfg install --force --verbose --install-scripts=/usr/local/Cellar/python/3.7.8/bin --install-lib=/usr/local/lib/python3.7/site-packages --single-version-externally-managed --record=installed.txt

==> /usr/local/Cellar/python/3.7.8/bin/python3 -s setup.py --no-user-cfg install --force --verbose --install-scripts=/usr/local/Cellar/python/3.7.8/bin --install-lib=/usr/local/lib/python3.7/site-packages --single-version-externally-managed --record=installed.txt

==> Caveats

Python has been installed as

  /usr/local/bin/python3

Unversioned symlinks `python`, `python-config`, `pip` etc. pointing to

`python3`, `python3-config`, `pip3` etc., respectively, have been installed into

  /usr/local/opt/python/libexec/bin

You can install Python packages with

  pip3 install <package>

They will install into the site-package directory

  /usr/local/lib/python3.7/site-packages

See: https://docs.brew.sh/Homebrew-and-Python

==> Summary

🍺  /usr/local/Cellar/python/3.7.8: 4,007 files, 61.1MB

Removing: /usr/local/Cellar/python/3.7.5... (4,049 files, 61.9MB)
```

It's seems that there are no real errors in fact?

jpcima:
Not familiar to me, there is a homebrew issue
https://github.com/Homebrew/brew/issues/1742
and a link to a fix commit at the bottom, we may be able to use the same

PR #315, opened 2020-07-05
Don't require an explicit hikey/key=-1 to trigger on CC

### LINKED ISSUE #239: Trigger on CC incorrect
The following example will trigger on CC in ARIA.
It does not on sfizz, unless `key=-1` is written explicitly.
Surely it's another case of ARIA where `-1` and `<not set>` have identical semantic.

```
<region>
sample=*sine
start_locc16=100
start_hicc16=127
```

paulfd:
Closes #239

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -186,15 +186,15 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
         break;
     // Region logic: key mapping
     case hash("lokey"):
-        triggerOnCC = false;
+        triggerOnNote = true;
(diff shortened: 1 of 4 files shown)


jpcima:
so `lokey` is not concerned by -1 values?

paulfd:
Well, in the PLJ page they said only key and hikey. Apparently it did not catch on 🙂

jpcima (approved):

paulfd:
Updated so that you may actually trigger on both CC and key, matching Sforzando's behavior. @jpcima if you're still OK I'll merge it now.

jpcima:
Sure yes, it makes sense, please do.

PR #319, opened 2020-07-15
Add the spinlock mutex

jpcima:
Not implemented in the synth, just the spinlock mutex and test.

### DIFF
--- a/dpf.mk
+++ b/dpf.mk
@@ -101,6 +101,7 @@ SFIZZ_SOURCES = \
 	src/sfizz/simd/HelpersAVX.cpp \
 	src/sfizz/Synth.cpp \
 	src/sfizz/Tuning.cpp \
+	src/sfizz/utility/SpinMutex.cpp \
 	src/sfizz/Voice.cpp \
 	src/sfizz/Wavetables.cpp
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,8 @@ set (SFIZZ_HEADERS
     sfizz/Config.h
     sfizz/Curve.h
     sfizz/Debug.h
+    sfizz/utility/SpinMutex.h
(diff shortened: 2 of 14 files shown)


paulfd:
Most of our uses of the callback lock are hardcore: so if there is a thread waiting on the mutex for `unlock()` the RT thread will probably not run for a little while so if it misses a deadline then it will miss it alot. I'm good with this however, we could merge it and give it some runtime before the next point release? (by plugging it in the synth already).

I was thinking at some point that a nice-ish way of handling this callback lock would be to signal the RT thread that we have to make deep changes, then let it ramp out the voices for example. Once it's done and cleaned up, the RT thread signals back that the callback can be locked out.

jpcima:
> Most of our uses of the callback lock are hardcore: so if there is a thread waiting on the mutex for unlock() the RT thread will probably not run for a little while so if it misses a deadline then it will miss it alot. 

Yes sure, anyway, I'm never so convinced that locks is as big a deal as some make it to be.
After all, juce uses callback mutexes, and the whole audio industry is based on juce.
In sfizz's case, locking is about loading sfz, and sfz rarely happens at the same time as there is playback, probably not so "hardcore" in practical conditions.

>I was thinking at some point that a nice-ish way of handling this callback lock would be to signal the RT thread that we have to make deep changes, then let it ramp out the voices for example. Once it's done and cleaned up, the RT thread signals back that the callback can be locked out.

There's that, but surely there can be strategies that don't even require locking when you are assuming SPSC.
Admitting you are loading a SFZ file
- control computes loaded SFZ structures into a completely independent structure of its own
- CAS into synth's imaginary `nextSFZ` variable
- RT notices nextSFZ!=null, completes its current synthesis, grab `nextSFZ` by atomic exchange with null
- RT makes `nextSFZ` its current, atomically puts `oldSFZ` into the GC list
- control cleans up GC list at every tick interval of some running timer

PR #320, opened 2020-07-15
Governance model

paulfd:
As we grow, I think it is important that we agree and publicize on how we wish to handle this project. I have prepared a governance model, to support both the technical management of the project as well as the handling of possible donations.

On a more personal note, I tend to put my faith in people's goodwill and intelligence, and I believe this is an aim to strive for in all matters related to people working together. As such, my current goal in this document is to promote a set of rules that foster openness and discussion, so that as we are able to make responsible decisions on possibly difficult topics, such as handling money.

This document and the way we're organized is also meant to evolve to adapt to the needs of the maintainers and the community. The current proposal can be amended, and when accepted it can also change as described within the document.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,19 @@ Other projects that bundle or use sfizz are:
 If you use the library, maybe you will have specific needs and requests that could be useful to you.
 Be sure to post an issue for this so we can improve the API, especially because we have to queue API changes so it stays a bit stable over time.
 
+### Financial contribution
+
+Sfizz and the work in the SFZ tools organization is purely driven by hobbyists up to now, who chose and still choose to use their free time to benefit this project.
+The core sfizz library and the plugins are available under permissive licenses to make them reusable within your own music projects or as components to build applications.
+We firmly believe in the honesty and goodwill of users as a whole, and we want to promote a healthy relationship to software and to the cost of producing quality software.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -217,6 +230,7 @@ The file `scripts/performance_report.py` is made to process and produce a nice i
 [build]: https://sfz.tools/sfizz/development/build/
 [benchmark]: https://github.com/google/benchmark
 [sfizz-render]: https://github.com/sfztools/sfizz-render
+[GOVERNANCE]: https://github.com/sfztools/sfizz/blob/develop/GOVERNANCE.md
```
redtide: I think you can simply use the file name here.

redtide (approved):

paulfd:
After trying it out, it seems that Paypal donations are only available from the "opensource" fiscal host and not (yet) on the Europe one. I do think Paypal make alot of sense, so I am inclined to migrate to that host.

I tried to make a donation to see the process and it's fairly simple, but the fees are actually 5% for OpenCollective + 5% for the host, for a total of 10% + whatever the bank will charge for the money wire. That still seems reasonable for me considering the massive headaches that would result from either tying the donations to an individual or to incorportate a real-life SFZ Tools organization with a Paypal account and the like.

jpcima (approved):

PR #321, opened 2020-07-18
Readable active voices

### LINKED ISSUE #80: Voice cout as LV2 control output
I find having access to this information extremely useful, as it allows me to know how heavy I am pushing the sampler. If the value reaches the voice limit, then I would know I likely have to increase it.

paulfd:
It seems Ardour does not understand the patch:Set for this particular property, I'll investigate.

Closes: sfztools/sfizz-ui#80


jpcima:
The same attempt for VST. However it works weird in several host software in various ways, so maybe not worth to add it.
https://github.com/sfztools/sfizz/commit/1594eaa013cceaaf21aeb1f5e8afa6bac485d5e2

jpcima:
Alternative implementation with control port. It works in Carla, Ardour.

jpcima (approved):

PR #322, opened 2020-07-18
Voice stealing tweaks

paulfd:
Some tweaks on the voice stealing. Comparisons between voices were a bit hazardous in particular, but nothing major.

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -676,8 +676,7 @@ void sfz::Synth::setSampleRate(float sampleRate) noexcept
 void sfz::Synth::renderVoiceToOutputs(Voice& voice, AudioSpan<float>& tempSpan) noexcept
 {
     const Region* region = voice.getRegion();
-    if (region == nullptr)
-        return;
+    ASSERT(region != nullptr);
 
     voice.renderBlock(tempSpan);
     for (size_t i = 0, n = effectBuses.size(); i < n; ++i) {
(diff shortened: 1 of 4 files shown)


src/sfizz/Synth.cpp
```diff
@@ -676,6 +676,9 @@ void sfz::Synth::setSampleRate(float sampleRate) noexcept
 void sfz::Synth::renderVoiceToOutputs(Voice& voice, AudioSpan<float>& tempSpan) noexcept
 {
     const Region* region = voice.getRegion();
+    if (region == nullptr)
```
jpcima: How does it happen that a running voice is not associated with a region?

jpcima:
@paulfd, missed entirely to check this PR, sorry about that.
feel free to be insistent next time I forget.

src/sfizz/Synth.cpp
paulfd: Could be an assert, I had trouble when working on this at some point where voice that were already "killed" were trying to be rendered, it's not necessary anymore.

src/sfizz/Voice.h
```diff
@@ -494,28 +494,31 @@ class Voice {
...
+
+    return true;
 }
 
 inline bool voiceOrdering(const Voice* lhs, const Voice* rhs)
 {
     if (lhs->getAge() > rhs->getAge())
         return true;
-    if (lhs->getAge() < rhs->getAge())
-        return false;
```
jpcima, line 516: If removing the second check, this makes it incorrect as ordering function. Same of other cases below.

One may also write it:
```
if (a.age != b.age) {
  return a.age > b.age;
}
[...]
return false;
```

jpcima (approved):

PR #323, opened 2020-07-18
lv2/sfizz.c: options terminator must have both key and value as 0

atsushieno:
This fixes the problem with such a condition that a host passes a list of 
options where any option with value `0` precedes the block sizes.

From http://lv2plug.in/ns/ext/options#options:

> To implement this feature, hosts MUST pass an LV2_Feature to the appropriate instantiate method with this URI and data pointed to an array of LV2_Options_Option terminated by an element with both key and value set to zero.


paulfd:
Nice, thanks!

PR #324, opened 2020-07-19
Release logic (multiple notes for release=trigger + rt_dead handling)

### LINKED ISSUE #299: Finer `release` behavior
In the current architecture and design, only a single voice may be started for a given event and a given region. However, rgc's sfz and sforzando can start multiple voices on a single event and region. This happens e.g. for
```
<region> sample=*saw lokey=61 hikey=63 trigger=release
```
in this case 3 notes can be started on sustain pedal up.

See also:
- #298 
- https://sfzformat.com/opcodes/trigger.html
- https://sfzformat.com/opcodes/sustain_cc.html

paulfd:
Closes: #299

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -1058,15 +1058,7 @@ bool sfz::Region::registerNoteOff(int noteNumber, float velocity, float randValu
         // If we reach this part, we're storing the notes to delay their release on CC up
         // This is handled by the Synth object
 
-        const auto sameNoteTest = [noteNumber](const std::pair<int, float>& noteAndValue) {
-            return noteAndValue.first == noteNumber;
-        };
-
-        auto it = absl::c_find_if(delayedReleases, sameNoteTest);
-        if (it == delayedReleases.end())
-            delayedReleases.emplace_back(noteNumber, midiState.getNoteVelocity(noteNumber));
-        else
-            it->second = velocity;
+        delayedReleases.emplace_back(noteNumber, midiState.getNoteVelocity(noteNumber));
     }
 
     return false;
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -111,7 +111,7 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             loopMode = SfzLoopMode::loop_sustain;
             break;
         default:
-            DBG("Unkown loop mode:" << std::string(opcode.value));
+            DBG("Unkown loop mode:" << opcode.value);
         }
         break;
     case hash("loop_end"): // also loopend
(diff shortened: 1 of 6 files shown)


paulfd:
At this point maybe it would be more "logical" to let the Region object be purely a data storage, and handle all the changes of the "status" of each regions within the Synth (i.e. instead of having the `registerNoteOn` and co. change the internal private status of the regions, all status variable are public and handled by the synth). What do you think?

src/sfizz/Region.cpp
```diff
@@ -1026,24 +1026,45 @@ bool sfz::Region::registerNoteOff(int noteNumber, float velocity, float randValu
...
+        // If we reach this part, we're storing the notes to delay their release on CC up
+        // This is handled by the Synth object
+
+        const auto sameNoteTest = [noteNumber](const std::pair<int, float>& noteAndValue) {
+            return noteAndValue.first == noteNumber;
+        };
+
+        auto it = absl::c_find_if(delayedReleases, sameNoteTest);
+        if (it == delayedReleases.end())
+            delayedReleases.emplace_back(noteNumber, midiState.getNoteVelocity(noteNumber));
```
jpcima: do we force voice count to be <= 1 per note number ?

jpcima:
>change the internal private status of the regions, all status variable are public and handled by the synth). What do you think?

Without a doubt yes. This helps to manage the complexity a bit as the software grows.
As source organization is concerned, there is an increasing number of sources at the root level which are purely utility, I would like these to be placed in a subdirectory preferably.

src/sfizz/Region.cpp
paulfd: You mean you want possibly multiple release samples per note? Maybe it could make sense for some instruments... Probably we should match what happens in sforzando.

jpcima: Can a single midi note have more that 1 Voice on it?
because that can be a desirable property.

paulfd: Yes sure, if multiple regions fire on the note trigger no problem. But this change is only for having multiple voices fire on a single region, which I think only happens conceptually if there is sustain and the region is a "release" region. Or maybe I misunderstand your question :slightly_smiling_face:

paulfd:
> > change the internal private status of the regions, all status variable are public and handled by the synth). What do you think?
> 
> Without a doubt yes. This helps to manage the complexity a bit as the software grows.
> As source organization is concerned, there is an increasing number of sources at the root level which are purely utility, I would like these to be placed in a subdirectory preferably.

Sure and sure. I'll defer the passive region in another pull request, and feel free to reorganize the sources :slightly_smiling_face:

paulfd:
I added tests and support for the `rt_dead` opcode. The behavior now matches sforzando/dimension, and the `trigger=release` regions will only sound if there is a corresponding attack sample, corresponding meaning compatible in the key and velocity range.

The architecture is not entirely satisfactory but I will defer overhauling the activations/dispatching in another PR. The tests added in this one will help ensure nothing breaks, and I will add a more comprehensive set in the upcoming PR as this is relatively easy to test programmatically.

jpcima (approved):

PR #327, opened 2020-07-19
Support for `sustain_lo`

paulfd:


jpcima:
Verified in Sforzando here, seems good

PR #328, opened 2020-07-23
Failing parser test on a define

paulfd:
@jpcima This test fails, any idea why?

If not I'll have time to investigate later :slightly_smiling_face:


paulfd:
Note: the second row is the failing line. The parser parses `-12 amplitude_oncc100=100` as the value for the `transpose` opcode.

jpcima (approved):

Issue #330, opened 2020-07-24
Add oscillator subtables


jpcima:
There are some audio formats which permit extraction of multiple wavetables from a single file.
It is a starting point to implement multi-wavetable in a way which doesn't yet require SFZ extensions.

- WT files
  see https://github.com/surge-synthesizer/surge/blob/master/resources/data/wavetables/wt%20fileformat.txt
- WAV files with extension blocks
  see https://github.com/surge-synthesizer/surge/blob/master/src/common/WavSupport.cpp

In the basic implementation, it's possible to implement the playback as a round robin in the sequence of wavetables.

baconpaul:
Hey @jpcima 

so the only authors on WavSupport.cpp are me (for 98% of it) and @mkruselj for 2 lines. I spoke with @mkruselj and he and I are fine for you to take a copy of that cpp file and use it in a non-GPL3 setting. Or if you want to make it a submodule with an MIT license we can work on that too. It's more an art than a science ... I just got people to send me various wavetables which worked in various synths and plucked apart what I thought they meant. 

Pretty sure you would want to eject different data structures anyway so a copy probably makes sense, but zero problem from either of us with an (attributed) copy in a BSD, MIT, Apache, or GPL<3 licensed open source product. 

What I really wanted to do was to push non-format block reading code up into libsndfile so I didn't have to parse the wav myself. But libsndfile won't give you non-standard CHUNKs in any way - it just (correctly) ignores them. So had to write that code :(

The WT are much simpler but only surge uses them. Some python code exists https://github.com/surge-synthesizer/surge/tree/main/scripts/wt-tool also to do some relabeling and stuff.

jpcima:
This is great, thank you to both.

- Regarding RIFF blocks:

These are familiar, we already deal with 'smpl' chunks to some extent.
We made this discovery that flag, ogg, and opus files can contain these very same 'smpl' blocks that riff has, and we have already some code to deal with the libsndfile insufficiency in this regard (for flac).

Do you know if these RIFF blocks apply to wav files only, or these other products will also use them from flac and others?

- Regarding WT

I have made an implementation of this one some ago, submitted as PR into Andrew Belt's WaveEdit, which does not seem to get any interest there. (but might be of interest for some users)

baconpaul:
[...]

PR #333, opened 2020-07-25
File metadata

jpcima:
#330

This adds a metadata reader, which is capable of identifying sound files which contain wavetables.
This is determined by vendor-specific blocks which are present in the file metadata.

To implement this, a refactor was desired so the RIFF blocks are more easily accessible, and particularly in WAV files.

Wavetables Support
- Surge
- Serum
- u-he
- ~~some Native Instruments~~

EDIT removed Native, it was incorrect. these wavetables have their definition provided as sidecar text file.


jpcima:
Added some tests to ensure wavetable is properly extracted.
- Surge: test file is made with `./scripts/wt-tool/add-surge-metadata.py`
- Clm ("Serum"): made with a custom code at `https://github.com/jpcima/WaveTableFactory/blob/627c20a7b35a4ab21dbb063eb0c5be35f0154acb/sources/wavetable.cpp#L71`
- U-he: too complex format, I don't know how to make (but we might create a dummy since it's just the chunk presence which is checked, not content)

src/sfizz/FilePool.cpp
```diff
@@ -218,11 +218,12 @@ absl::optional<sfz::FileInformation> sfz::FilePool::getFileInformation(const Fil
...
     SF_INSTRUMENT instrumentInfo {};
 
-    const int sndFormat = reader->format();
-    if ((sndFormat & SF_FORMAT_TYPEMASK) == SF_FORMAT_FLAC)
-        sfz::FileInstruments::extractFromFlac(file, instrumentInfo);
-    else
-        reader->getInstrument(&instrumentInfo);
+    if (!reader->getInstrument(&instrumentInfo)) {
+        // if no instrument, then try extracting from embedded RIFF chunks (flac)
+        FileMetadataReader reader;
```
paulfd: This shadows the other reader

src/sfizz/FileMetadata.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+const RiffChunkInfo* FileMetadataReader::riffChunk(size_t index) const
+{
+    return impl_->riffChunk(index);
+}
+
+const RiffChunkInfo* FileMetadataReader::Impl::riffChunk(size_t index) const
+{
+    const std::vector<RiffChunkInfo>& riffChunks = riffChunks_;
+    return (index < riffChunks.size()) ? &riffChunks[index] : nullptr;
```
paulfd: I didn't try, but a static cast to `const RiffChunkInfo*` is not working?

paulfd (approved):

paulfd:
I trust you for the file formats ;-)

src/sfizz/FileMetadata.cpp
jpcima: Sorry explain what is the problem please, I don't understand it.

paulfd: Not really a problem per-say. I'm thinking you're creating the const ref to get a const pointer from the original vector, and I'm wondering if for some reason it's not possible to just cast to `const RiffChunkInfo*`. But it's not blocking, just curiosity. I will just try on my copy :D

paulfd:
```c++
const RiffChunkInfo* FileMetadataReader::Impl::riffChunk(size_t index) const
{
    return (index < riffChunks_.size()) ? &riffChunks_[index] : nullptr;
}
```
This seems to compile for me.

jpcima: Oh you prefer the one-line, sure I'll change it.

PR #335, opened 2020-07-27
Add modulation matrix

jpcima:
Modulation matrix update

This model is operational and it can handle CC.
Support
- Amplitude
- Pan
- Width
- Position
- Pitch
- Volume

Not much tested yet, also I will clean old source which is made obsolete by this.
This will also make obsolete #326.
The CC list will be obtainable by inspecting matrix edges of CC source. Also it permits to have access to the target, and determine a useful name for the midnam item.


jpcima:
There is some `EventEnvelopesT` code `#if 0`'d in the removal commit above. I'm not sure yet what to replace it with.
I removed all old things regarding modifiers.

jpcima:
The commit 51b7322 ensures to always run the modulation source, even if no voice uses it.
This ensures the CC will continue to smooth in the back, even with no voice playing.

jpcima:
@paulfd this faces a problem with the smooth filter shortcut..

For instance I'm smoothing pitch CC at center 600, value 1200.
The filter basically stops rising at around `599,121033, and it's too far off of the `virtuallyZero` difference threshold such that shortcutting would be possible.

What to do in this case? consider relative difference?

paulfd:
> There is some `EventEnvelopesT` code `#if 0`'d in the removal commit above. I'm not sure yet what to replace it with.
> I removed all old things regarding modifiers.

You can remove it I think, these tests were not great and we can make better ones for the FlexEGs

paulfd:
> @paulfd this faces a problem with the smooth filter shortcut..
> 
> For instance I'm smoothing pitch CC at center 600, value 1200.
> The filter basically stops rising at around `599,121033, and it's too far off of the `virtuallyZero` difference threshold such that shortcutting would be possible.
> 
> What to do in this case? consider relative difference?

Hmm I'd say yes, but is it OK on both ends ? (e.g. for small values too)

jpcima:
>but is it OK on both ends ? (e.g. for small values too)

hmm, as a relative distance, how about:
`(abs(x-y)) / (abs(y)+virtuallyZero)`

paulfd:
Yes that seems reasonable.

jpcima:
The only remaining feature that I can see missing, it's a modulation affecting the depth of another modulation.
eg. `lfoN_amplitude_onccX`
This will be a feature for another PR.

src/sfizz/modulations/ModGenerator.h
```diff
@@ -0,0 +1,66 @@
...
+ * @brief Generator for modulation sources
+ */
+class ModGenerator {
+public:
+    virtual ~ModGenerator() {}
+
+    /**
+     * @brief Set the sample rate
+     */
+    virtual void setSampleRate(double sampleRate) = 0;
```
paulfd: Is the `=0` needed here?

src/sfizz/modulations/ModGenerator.h
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @brief Set the maximum block size
+     */
+    virtual void setSamplesPerBlock(unsigned count) = 0;
```
paulfd: Is the `=0` needed here?

tests/RegionT.cpp
```diff
@@ -541,28 +545,29 @@ TEST_CASE("[Region] Parsing opcodes")
 
     SECTION("pan_oncc")
     {
-        REQUIRE(region.modifiers[Mod::pan].empty());
+        const ModKey target = ModKey::createNXYZ(ModId::Pan, region.getId());
```
paulfd, line 548: Why isn't this a CC source? (`createCC`)

src/sfizz/modulations/sources/Controller.cpp
```diff
@@ -0,0 +1,94 @@
...
+struct ControllerSource::Impl {
+    double sampleRate_ = config::defaultSampleRate;
+    Resources* res_ = nullptr;
+    absl::flat_hash_map<ModKey, Smoother> smoother_;
+};
+
+ControllerSource::ControllerSource(Resources& res)
+    : impl_(new Impl)
+{
+    impl_->res_ = &res;
```
paulfd, line 27: We could assert that the pointer is non-null, since this should be a reference but can't because of the oo structure.

src/sfizz/modulations/ModMatrix.cpp
```diff
@@ -0,0 +1,427 @@
...
+#include <algorithm>
+
+namespace sfz {
+
+struct ModMatrix::Impl {
+    double sampleRate_ {};
+    uint32_t samplesPerBlock_ {};
+
+    uint32_t numFrames_ {};
+    NumericId<Voice> voiceId_ {};
```
paulfd: could be named `currentVoiceId_`

src/sfizz/modulations/ModMatrix.cpp
```diff
@@ -0,0 +1,427 @@
...
+    NumericId<Region> regionId_ {};
```
paulfd: could be named `currentRegionId_`

src/sfizz/modulations/ModMatrix.cpp
```diff
@@ -0,0 +1,427 @@
...
+        const float sourceDepth = sourcesPos->second.sourceDepth_;
+        const int sourceFlags = source.key.flags();
+
+        // only accept per-voice sources of the same region
+        bool useThisSource = true;
+        if (sourceFlags & kModIsPerVoice)
+            useThisSource = (regionId == source.key.region());
+
+        if (useThisSource) {
+            if (isFirstSource) {
```
paulfd, line 335: Could shortcut here if `depth == 0.0f`

paulfd (approved):

src/sfizz/modulations/ModMatrix.cpp
jpcima: agreed it's better naming, modified in latest

src/sfizz/modulations/ModGenerator.h
jpcima: No, now it's gone.
In reality it's only CC that need these methods, but with some refactor they might go altogether.

tests/RegionT.cpp
jpcima: Not sure of what you mean here, I'll just summarize what goes on.

- Region is going to parseOpcode of everything looking like `pan_onccN` as well as curve/step/etc
  These virtual connections for the region are cached into a list.
  The connection is made later at Synth level, because CC sources are not definitive in an incremental process.
  (example you can receive a smoothcc/curve/step/value for this connection, this will modify its source key)

- As result, you get a bunch of connections in a list
  ```
  cc45 -> pan
  cc17 -> pan
  [...]
  ```
  It's not possible ending with several connections of the same CC number and destination
  ```
  cc45 smooth=10 -> pan
  cc45 smooth=50 -> pan  /// not possible
  ```

- So when you specify a region, target id, cc number, you designate an unique item in there.
  `RegionCCView` gives an interface to access this, while not much modifying too much the structure of the old unit test. (think of it like CCMap in O(n))

src/sfizz/modulations/sources/Controller.cpp
jpcima: `res_` can be a reference if there exists the `Impl(Resources &)` ctor.
I don't see too much why this is a big deal though.

src/sfizz/modulations/ModMatrix.cpp
jpcima: Yes not only you're correct, but there is a greater problem.
The source is supposed to be buffered per region or cycle, not have its generation output used directly.
This makes the source generate more than once if in use by multiple targets.

ie. make use of Source `buffer`/`bufferReady`

jpcima: See a97f506.
Regarding the shortcut: a future consideration is that depth can be modulated.
When this is implemented, `sourceDepth`=0 is an insufficient condition, one has also to check that depth is free of any modulations.

I'm sure I to have this idea in mind previously, but postponed it until making the depth modulation.
There is not an immediate benefit anyway, the source will have to implement a fast path in `generateDiscarded`.

As CC as concerned, I believe the fast path is implementable in case of not smoothing, or smoothing near completed.

src/sfizz/modulations/sources/Controller.cpp
paulfd: I must have misplaced the comment I was thinking of something else and can't figure out what now. Nvm :)

paulfd (approved):

PR #338, opened 2020-07-28
LFO

jpcima:
This adds SFZv2 LFO on top of the mod matrix PR.

There are the basic targets for now (amplitude, pitch, etc)

Simple usage example
```
<region>
sample=*sine
lfo1_freq=1.0
//lfo1_amplitude=100
lfo1_pitch=100
```


src/sfizz/LFO.cpp
```diff
@@ -0,0 +1,295 @@
...
+        phase += incrPhase;
+        int numWraps = (int)phase;
+        phase -= numWraps;
+    }
+
+    impl.subPhases_[nth] = phase;
+}
+
+template <LFOWave W>
+void LFO::processSH(unsigned nth, absl::Span<float> out)
```
paulfd, line 149: These functions share alot of code, I wonder if we could avoid this repetition. I can try to work on this unless you have an idea.

src/sfizz/LFO.cpp
```diff
@@ -0,0 +1,295 @@
...
+            processSH<LFOWave::RandomSH>(subno, out);
+            break;
+        }
+    }
+
+    float fadeIn = impl.fadeInMemory_;
+    const float fadeInPole = impl.fadeInPole_;
+    for (size_t i = 0; i < numFrames; ++i) {
+        out[i] *= fadeIn;
+        fadeIn = fadeInPole * fadeIn + (1 - fadeInPole);
```
paulfd: This is the fade in used in ARIA/Cakewalk?

src/sfizz/LFO.cpp
```diff
@@ -0,0 +1,295 @@
...
+    float baseFreq = desc.freq;
+    float offset = sub.offset;
+    float ratio = sub.ratio;
+    float scale = sub.scale;
+    float sampleHoldValue = impl.sampleHoldMem_[nth];
+
+    for (size_t i = 0; i < numFrames; ++i) {
+        out[i] += offset + scale * sampleHoldValue;
+
+        // TODO(jpc) lfoN_count: number of repetitions
```
paulfd, line 167: Is there more than counting the cumulative number of numWraps?

src/sfizz/Region.cpp
```diff
@@ -1324,3 +1546,22 @@ float sfz::Region::getBendInCents(float bend) const noexcept
...
+    auto it = std::find_if(connections.begin(), connections.end(), pred);
+    if (it != connections.end())
+        return *it;
+
+    sfz::Region::Connection c;
+    c.source = source;
+    c.target = target;
+
+    connections.push_back(c);
+    return connections.back();
```
paulfd, line 1566: Why don't we have this as a return value of push_back/emplace_back in C++11, I'll never understand...

src/sfizz/Voice.cpp
```diff
@@ -34,6 +35,10 @@ sfz::Voice::Voice(int voiceNumber, sfz::Resources& resources)
         filter.setGain(vaGain(config::filteredEnvelopeCutoff, sampleRate));
 }
 
+sfz::Voice::~Voice()
```
paulfd, line 38: Any reason to have this?

src/sfizz/Region.cpp
```diff
@@ -786,6 +786,228 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
...
+        break;
+    case hash("lfo&_phase"):
+        {
+            const auto lfoNumber = opcode.parameters.front();
+            if (lfoNumber == 0)
+                return false;
+            if (!extendIfNecessary(lfos, lfoNumber, Default::numLFOs))
+                return false;
+            if (auto value = readOpcode(opcode.value, Default::lfoPhaseRange)) {
+                float normalPhase = *value * (1.0 / 360.0);
```
paulfd, line 808: Are you sure about this? The sfzformat page seems to contradict this.

paulfd:
I added some tests, trying to automate them also on travis. It may fail, bear with me :)

src/sfizz/LFO.cpp
jpcima: It's what it's supposed to be. In Aria, this is unimplemented, and I never tried it in Cakewalk.
The reason of omitting it for time being, it's vectorizability of this whole loop is challenged by this conditional branch.
(but surely there are some workable alternative solutions)

src/sfizz/Region.cpp
jpcima: same

src/sfizz/Voice.cpp
jpcima: This was a recommendation of google c++ guidelines.
This would ensure destructions don't get inlined in the caller, for economy of code size.

src/sfizz/LFO.cpp
jpcima: Fade is too slow, I'll adjust the ratio to make this better.

Sfizz-Cakewalk and Sfizz-Aria comparison
sfizz-cakewalk
sfizz-aria

jpcima: As it turns out, the fade-in is just a linear progression. Implemented in latest.
(as a note, it would be trivial to let fade-in be defined by curve, if that might be desired)

src/sfizz/LFO.cpp
jpcima: Repetitions, in particular `processWave`, it's not just for avoid the branch in the frame loop, also I wish for SSE impls of wave functions. All are vectorizable.

src/sfizz/Region.cpp
jpcima: Phase is definitely expressed in degree unit.

phase=0
LFO phase=0

phase=90
LFO phase=90

jpcima:
On reference data points, may we possibly flac them up to reduce their size?

paulfd (approved):
So if you're certain about the phase I guess we need to amend sfzformat.com, and if my solution for the reference sampling point is OK, I'm good with the rest :slightly_smiling_face:

jpcima:
Last commit introduces a terrible hack for Garritan instrument compatibility.
- if `lfoN_phase` is 0<=x<=1, it's considered with the ARIA meaning which is 0-1 normalized
- otherwise, it's taken as degrees, as standard SFZ suggests
- if the user adds trailing text to the value, this forces a particular choice of unit, eg. `0.5d` → 0.5 degrees

paulfd:
This seems like a reasonable solution to a messy problem!

AlexMitchellMus:
`lfoN_phase_onccX` would then follow the SFZv2 standard, or use both as well?

jpcima:
I have done some changes as follows to help make the work more mergeable
- control all of `plot_lfo` settings by arguments
- squashed the test commit and moved on top
- postponed the ARIA compatibility work, kept in a side branch; there is discussion to have on this topic, as every proposed solution has its advantages and inconvenients, and I'm not interested to take the decision here, just advance LFO for coming work

On the tests, I wish to have them preferably included in the regular test suite, I will add some work towards this direction.

jpcima:
Regression tests are moved into the main test suite. Good for merge.
Do we keep these numpy files?

paulfd:
I like the plotting one but it can go somewhere else. I'm not used to whipping out gnuplots on the fly :)

We still need something like it to validate changes visually and possibly update the reference.

PR #340, opened 2020-08-01
Some speedup with SSE 4.1

jpcima:
This speeds up `sample_quality=2` by 15 to 20%, using SSE4.1 dot-product primitive, and avoiding a bit of instruction latency.
Just for illustrating, this optimization should be made CPU-dispatched.
Possibly `strings` can benefit from a similar optimization.

### DIFF
--- a/cmake/SfizzConfig.cmake
+++ b/cmake/SfizzConfig.cmake
@@ -35,7 +35,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
     add_compile_options(-ffast-math)
     add_compile_options(-fno-omit-frame-pointer) # For debugging purposes
     if (SFIZZ_SYSTEM_PROCESSOR MATCHES "^(i.86|x86_64)$")
-        add_compile_options(-msse2)
+        add_compile_options(-msse4.1)
     endif()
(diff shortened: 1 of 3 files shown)


paulfd:
Nice one! With the runtime dispatch I think we can target interesting optimization like this. Do you want that I benchmark it on Intel/AMD? I was thinking on working on ARM in the holidays, if I have some time :slightly_smiling_face:

paulfd:
btw the `meanSquared` SIMD helper is also a dot product.

paulfd:
Considering the simde version you proposed, is this speedup obsolete? We could maybe have a runtime dispatcher.

jpcima:
> Considering the simde version you proposed, is this speedup obsolete? We could maybe have a runtime dispatcher.

It's by no means obsolete but it would be desirable to have the cpu dispatcher.

From experimenting with the strings effect, I discovered that one can extract great speed benefits from loop unrolling, and more so when coupled with some inlining. (some greater than 4x on SSE, which might be explained by latency effects of memory or individual instructions)
I'd like the same to be experimented with the resampler; but the simde PR should be dealt with first.

paulfd:
Sure, I think the simde PR is fine now.

Dec 26, 2020 15:25:18 JP Cimalando <notifications@github.com>:

PR #342, opened 2020-08-05
Add compressor and gate

jpcima:
The compressor effect, oversampled with factor 2x.

- if `stlink=off`, independent gains are computed for each channel
- if `stlink=on`, a single gain is computed for the L+R sum

Example
```
<region>
sample=*sine

<effect>
type=comp
comp_attack=.05
comp_release=1
comp_threshold=-10
comp_ratio=1.5
comp_gain=3.0
comp_stlink=on
```


jpcima:
Also add gate, a similar effect.

```
<effect>
type=gate
gate_attack=0.1
gate_release=0.1
gate_stlink=off
gate_threshold=-40.0
gate_oncc21=100
```

PR #345, opened 2020-08-08
Fix build: update debug messages and assertions to match the latest API.

atsushieno:
These build failures don't seem to have been caught by CI builds because they were debug-only failures.


jpcima:
Hi, thank you.
Yes it's this problem again.

@paulfd what do you think of expanding release `DBG` into a form that checks the code but not run it?
```
if (0)
  std::cerr << ...;
```

PR #346, opened 2020-08-08
Check the validity of DBG() under release mode

jpcima:
Ensures that DBG will always compile.

### DIFF
--- a/src/sfizz/Debug.h
+++ b/src/sfizz/Debug.h
@@ -64,10 +64,10 @@
 #endif
 
 // Debug message
-#if !defined(NDEBUG) || defined(SFIZZ_ENABLE_RELEASE_DBG)
 #include <iostream>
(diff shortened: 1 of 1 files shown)


paulfd:
We could also add a CI run in debug mode (which would also be good practice anyway). What about using this to make it a bit more symmetrical and remove the `do .. while` too?

```c++
#include <iostream>
#if !defined(NDEBUG) || defined(SFIZZ_ENABLE_RELEASE_DBG)
#include <iomanip>
#define DBG(ostream) if (true) { std::cerr << std::fixed << std::setprecision(2) << ostream << '\n'; }
#else
#define DBG(ostream) if (false) { std::cerr << ostream << '\n'; }
#endif
```

jpcima:
`Do..while` block should stay, the purpose is to protect against some problems when this expands within if-else chains.

paulfd:
ok then it's good like this

PR #347, opened 2020-08-09
Higher quality wavetable mipmaps

jpcima:
This increases the number of mipmaps and modifies the mapping formula frequency→table.
The older mipmaps could have a gap in higher frequencies which reaches as low as 11 kHz over some ranges.

This implements a case similar to the 24-table as seen here, which performs significantly better
https://www.earlevel.com/main/2019/04/30/waveutils-updated/

The mapping uses a base ≠ 2 for logarithm which is not so trivially computable in this case, the computation has been replaced with a look-up table. Formula is documented in comments.

A suggestion from these articles is to raise the `refSampleRate` by some amount, eg. +10%, such that some slight aliasing may be allowed, and this permits to reduce further the table transition harmonic gap.
If this is used, this can permit to maintain a comparable quality level but without increasing the number of tables.
To measure. Not done as a part of this commit but it's trivial changing it.

The formula is designed such that it adapts to some parameter constants.
Following can be edited for experimenting
- number of tables: `static constexpr unsigned N = 24;`
- reference sample rate for cutoff `double refSampleRate = 44100.0`
  (eg. change to `1.1 * refSampleRate` for 10% aliasing permissivity)

wavetables

Gnuplot code
```
set xrange [10:22000]
set logscale x
set grid
set key top left

log2(x)=log(x)/log(2)
min(a, b)=(a<b)?a:b
max(a, b)=(a>b)?a:b

# defaults
f0=20.0       # minimum frequency
fK=12000.0    # start frequency of last table

# compute adjustment variables b and k, such that first/last table begins at frequency f0/fK
k(f0, fK, nt)=1.0/f0
b(f0, fK, nt)=exp(log(fK/f0)/(nt-1.0))

# compute table index for frequency
t(f, f0, fK, nt)=max(0, min(nt-1, (f<f0)?0:floor(log(k(f0, fK, nt)*f)/log(b(f0, fK, nt)))))

plot \
     t(x, f0, fK, 16) t "new 16 tables", \
     t(x, f0, fK, 20) t "new 20 tables", \
     t(x, f0, fK, 24) t "new 24 tables", \
     floor(max(0, log2(0.05*x))) t "old"
```

### DIFF
--- a/src/sfizz/Config.h
+++ b/src/sfizz/Config.h
@@ -95,6 +95,7 @@ namespace config {
     constexpr int maxEffectBuses { 256 };
     // Wavetable constants; amplitude values are matched to reference
     static constexpr unsigned tableSize = 1024;
+    static constexpr double tableRefSampleRate = 44100.0 * 1.1; // +10% aliasing permissivity
     static constexpr double amplitudeSine = 0.625;
     static constexpr double amplitudeTriangle = 0.625;
     static constexpr double amplitudeSaw = 0.515;
--- a/src/sfizz/Wavetables.cpp
+++ b/src/sfizz/Wavetables.cpp
@@ -283,65 +283,70 @@ const HarmonicProfile& HarmonicProfile::getSquare()
 }
 
 //------------------------------------------------------------------------------
-constexpr unsigned WavetableRange::countOctaves;
-constexpr float WavetableRange::frequencyScaleFactor;
+constexpr unsigned MipmapRange::N;
+constexpr float MipmapRange::F1;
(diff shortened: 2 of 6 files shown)


jpcima:
On a side note, vocabulary has been changed a bit.
`octave` is gone, renamed to `index`, because table starts are no longer spaced by 2x.

The term mipmap is introduced. In the future, the current term `multi` is to become `mipmap`, and I prefer `multi` to become the term for multi-dimensional wavetables.

paulfd (approved):

PR #348, opened 2020-08-09
Ignore released voices in search of selfmask candidate

jpcima:
Is it correct to think that selfmask search should ignore voices which are already released?
this should prevent it to reselecting a voice which has been previously self-masked but not finished its envelope? @paulfd


paulfd:
Yes, you're right. The check could also be first in the `if (..)` statement to shortcut the other 2 for freed voices.

PR #353, opened 2020-08-10
Add runtime configs for loading in ram and voice stealing

paulfd:
Still needs some testing; allows for `hint_` opcodes to control ram loading and voice stealing.


jpcima:
Are hint opcodes handled at global level, or can it be also per region (sample) ?

paulfd:
For sforzando apparently they're in a `<control>` block, but I did not test anything.

I guess they could be anywhere. You're thinking of possibly tagging individual samples to load in ram?

For the voice stealing it's a bit touchy, since the algorithm choice is rather from the killed voice than the killing one...

paulfd:
Now disables/enables the power follower depending on the stealing algorithm.

jpcima (approved):

PR #359, opened 2020-08-12
Add VST info panel

jpcima:
Adds an info panel which displays VST active polyphony and other things. @kinwie

Capture du 2020-08-12 19-24-24


kinwie:
This is great!
I just think better to have Globals than Curves though

paulfd:
Well since I thought that there was only 1 global per file, there are no counters in sfizz for the number of globals :P

paulfd (approved):

kinwie:
Ohh okay, so we are on the way to having more headers then :))

jpcima:
`<global>` and `<control>` are missing.
It would be simply the number of appearances of these headers.
Also missing is the name of the last sample which has been triggered.

PR #371, opened 2020-08-17
Fix amp_veltrack

### LINKED ISSUE #370: Amp_veltrack curve update
Current amp_veltrack curve in sfizz not yet fully matched rgc sfz and sforzando. These two players has identical amp_veltrack curve.

Default, amp_veltrack=100 is matched. But setting it to other value, the curve has different value than those two

jpcima:
Fixes #370
@kinwie


jpcima:
Velocity plot attached
Capture du 2020-08-17 15-35-44

kinwie:
Checked in details. It is now match and identical to rgc sfz, DropZone and sforzando. Super job @jpcima !!

paulfd (approved):

PR #376, opened 2020-08-18
Adjust the ADSR exponential rate

jpcima:
This makes exponential curves a tiny bit nearer to what Cakewalk has.
@kinwie


kinwie:
Yes checked, nicer now. I agree to closer to Cakewalk has

jpcima:
Decreased the release end threshold to -80 dB.
It's to better match the other software, which accept to go to lower levels.

paulfd (approved):

Issue #372, opened 2020-08-17
Release Trigger adjustment


kinwie:
Current sfizz release trigger have some disadvantages.
These are what I think may need to be adjusted :

1. Sfz specs default behavior: By setting trigger to `release` or `release_key` will automatically set the region into one-shot mode.
_*Current state in sfizz : It is not and we need to put ampeg_hold= or ampeg_release= with high value to make the release region play, which is no needed if it automatically put into loop_mode=one_shot._

2. Playing notes with sustain pedal down (especially repeated notes) then release the pedal, need `note_polyphony=1` to limit each released notes to 1 voice per key.
_*Current state in sfizz :_
- note_polyphony not yet recognized by the release regions, so for example, with sustain pedal down, playing fast 10 notes of the same key, then release the pedal, all 10 released notes will play. It should be limited follow the note_polyphony value.
- off_time not yet working properly. Sfizz default to off_mode=normal and depends on the ampeg_release value for the attack region's fade-out time, so we can't set different fade-out time independently. This can be an issue for the high undampened piano keys which have long release time, but to have a shorter off_time (In Aria, this is set to off_mode=time by default).

3. Adding `note_polyphony=1` will also cause both the attack and release regions of the same key, muting each other (bi-directional). This is the default behavior with no `group=` set (both attack and release regions are in the same `group=`)
_*Current state in sfizz : release region can't get muted by the attack region because the use of long ampeg_release value for it to play_

Point #3 could be just gone if #2 fixed

kinwie:
This simply will just works in both DropZone and sforzando :
```
<group> note_polyphony=1
<region> sample=sine.wav
<region> sample=saw.wav trigger=release
```
Both will mute each other (per key) in both pedal up and pedal down

paulfd:
I have yet to consolidate the polyphony logic to note-off and cc-triggered values, so points 2 and 3 are expected in a way ;-)

For the first point it's a valid remark. We do take into account the amplitude envelop for release voices, so you can actually do
```
<region> trigger=release_key sample=*sine ampeg_attack=0.02 ampeg_release=0.1
```
In sforzando afair this generates an inifinite sine wave.

kinwie:
Yes, as the `note_polyphony` was just fixed last week, I'm also guessing that not yet put into release trigger.

Yes, `ampeg_attack` `ampeg_hold` `ampeg_decay` `ampeg_sustain` should works  for release trigger. Only `ampeg_release` is ignored.

[...]

PR #379, opened 2020-08-18
Set loop_mode to one_shot for release regions

paulfd:
See #372

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -349,9 +349,11 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             break;
         case hash("release"):
             trigger = SfzTrigger::release;
+            loopMode = SfzLoopMode::one_shot;
             break;
         case hash("release_key"):
             trigger = SfzTrigger::release_key;
+            loopMode = SfzLoopMode::one_shot;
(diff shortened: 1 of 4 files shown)


src/sfizz/Region.cpp
```diff
@@ -349,9 +349,11 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             break;
         case hash("release"):
             trigger = SfzTrigger::release;
+            loopMode = SfzLoopMode::one_shot;
             break;
         case hash("release_key"):
             trigger = SfzTrigger::release_key;
+            loopMode = SfzLoopMode::one_shot;
             break;
```
jpcima: This behavior will not match to aria.
This is overriden by the `loop_mode` if present, regardless if happening before or after `trigger`.
For instance both of these will have `loop_sustain`.
```
<region> sample=sine440.wav pitch_keycenter=69 loop_mode=loop_sustain trigger=release_key
<region> sample=sine440.wav pitch_keycenter=69 trigger=release_key loop_mode=loop_sustain
```

paulfd: an loopmode is already an optional too! :)

jpcima (approved):

Issue #373, opened 2020-08-18
Noise print or denormal occur when ampeg_release not reached


kinwie:
When the ampeg_release stage not yet reached and the region ends, sfizz produce some kind of noise print of denormal after it.

paulfd:
There should be no denormals, so it might be noise or garbage. Do you have a test case?

kinwie:
Yes, tempo 120 bpm
```
<region>
sample=*sine
ampeg_attack=1
ampeg_hold=1
ampeg_decay=1
ampeg_sustain=0
ampeg_release=1
```

Draw one bar midi note, so it reach the release stage, no noise :
![Untitled](https://user-images.githubusercontent.com/65444898/90631022-3c94d480-e24c-11ea-85b0-1145c1d0ad1c.jpg)

Draw two bars note, it's not reach the release stage, there is noise after it (it should ends in the 2nd bar, 3rd beat, like first image) :
![Untitled2](https://user-images.githubusercontent.com/65444898/90631306-b5942c00-e24c-11ea-8b38-598e6836843b.jpg)

This seems only occur when the ampeg_sustain value is 0

paulfd:
Was this one improved this summer?

jpcima:
It's because we cut the envelope release at -60 dB.
You can lessen it by aiming at -80 dB.
If we rewrite the ADSR entirely in curves, there is no more need to take care of this asymptotic behavior.
(but then unnatural behavior on level/time dynamic updates. it's the ARIA behavior)

paulfd:
I think it could make sense to have the ampeg as a FlexEG too, if only to have a single codepath to check and optimize for all this.

paulfd:
@kinwie does this still happen? I have trouble seeing it on Reaper

kinwie:
@paulfd Using same example case as above, the release is cut shorter (sfizz - DropZone - sforzando)
![Untitled](https://user-images.githubusercontent.com/65444898/112020608-0feac780-8b63-11eb-80c2-1272eb1292ec.png)

jpcima:
A simple solution is to add a final phase that finishes the envelope in linear.

PR #380, opened 2020-08-18
End disabling regions

### LINKED ISSUE #365: Adding end=-1 or ampeg_sustain=0 to *silence region make the voice not counted
Adding `end=-1` or `ampeg_sustain=0` to a wav or flac file cause the region not played and the voice also not counted (0).

This should be also behave the same for `*silence` region. At this moment, the voice still counted. So for some usage combination with `on_lo/hiccN`, the voices of the `*silence` region will get stuck and never released.

Maybe also for other `*sample`, like the `*sine` `*saw` etc...

paulfd:
Closes #365

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -92,7 +92,11 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             offsetCC[opcode.parameters.back()] = *value;
         break;
     case hash("end"):
-        setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
+        if (opcode.value == "-1") {
+            disabled = true;
+        } else {
+            setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
+        }
         break;
     case hash("count"):
         setValueFromOpcode(opcode, sampleCount, Default::sampleCountRange);
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -92,11 +92,7 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             offsetCC[opcode.parameters.back()] = *value;
         break;
     case hash("end"):
-        if (opcode.value == "-1") {
-            disabled = true;
-        } else {
-            setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
-        }
+        setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
         break;
     case hash("count"):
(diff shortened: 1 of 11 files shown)


src/sfizz/Region.cpp
```diff
@@ -92,7 +92,11 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode)
             offsetCC[opcode.parameters.back()] = *value;
         break;
     case hash("end"):
-        setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
+        if (opcode.value == "-1") {
+            disabled = true;
+        } else {
+            setValueFromOpcode(opcode, sampleEnd, Default::sampleEndRange);
+        }
```
jpcima: We probably want to re-enable the region if sampleEnd is reset to a positive value after -1.
`<region>end=-1 end=10000`

Here I don't think the `disabled` boolean to be a great idea, I'd rather have `bool disabled() const` to check for `sampleEnd` where the member `sampleEnd` is changed to an optional.

Btw does `end` not also disable the voice if 0 ? I recall it to be the case.
Hence we will want to check `disabled` = `end` is set, and is <= 0.

paulfd: Ok then I guess the best is to make it an int64, and create this `disabled()` function.

paulfd: In the end I chose to keep the type. ARIA complains when sampleEnd==0 but still disables the region, so we can just do the same.

Otherwise I guess it'd be better to audit the uses of int64/uint32 in file and loop lengths. libsndfile uses int64 for the file length but the wav chunks are uint32.

src/sfizz/Synth.cpp
```diff
@@ -494,7 +494,7 @@ void sfz::Synth::finalizeSfzLoad()
                 continue;
             }
 
-            region->sampleEnd = std::min(region->sampleEnd, fileInformation->end);
+            region->sampleEnd = min(region->sampleEnd, fileInformation->end);
```
jpcima: What is the reason for this edit here?

paulfd: It's a leftover but I can revert it.

paulfd:
The latest should match the behavior described in #365.

I amended the rate transformation methods for the cases where the `timeInSeconds` is exactly `0`, since by default the decaying envelope for example was actually quite long (around 1 000 samples).

I also believe this could fix #373 since the value for the sustain is taken as-is, and we don't add `virtuallyZero` to it anymore.

jpcima (approved):

PR #381, opened 2020-08-19
Volume matching

### LINKED ISSUE #367: Request to reduce main volume to -3 or -4 or -6 dB
Some sfz instruments with normalized samples clipped the DAW track when just load and play with current default volume at 0 dB

jpcima:
This matches volumes to ARIA exactly. Fixes #367
It's tried with waveforms, mono samples, stereo samples.

It does so by connecting CC7 to region amplitude, and initializing the CC to 100, with concave curve.
The CC10 will be connected with pan, with default to middle, with bipolar curve.

The stereo samples needed +3dB, because the succession of pan and position attenuate by -6dB in total.

### DIFF
--- a/src/sfizz/Config.h
+++ b/src/sfizz/Config.h
@@ -103,10 +103,14 @@ namespace config {
     // Wavetable constants; amplitude values are matched to reference
     static constexpr unsigned tableSize = 1024;
     static constexpr double tableRefSampleRate = 44100.0 * 1.1; // +10% aliasing permissivity
-    static constexpr double amplitudeSine = 0.625;
-    static constexpr double amplitudeTriangle = 0.625;
-    static constexpr double amplitudeSaw = 0.515;
-    static constexpr double amplitudeSquare = 0.515;
+    /**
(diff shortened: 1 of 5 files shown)


paulfd (approved):

kinwie:
Yes, that will be much convenient.
If i may add another suggest, set the CC value to 96, not 100. I'm pretty sure ARIA use CC7 value 96, and also this is the default value in most DAW midi track afaik. For example, if my midi track volume has been modified, I just "Alt+click" its fader to set it to default and it will go to value 96. It is match to ARIA's -4 dB default (or exactly -4.188 dB).

jpcima:
Where does this value 96 come from?
MIDI 96 is -4.86 dB

paulfd:
96 is 3/4 of 128 maybe?

kinwie:
Ahh, sorry, just got home and re-check this.
@jpcima you are right. It should be midi cc value 100 to match ARIA's -4 dB.
My bad memory...getting old :)

PR #384, opened 2020-08-20
Simpler power/envelope follower

paulfd:
This one is much cheaper and should do the job well enough for what we use it for.

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -30,9 +30,6 @@ sfz::Voice::Voice(int voiceNumber, sfz::Resources& resources)
 
     gainSmoother.setSmoothing(config::gainSmoothing, sampleRate);
     xfadeSmoother.setSmoothing(config::xfadeSmoothing, sampleRate);
-
-    for (auto & filter : channelEnvelopeFilters)
-        filter.setGain(vaGain(config::filteredEnvelopeCutoff, sampleRate));
 }
 
 sfz::Voice::~Voice()
@@ -246,20 +243,19 @@ void sfz::Voice::setSampleRate(float sampleRate) noexcept
     gainSmoother.setSmoothing(config::gainSmoothing, sampleRate);
     xfadeSmoother.setSmoothing(config::xfadeSmoothing, sampleRate);
 
-    for (auto & filter : channelEnvelopeFilters)
-        filter.setGain(vaGain(config::filteredEnvelopeCutoff, sampleRate));
-
     for (WavetableOscillator& osc : waveOscillators)
         osc.init(sampleRate);
 
     for (auto& lfo : lfos)
         lfo->setSampleRate(sampleRate);
+
+    trackingFactor = samplesPerBlock / sampleRate * config::powerFollowerFactor;
 }
 
 void sfz::Voice::setSamplesPerBlock(int samplesPerBlock) noexcept
 {
     this->samplesPerBlock = samplesPerBlock;
-    this->minEnvelopeDelay = samplesPerBlock / 2;
+    this->trackingFactor = samplesPerBlock / sampleRate * config::powerFollowerFactor;
 }
 
 void sfz::Voice::renderBlock(AudioSpan<float> buffer) noexcept
@@ -736,10 +732,7 @@ void sfz::Voice::reset() noexcept
     floatPositionOffset = 0.0f;
     noteIsOff = false;
 
-    for (auto& f : channelEnvelopeFilters)
-        f.reset();
-
-    for (auto& p : smoothedChannelEnvelopes)
+    for (auto& p : meanChannelPowers)
         p = 0.0f;
 
     filters.clear();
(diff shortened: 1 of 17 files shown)


jpcima:
Ok I would like to suggest a bit more
- since we don't apparently need to analyze level of distinct channels, may we do a single analysis of stereo-linked input?
   (take in the sum of left and right channels absolute values, cf. gate/compressor `stlink`)
- may we buffer the frames using a fixed-size analysis window?

src/sfizz/Voice.cpp
```diff
@@ -869,16 +862,14 @@ void sfz::Voice::setupOscillatorUnison()
...
 
-    for (unsigned i = 0; i < smoothedChannelEnvelopes.size(); ++i) {
+    const float factor = buffer.getNumFrames() / samplesPerBlock * trackingFactor;
+    for (unsigned i = 0; i < meanChannelPowers.size(); ++i) {
         const auto input = buffer.getConstSpan(i);
-        for (unsigned s = 0; s < buffer.getNumFrames(); ++s)
-            smoothedChannelEnvelopes[i] =
-                channelEnvelopeFilters[i].tickLowpass(std::abs(input[s]));
+        const float meanPower = sfz::meanSquared(input);
+        meanChannelPowers[i] = meanChannelPowers[i] * (1 - factor) + meanPower * factor;
```
jpcima: The analysis signal is not windowed at the input.
In former experience, RMS value of raw input can be very imprecise.

paulfd: Windowed as in https://en.wikipedia.org/wiki/Hann_function ?

paulfd:
> Ok I would like to suggest a bit more
> 
>     * since we don't apparently need to analyze level of distinct channels, may we do a single analysis of stereo-linked input?
>       (take in the sum of left and right channels absolute values, cf. gate/compressor `stlink`)

Good point.

>     * may we buffer the frames using a fixed-size analysis window?

Won't this be expensive if we do it for each voice?

src/sfizz/Voice.cpp
paulfd: We need *some* indication or the power btw, it does not need to be too precise.

jpcima: Yes, I would once find the measurement to be very noisy, (this was another project out of sfizz)
and the curve next to unusable.
Imo, at the first look, I'm not feeling convinced by this method.
It's tied to the samples per block. If the signal is too low frequency, and the block size is short, the method seems likely to fail.

paulfd: It's supposed to be somewhat independent on the samples per block, the factor changes depending on the expected block size and the actual one. I will double check since I did this at 2 in the morning.

I don't see how windowing would help though, it will bias the average? The noisiness can be smoothed away, the issue being that attacks and releases are not tracked quickly enough. Apparently people also take the max of 2 parallel filters for such cases, with different responses, to track attacks fast and releases less so.

paulfd:
This implements the same thing, except I had a useless normalization by the maximum block size. I also put the combined filters for release and attacks. The test files are organ ranks (cuz I had them right here); "Plein jeux" is a relatively high-frequency mixture, "Bourdon 16" is a very low frequency quasi-pure sine wave, and "Doublette 2" is a very high frequency rank.

[There](https://box.ferrand.cc/random/envelope-follower.html)

It is not block-size independent indeed. I'll try to tweak it further, in the end I'd just like a minimum bias compared to sample-by-sample.

src/sfizz/Voice.cpp
paulfd: > It's supposed to be somewhat independent on the samples per block, the factor changes depending on the expected block size and the actual one. I will double check since I did this at 2 in the morning.

Disregard, as it stands you're right it's not independent since I had a no-op multiplication in the normalizations

paulfd:
I added a small correction to the coefficient calculation based on the assumption that:
1. The power for each sample in the block is the same as the mean power of the block ;
2. The coefficient per-sample is small

The results are [there](https://box.ferrand.cc/random/envelope-follower-corrected.html). It does not meaningfully improve or change the behavior of the envelope follower. I also tried to go deeper in the Taylor expansions for the block version but also there it does not seem to impact the algorithm positively. I suspect most of the bias stems from assumption 1, which appears a bit unavoidable if we want simpler per-block version for the follower. I'm tempted to label this "good enough for the purpose", what do you think?

I wonder also if #353 could just disable the envelope follower in the voices if the voice stealing algorithm does not use it?

jpcima:
It's made independent of block size. It's replaced with a fixed block size which is set in config.
Performance is unaffected, and result is looking OK at first.
Check it, merge if you want.

paulfd:
The memory cost is a bit higher but I'm OK with this solution. Note that it will still be indirectly dependent on sample rate, but as I said, I don't think it's a big issue given the current use we make of this. Thanks for the improvements :slightly_smiling_face:

PR #386, opened 2020-08-21
Modulations trigger and release

jpcima:
This adds a couple of desired features for implementing EG in MM.
- ability to pass the frame `delay` to make the modulator source startup frame-accurate +update LFO to use it
- allow MM sources to be notified from `Voice::release`, so it's able to signal all EG present in a voice to release themselves


paulfd:
Ok for me! Approving bugs on my phone...

PR #388, opened 2020-08-21
Add the Flex EG

jpcima:
Based on PR #386

This implements the Flex EG.

Most important to do:
- implement ARIA `egN_ampeg`
  When this is not set, and `egN` is connected to region amplitude, the region is subject to `ampeg` as its master envelope, and `egN` does not have a release under the default ampeg.
  When `egN_ampeg` ≠ 0, `ampeg` is overriden with `egN`.

More minor things to do:
- implement the modulations of levels and times.
- implement the `egN_dynamic` setting.
  - if `level` is modulated, reconfigure the current stage by setting FlexEG's target level to `level`, source level FlexEG's `currentLevel`, and `currentTime` to 0, `stageEndTime` to `stageEndTime-currentTime`
    (meaning to start over, as if the remainder of the stage is treated as a new curve. it's how ARIA does)
  - if `time` is modulated, update `stageEndTime`

### DIFF
--- a/src/sfizz/FlexEnvelope.cpp
+++ b/src/sfizz/FlexEnvelope.cpp
@@ -1,0 +1,206 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+// This code is part of the sfizz library and is licensed under a BSD 2-clause
+// license. You should have receive a LICENSE.md file along with the code.
+// If not, contact the sfizz maintainers at https://github.com/sfztools/sfizz
+
+/*
+   Note(jpc): implementation status
+
+- [ ] egN_points (purpose unknown)
+- [x] egN_timeX
+- [x] egN_levelX
+- [x] egN_shapeX
+- [x] egN_sustain
+- [ ] egN_dynamic
+*/
+
+#include "FlexEnvelope.h"
+#include "FlexEGDescription.h"
+#include "Curve.h"
+#include "Config.h"
+#include "SIMDHelpers.h"
+#include <absl/types/optional.h>
+
+namespace sfz {
+
+struct FlexEnvelope::Impl {
+    const FlexEGDescription* desc_ { nullptr };
+    float samplePeriod_ { 1.0 / config::defaultSampleRate };
+    size_t delayFramesLeft_ { 0 };
+
+    //
+    float stageSourceLevel_ { 0.0 };
+    float stageTargetLevel_ { 0.0 };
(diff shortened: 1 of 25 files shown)


src/sfizz/FlexEnvelope.cpp
```diff
@@ -0,0 +1,206 @@
...
+    if (skipFrames > 0) {
+        delayFramesLeft_ -= skipFrames;
+        fill(absl::MakeSpan(out.data(), skipFrames), 0.0f);
+        out.remove_prefix(skipFrames);
+        numFrames -= skipFrames;
+    }
+
+    // Envelope finished?
+    if (currentStageNumber_ >= desc.points.size()) {
+        fill(out, 0.0f);
```
paulfd, line 130: All envelopes end at 0? Or should it be `currentLevel_`?

src/sfizz/FlexEnvelope.cpp
```diff
@@ -0,0 +1,206 @@
...
+            isReleased_ = true;
+            currentFramesUntilRelease_ = absl::nullopt;
+        }
+
+        // Perform stage transitions
+        const bool isReleased = isReleased_;
+        while ((!stageSustained_ && currentTime_ >= stageTime_) ||
+               (stageSustained_ && isReleased)) {
+            if (!advanceToNextStage()) {
+                fill(out, 0.0f);
```
paulfd, line 149: Same

src/sfizz/FlexEnvelope.cpp
jpcima: Yes envelopes end at zero. Try in ARIA:
```
<region> sample=*sine
//eg15_amplitude=100
eg15_ampeg=1
eg15_time1=.1  eg15_level1=.25
eg15_time2=.15  eg15_level2=1
eg15_time3=.2  eg15_level3=.5 eg15_sustain=3
eg15_time4=.4 eg15_level4=1
```

The envelope will return to 0 after the final ramp.
This will create a very noticeable discontinuity.

Capture du 2020-09-13 07-59-55

paulfd: Hmm ok, that actually feels a bit strange

paulfd (approved):

paulfd:
I added tests and corrected a couple bugs they uncovered. Other than this it looks ok to me!

PR #393, opened 2020-08-26
Polyphony cleanups

### LINKED ISSUE #372: Release Trigger adjustment

paulfd:
This is mostly splitting up the various polyphony checks and overall voice starting logic in the synth into smaller methods, and using them on all events instead of only note ons.

This should close #372.

I'll add some inbuilt tests from the sfz test suite before merging I think, to check that we do cover relevant behavior on all fronts.

### DIFF
--- a/src/sfizz/SisterVoiceRing.h
+++ b/src/sfizz/SisterVoiceRing.h
@@ -128,14 +128,6 @@ struct SisterVoiceRing {
  */
 class SisterVoiceRingBuilder {
 public:
-    ~SisterVoiceRingBuilder() noexcept {
-        if (lastStartedVoice != nullptr) {
-            ASSERT(firstStartedVoice);
-            lastStartedVoice->setNextSisterVoice(firstStartedVoice);
-            firstStartedVoice->setPreviousSisterVoice(lastStartedVoice);
-        }
-    }
-
     /**
      * @brief Add a voice to the sister ring
      *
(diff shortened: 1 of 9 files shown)


jpcima (commented):
This system does not speak to me a lot, you can feel free to merge it.
Code looks fine from what I can tell.

kinwie:
I seems can't get the trigger=release to works in the newest build. Something's missing?

PR #397, opened 2020-08-31
Editor vstgui lv2

jpcima:
This implements the current UI made in VSTGUI into LV2.
It's decoupled from VST from now, and it can serve as transitional PR into the more elaborate UI.

Needs test for regressions.

There is a segfault in jalv.gtk 1.6.4 when exiting, but it may be the fault of jalv only.


jpcima:
Note: LV2 bundle had to be reorganized with a `Contents/` directory structure, so vstgui can access resources files without modifying the code.
This is going to disturb states which refer to the default instrument/scala files inside.

jpcima:
absl::variant is broken when the Travis builds set standard to C++17..
It has a link error with `ThrowBadVariantAccess`.
so I'll substitute it with something else.

jpcima:
Segfault problems are resolved in the latest commit.
It seems OK for merge.

PR #404, opened 2020-09-07
User interface update

jpcima:
New user interface. This has all functionality of the old one.

The new functionality will need support on sfizz side in a future work.
- memory usage
- controller values (read/write)
- key switches
- key ranges
- sending notes


paulfd:
Do you have a screen of how the interface is supposed to look?

editor/src/editor/Editor.cpp
```diff
@@ -284,405 +317,319 @@ void Editor::Impl::uiReceiveValue(EditId id, const EditValue& v)
...
+        lightTheme.valueBackground = { 0x2e, 0x34, 0x36 };
+        Theme darkTheme;
+        darkTheme.boxBackground = { 0x2e, 0x34, 0x36 };
+        darkTheme.text = { 0xff, 0xff, 0xff };
+        darkTheme.titleBoxText = { 0x00, 0x00, 0x00 };
+        darkTheme.titleBoxBackground = { 0xba, 0xbd, 0xb6 };
+        darkTheme.icon = darkTheme.text;
+        darkTheme.valueText = { 0x2e, 0x34, 0x36 };
+        darkTheme.valueBackground = { 0xff, 0xff, 0xff };
+        Theme& defaultTheme = lightTheme;
```
paulfd, line 358: Do we expect to change the theme at runtime later on?

AlexMitchellMus:
Further to the features needed in sfizz backend:
* output level meter

jpcima:
Current UI looks like this on the main panel.

Capture du 2020-09-10 20-40-50

paulfd:
With my system it did not look like this at all. I embedded the "logo" font and used it, but we can change :slightly_smiling_face: 
Screenshot from 2020-09-11 03-25-05

In the logo it's fine but in the interface it looks slightly childish I think.

paulfd (approved):

alcomposer:
Possibly use the text logo as opposed to the icon?

AlexMitchellMus:
Roboto font family is a very high quality open font. Example of it used in sfizz. I think this looks much better, especially for numbers.
ABeeZee font (current) is a children's learning font? I don't think this is the best place for it.

roboto

## NOTES
* Remove icons for `open` & `edit in external editor` and replace with downward chevron dropdown menu that contains both (and possibly more items)
* Have the file name area **itself** as an open dialogue LMB action. (This is expected behaviour- and nothing wrong in having open file in two places)
* Add downward chevron to keyswitch select & `Max` polyphony value
* I'm not sure about using `home` icon, is there a home in sfizz? I would suggest 4 icons:
Screenshot from 2020-09-04 19-43-52
* Icons arranged in most used to the centre of the interface: `Settings`, `Info`, `Controls`, `SFIZZ-XML / artwork`.
* Highlighting the icon on hover feels wrong- it could be a button area highlight?
* LMB over `sfizz logo` icon to open `About Sfizz` pop-up.
* Consider changing `KS` to `K-Switch`

editor/src/editor/Editor.cpp
jpcima: The purpose is to enable distinct UI parts to have different colors schemes.
If a part in the layout file contains `theme=darkTheme`, everything which is hierarchized under it is going to be processed under this theme.
The prime example is the header bar, which is processed in dark theme.

jpcima:
- +1 for roboto font
- if you add a picture, include a 2x variant and run it through optipng

>I'm not sure about using home icon, is there a home in sfizz? I would suggest 4 icons:

Imo, large text+icon is excessive at the top right.
Remove sfizz's large icon to leave just the text, then change the home icon to sfizz icon.

paulfd:
>     * +1 for roboto font
> 
>     * if you add a picture, include a 2x variant and run it through optipng
> 
> 
> > I'm not sure about using home icon, is there a home in sfizz? I would suggest 4 icons:
> 
> Imo, large text+icon is excessive at the top right.
> Remove sfizz's large icon to leave just the text, then change the home icon to sfizz icon.

It's a bit sad to put unfa's work small no? Maybe an engraving in the background...

jpcima:
>Maybe an engraving in the background...

I'll be ok with such a choice.
Personnally, I am disturbed by the sfizz text not being center-aligned in the box.
Or perhaps, expand the hbox a bit more and space the buttons, as in Alex's mockup above.

editor/src/editor/GUIComponents.cpp
```diff
@@ -427,3 +427,21 @@ void SValueMenu::onItemClicked(int32_t index)
...
+{
+    hoverColor_ = color;
+}
+
+CMouseEventResult CHoverButton::onMouseEntered (CPoint& where, const CButtonState& buttons)
+{
+    backupColor_ = getTextColor();
+    setTextColor(hoverColor_);
+    return CTextButton::onMouseEntered(where, buttons);
+}
```
jpcima, line 453: There should be a cleanest way that you can cheat it, if you prefer to avoid messing with color values.
- `setHoverColor` and `onMouse*` set the dirty flag on the control
- override `CTextButton`'s draw method
```
if (hovered) {
    oldTextColor = textColor;
    textColor = myHoveredColor; // textColor is protected
}
CTextButton::draw;
if (hovered)
    textColor = oldTextColor;
```

Also, note that I went with the `S` prefix convention to distinguish these parts from vstgui.

paulfd: Hmm is it so much simpler? 

> Also, note that I went with the `S` prefix convention to distinguish these parts from vstgui.

OK! Will change it later.

jpcima: Not really but it's for cleaner API.
If I'm doing `setTextColor(green)`, I would not expect that `getTextColor()` returns red depending if the mouse is over the part or isn't.
Regarding this button, I'd be tempted renaming it as `STextButton` (a `CTextButton` with extra features)

paulfd: That's a good point.

AlexMitchellMus:
WIP Mock-up of left two panels:
Screenshot from 2020-09-11 19-58-33
## Notes:
* Roboto font
* sfizz brand text center
* Grey highlight under tab icon when mouse hover
* Active dot under tab icon for active tab
* Left-Right chevron for moving to next-previous SFZ script in folder
* Downward chevron for file options dropdown: `load`, `edit in external editor`, `duplicate & edit` etc
* Green SFZ engine status indicator. GREEN= all good. RED Cross Icon= error. LOADING throbber= loading.
* LMB on status indicator to bring up per instance sfizz engine log.

jpcima:
I updated the file dialog of vstgui according to Robin's advice, this is expected to work in the Ardour6 binary bundle.

PR #410, opened 2020-09-08
Modulation matrix tweaks

paulfd:
Hiya,

I'm checking out the overall performance and regressions that possibly happened in the past months. I've identified some areas where we could improve, and I've started to chop it into smaller commits (e.g. #405 and #353 to disable completely the envelope follower). This is in the same vein.

The `flags()` function and the `operator==` are super hot code paths, being called literally billions of times over a couple minutes of rendering according to valgrind when there are many modulations/ccs/lfos in the file. I've tried here to:
- Memorize the flags on `ModKey` construction, so we avoid the big switch statement.
- Allow inlining of the operator by putting it in the header. I would have thought that LTO would catch such things but apparently not...

I did not test on other platforms yet but on my machine here it reduces the average callback duration by 25% on DSmolken's Big Rusty Drums. The binary size does not increase significantly.

https://box.ferrand.cc/sfizz-reports/2020-09-08-mm-tweaks.html

### DIFF
--- a/src/sfizz/modulations/ModKey.cpp
+++ b/src/sfizz/modulations/ModKey.cpp
@@ -7,7 +7,6 @@
 #include "ModKey.h"
 #include "../Debug.h"
 #include <absl/strings/str_cat.h>
-#include <cstring>
 
 namespace sfz {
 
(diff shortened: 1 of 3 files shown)


jpcima:
It's good. Although, I was a little surprised by the observation.
If that kind of thing is important, it's possible to research some other improvements.

- when modulations are looked up by `ModKey` into the MM (cf. `getModulationByKey`), we have looking up into a map;
as apposed to using an index (`TargetId`). One could get more speed by caching the index, but keeping in mind that modifying targets or sources will invalidate.

- one can pre-normalize percent depths in MM to 0-1 to save some multiplies.
  (if you are tempted to do, please merge Flex EG first)

- we can entirely skip running "discarded" sources and targets which are per-voice
  (this creates a problem if there is some cycles C1->C2->C3, and the modulation receiver will compute the target in C1 and C3 skipping C2, this will create a glitch. however, I couldn't yet think of one case which could cause this; it calls to be careful if there is)

- disregard any items of MM which are depth=0, vel2depth=0, and depth_oncc=0.
(atm, vel2depth and oncc are not handled, but I wish these to be in MM)

- every time processing a voice starts or ends, MM goes over all sources and checks their flag and region id.
one could pre-cache this such that: (1)no need to go through the whole list (2)no need to check the condition on flags
(also, same thing at global level)
cf. MM `beginVoice`, `endVoice`, `beginCycle`, `endCycle`

PR #413, opened 2020-09-09
Regions on the same note but with different groups do get muted

paulfd:
This happens e.g. with keyswitches. This behavior is slightly different from Cakewalk's apparently, I'll let @kinwie detail :)

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -713,9 +713,9 @@ bool sfz::Voice::checkOffGroup(const Region* other, int delay, int noteNumber) n
     if (region == nullptr || other == nullptr)
         return false;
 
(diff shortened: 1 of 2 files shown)


kinwie:
Yes, tested. Working as expected and enhanced Cakewalk's behavior :+1:

kinwie:
The original RGC sfz player and Cakewalk DropZone has a nice self-choking behavior (same group= and off_by number) which preserved its polyphonic voicing per note. But it has one flaw, which it can't choke the same note when some other functions combined, e.g. using CC 4 for hihats control.

So this sfizz enhanced behavior solved that flaw and will works perfectly fine with any specific behavior of sfz files made for other sfz players, e.g. Cakewalk players or ARIA/sforzando, and even works fine for the old soundfont behavior

PR #421, opened 2020-09-16
Store the initial values of controllers

jpcima:
Keep the initial values of controllers for the active instrument.
Having this data will permit to make the UI aware about default values for the CC sliders.


paulfd:
So purely philosophical, but I guess the `init` functions could be private? The getter on the other hand could even have a public API I guess.

paulfd (approved):

PR #422, opened 2020-09-16
Filter and EQ modulation targets

paulfd:
This registers filters and EQ modulation targets, as well as the bindings to midi CC. I also removed the filter pools and EQ pools to put the filters and EQ objects directly in the voices.

Missing some testing although things seem to work OK for now.

### DIFF
--- a/src/sfizz/modulations/ModKey.cpp
+++ b/src/sfizz/modulations/ModKey.cpp
@@ -87,6 +87,18 @@ std::string ModKey::toString() const
         return absl::StrCat("Pitch {", region_.number(), "}");
     case ModId::Volume:
         return absl::StrCat("Volume {", region_.number(), "}");
+    case ModId::FilGain:
+        return absl::StrCat("FilterGain {", region_.number(), "N=", params_.N, "}");
+    case ModId::FilCutoff:
+        return absl::StrCat("FilterCutoff {", region_.number(), "N=", params_.N, "}");
+    case ModId::FilResonance:
+        return absl::StrCat("FilterResonance {", region_.number(), "N=", params_.N, "}");
+    case ModId::EqGain:
+        return absl::StrCat("EqGain {", region_.number(), "N=", params_.N, "}");
+    case ModId::EqFrequency:
+        return absl::StrCat("EqFrequency {", region_.number(), "N=", params_.N, "}");
+    case ModId::EqBandwidth:
+        return absl::StrCat("EqBandwitdth {", region_.number(), "N=", params_.N, "}");
 
     default:
         return {};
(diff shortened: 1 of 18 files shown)


jpcima:
Very nice. After a quick look, it looks good to me.
It's a bit difficult to read the diff in places. (good thing to have a lot of red, certainly not a complaint here :+1:)

Regarding removed tests, it's a good case for rewriting with `RegionCCView` in `TestHelpers`. (cf. RegionT various tests)

paulfd:
Hmm I forgot about the helper but I had in mind rewriting it xD Anyway I chose to compare the dot graph in the end because it would have gotten quite heaver and verbose (you need a target and thus a RegionCCView for each filter target/filter number, e.g. you would have alot of
```
region.parseOpcode({ "cutoff_oncc45", "4.2" });
REQUIRE(RegionCCView(region, ModKey::createNXYZ(ModId::FilCutoff, region.getId(), 0)).at(...).value == ...)
```
which looked a bit verbose.
If you think it's necessary I'll rewrite them, sure :slightly_smiling_face: 
The dot graph is a bit more readable in this sense, and it tests the end-to-end processing and creation of connections (from parsing the file to the state of the modulation matrix).
It does not test the connection list within the region, but I think these are unused beyond the mod matrix initialization.

jpcima:
> It does not test the connection list within the region, but I think these are unused beyond the mod matrix initialization.

They are not used after the initialization. Their purpose is to create an intermediate state that serves to prepare the CC keys.

As a note, regarding the CC, I wish to go back on a particular design decision: `value` should be removed from the mod key, and instead changed into `sourceDepth`. This way if there exist some CCs with identical everything except depth, there is only need to compute one of them.

paulfd:
> As a note, regarding the CC, I wish to go back on a particular design decision: `value` should be removed from the mod key, and instead changed into `sourceDepth`. This way if there exist some CCs with identical everything except depth, there is only need to compute one of them.

Sure !

paulfd:
Seems to work fine in my tests.

src/sfizz/modulations/ModKey.cpp
```diff
@@ -87,6 +87,18 @@ std::string ModKey::toString() const
         return absl::StrCat("Pitch {", region_.number(), "}");
     case ModId::Volume:
         return absl::StrCat("Volume {", region_.number(), "}");
+    case ModId::FilGain:
+        return absl::StrCat("FilterGain {", region_.number(), ", N=", params_.N, "}");
```
jpcima, line 91: The displayed filter and EQ number should be from 1.
It's suggested to increment, same as envelope and LFO.

paulfd: I think I changed it in the next PR.

paulfd: Yes I did in 83cd4b631796875cddf5d6c3a358bdf84eaeaea6

PR #427, opened 2020-09-19
Aligned vector

jpcima:
This possibly fixes the alignment problem that occurs on ARM machines, try it.
This forces `std::vector` to allocate with 16 byte minimum alignment in the oversampler, regardless of the platform.


paulfd:
Thanks it's great ! If the change I made is OK for you I guess we can merge it already.

jpcima:
Sure, it's fine to merge it.

PR #435, opened 2020-09-22
Ring modulation

jpcima:
Implements ring modulation, as described in the cakewalk book

### DIFF
--- a/src/sfizz/Defaults.h
+++ b/src/sfizz/Defaults.h
@@ -68,6 +68,7 @@ namespace Default
     constexpr Range<int> oscillatorMultiRange { 1, config::oscillatorsPerVoice };
     constexpr float oscillatorDetune { 0 };
     constexpr Range<float> oscillatorDetuneRange { -9600, 9600 };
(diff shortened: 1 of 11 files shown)


jpcima:
Ring modulation example
```
<control>
set_hdcc21=0.5

<region>
sample=*saw
oscillator=on
oscillator_mode=0
oscillator_multi=2
oscillator_detune=25
oscillator_mod_depth=100
oscillator_mod_depth_oncc21=100
```

paulfd:
Does the book specify if we can have different waveforms for the multiple oscillators?

jpcima:
There is no indication about this whatsoever, although this is indeed important to modulation synthesis.
Under rapture, you can process RM, but only by user interface, not via SFZ oscillator opcodes; modulator and carrier can select different waves.
I make this choice by naively following what unison did; assign identical waves to all the oscillators.
For distinct waves, it will require a SFZ extension.

jpcima:
superseded by #436 (which includes a fix for unison)

PR #436, opened 2020-09-22
Add FM synthesis with 2 operators

jpcima:
FM example
```
<control>
set_cc21=100

<global>
volume=-3.0

<region>
sample=*sine
oscillator=on
oscillator_mode=2
oscillator_multi=2
oscillator_detune=5
oscillator_mod_depth=100
oscillator_mod_depth_oncc21=1000
```

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -661,6 +661,8 @@ void sfz::Voice::fillWithGenerator(AudioSpan<float> buffer) noexcept
     } else {
         const auto numFrames = buffer.getNumFrames();
 
+        buffer.fill(0.0f);
+
         auto frequencies = resources.bufferPool.getBuffer(numFrames);
         if (!frequencies)
             return;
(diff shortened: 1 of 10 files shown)


src/sfizz/Voice.cpp
```diff
@@ -661,6 +661,8 @@ void sfz::Voice::fillWithGenerator(AudioSpan<float> buffer) noexcept
     } else {
         const auto numFrames = buffer.getNumFrames();
 
+        buffer.fill(0.0f);
```
paulfd: I think it's filled at the beginning of renderBlock?

paulfd (approved):

src/sfizz/Voice.cpp
jpcima: Voices are rendered into a temporary buffer. This could go, however this leads to another problem:
Synth.cpp `tempSpan->fill(0.0f);` -> called once per cycle
The `tempSpan` will have to be cleared each time before the voice rendering.

PR #438, opened 2020-09-23
Pitch and Filter EG v1

jpcima:
This implements v1 other EG, adds it to the MM.

Remaining to do:
- [ ] `ampeg` not matrixed yet (this can wait)
- [x] do `vel2depth` opcodes

### DIFF
--- a/src/sfizz/CCMap.h
+++ b/src/sfizz/CCMap.h
@@ -38,6 +38,8 @@ public:
     CCMap(CCMap&&) = default;
     CCMap(const CCMap&) = default;
     ~CCMap() = default;
+    CCMap& operator=(CCMap&&) = default;
+    CCMap& operator=(const CCMap&) = default;
 
     /**
      * @brief Returns the held object at the index, or a default value if not present
@@ -105,7 +107,7 @@ private:
     // typename std::vector<std::pair<int, ValueType>>::iterator begin() { return container.begin(); }
     // typename std::vector<std::pair<int, ValueType>>::iterator end() { return container.end(); }
 
-    const ValueType defaultValue;
+    ValueType defaultValue;
     std::vector<CCData<ValueType>> container;
     LEAK_DETECTOR(CCMap);
 };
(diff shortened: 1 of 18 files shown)


jpcima:
Added the support of vel2depth, by some MM changes
Fully supersedes the old PR #163

src/sfizz/modulations/ModMatrix.cpp
```diff
@@ -398,6 +406,12 @@ float* ModMatrix::getModulation(TargetId targetId)
                 source.bufferReady = true;
             }
 
+            float sourceDepth = sourcesPos->second.sourceDepth_;
+            if (sourceFlags & kModIsPerVoice) {
+                const float velToDepth = sourcesPos->second.velToDepth_;
+                sourceDepth += triggerValue * velToDepth;
+            }
```
paulfd, line 413: Probably bikeshedding in the grand scheme of things, but this could be computed and stored in `beginVoice` so that we're not recomputing it each cycle.

src/sfizz/CCMap.h
```diff
@@ -105,7 +107,7 @@ class CCMap {
     // typename std::vector<std::pair<int, ValueType>>::iterator begin() { return container.begin(); }
     // typename std::vector<std::pair<int, ValueType>>::iterator end() { return container.end(); }
 
-    const ValueType defaultValue;
+    ValueType defaultValue;
```
paulfd, line 110: I'd expect that moving from the container just move its contents but keeps the default value. I think you encountered a move somewhere ? In this case we probably need to use a non-default move constructor/assignment.

jpcima: From memory it all stems from this use case
- want to make `absl::optional` of `EGDescription` which contains `CCMap` members
- it needs `operator=`
- `operator=` is auto-deleted when the class has the copy-ctor defined
- it can't default-implement `operator=` when the member is const

src/sfizz/modulations/ModMatrix.cpp
jpcima: I thought of it but then just considered the result of a single multiply is not worth caching.

paulfd (approved):

Issue #440, opened 2020-09-25
Normalize boolean values parsing


jpcima:
The other sfz implementations have different ways of expressing their booleans depending on opcode, and it's not compatible.
Probably it's best that we let `readBooleanFromOpcode` accept both forms for boolean-accepting opcodes, and then adapt current opcodes to use it.
- ARIA: `0` or `1`
- Cakewalk: `off` or `on`

PR #445, opened 2020-09-26
Add egN_ampeg

jpcima:
This implements opcode `egN_ampeg`, allowing a flex EG to override the ampeg.

Notes:
- when multiple flex EG have `egN_ampeg`, the lowest one becomes the amp EG
- ampeg has been matrixed; depending if the sfz uses flex ampeg, either v1 ampeg or flex gets connected to amplitude
- under ARIA, one can have both `egN_ampeg` and `egN_amplitude` in a region, and they both take effect.
That means the equivalent of 2 independent connection (egN->amplitude) in the MM which must coexist.
Since we can't represent this, I created a special target `MasterAmplitude` which is reserved for ampeg and is always connected.
- this doesn't implement fast release, so I left in `Voice.cpp` some TODO comments "Flex AmpEG"

In this way, ARIA and sfizz work in equivalent way.

**Except**
There is a bug, which I'll do separately, it's that ARIA goes directly to the point after sustain on releasing the EG, whereas ours finishes to go through intermediate points first.

```
<region>
sample=*sine
eg15_ampeg=1
eg15_time1=.1  eg15_level1=.25
eg15_time2=.15  eg15_level2=1
eg15_time3=.2  eg15_level3=.5 eg15_sustain=3
eg15_time4=.4 eg15_level4=1
eg15_amplitude=100
```

On tapping the key, you can observe the problem on scope quite easily.
Capture du 2020-09-26 17-18-59

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -380,14 +380,16 @@ void sfz::Voice::amplitudeEnvelope(absl::Span<float> modulationSpan) noexcept
 
     ModMatrix& mm = resources.modMatrix;
 
+    // Amplitude EG
+    absl::Span<const float> ampegOut(mm.getModulation(masterAmplitudeTarget), numSamples);
+    ASSERT(ampegOut.data());
+    copy(ampegOut, modulationSpan);
+
     // Amplitude envelope
     applyGain1<float>(baseGain, modulationSpan);
     if (float* mod = mm.getModulation(amplitudeTarget)) {
         for (size_t i = 0; i < numSamples; ++i)
-            modulationSpan[i] = normalizePercents(mod[i]);
-    }
-    else {
-        ASSERTFALSE;
+            modulationSpan[i] *= normalizePercents(mod[i]);
     }
 
     // Volume envelope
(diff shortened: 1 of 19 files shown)


AlexMitchellMus:
Hi Jean,

I have tested the branch, but the test example below does not look like it's working:
```
<region>
sample=*sine
eg1_time1=0  eg1_level1=1
eg1_time2=1  eg1_level2=0
eg1_time3=1  eg1_level3=.5 eg1_sustain=3
eg1_time4=1  eg1_level4=1
eg1_ampeg=1
```
Screenshot from 2020-09-27 13-40-06

From Discord discussion- I would have expected for the plot to look like the one you provided? Maybe my FlexEG is wrong?
Capture_du_2020-09-26_10-36-57

jpcima:
@alcomposer stages with time=0 needed a special case, done in a137edc. Thanks for noticing.

src/sfizz/Voice.cpp
```diff
@@ -152,21 +151,29 @@ void sfz::Voice::release(int delay) noexcept
     if (state != State::playing)
         return;
 
-    if (egAmplitude.getRemainingDelay() > delay) {
-        switchState(State::cleanMeUp);
-    } else {
-        egAmplitude.startRelease(delay);
+    if (!region->flexAmpEG) {
+        if (egAmplitude.getRemainingDelay() > delay)
+            switchState(State::cleanMeUp);
```
paulfd, line 156: We're missing the `startRelease(delay)` here from the previous logic. Did it go elsewhere?

paulfd:
Other than the `startRelease` part I see no glaring issue. Thanks alot !

Btw I added stuff to the boolean opcode helper, is it what you meant in #440?

src/sfizz/Voice.cpp
jpcima: Yes because it's now supposed to be called from MM.
see `ADSREnvelopeSource::release`

jpcima:
It's indeed what I mean but it lacks some cases.
I recall that under ARIA, numbers ≠ 0 pass as `true` for booleans.
Preferably, this routine should parse numbers. (also I'd review all possible cases of cw/aria, and what happens for real value `0.1`)

src/sfizz/Voice.cpp
paulfd: OK I see.

paulfd (approved):

jpcima:
Regarding booleans, ARIA's behavior is surprising to say the least.
- `egN_ampeg`, any value enables it, no matter if 0, 1, on, off, or anything else

PR #451, opened 2020-09-27
Have the sequence start at the first item

### LINKED ISSUE #450: Seq_ bug : sfizz always starts from 2nd position
From this example, the first note play is always `*noise` :

```
<group>
seq_length=5
<region> seq_position=1 sample=*sine
<region> seq_position=2 sample=*noise
<region> seq_position=3 sample=*saw
<region> seq_position=4 sample=*tri
<region> seq_position=5 sample=*square
```

jpcima:
Fixes #450
@kinwie


kinwie:
Yes, checked, works fine now, also when combine with the lo/hirand. Thanks!

paulfd (approved):

PR #452, opened 2020-09-27
Make all the random opcodes bipolar

jpcima:
Reported by @kinwie. It's an inconsistency between random opcodes.
- apparently Cakewalk uses bipolar for everything
- ARIA uses unipolar
bipolar may be considered as more useful, and probably it makes the most sense.


paulfd (approved):

paulfd:
Yes I think I did not think too much about the behavior at the time and I had only `offset_random`, which I guess should still be unipolar?

PR #454, opened 2020-09-27
Attempt at building macOS universal for ARM

jpcima:


jpcima:
Homebrew doesn't have the ARM64 packages yet, aka "Apple Silicon", so this PR hangs while waiting it to happen.

PR #459, opened 2020-09-28
Move Doxygen scripts

redtide:
The purpose of this PR is to move both `.travis` and `doxygen` directories in the `scripts` one, to keep all scripts related data in one place, plus removing some old and unused Doxygen related scripts, including also the related Travis job based on the `update_dox.sh` build script (which at some point wasn't triggered anymore, see #164).


jpcima:
For now travis is kept like it is, only dox changes are kept.
A backup branch is kept at `jpcima/moved-scripts.bak`

Issue #395, opened 2020-08-30
Loop ticking on pitched keys


kinwie:
In attached example file, the looped sample played fine at its root key. But when pressing other (pitched) keys, the loop is ticking in sfizz
[Loop_ticking_on_pitched_keys.zip](https://github.com/sfztools/sfizz/files/5146335/Loop_ticking_on_pitched_keys.zip)

kinwie:
I think this problem is that, sfizz not reading well the loop end point in the metadata, if the value is maximum (same as the sample's end).
I edit the the wav in the attached example file, move the loop-start and loop-end less 1 point, the ticking is gone.

jpcima:
I believe there's a bigger problem at play here. The loop code suffers of an issue in relation with the block rendering.
Not tried yet, but based on my current understanding.

To explain it shortly, suppose there is a audio loop of size 100 frames.
- If the audio host asks sfizz to serve 200 frames, 
      - block 0-100 will be filled with our loop 0-100 frames
      - block 100-200 will be filled again with our loop 0-100 frames
- Now if the audio host asked to serve 300 frames
      - block 200-300 would be filled **wrongly** with input audio range 100-200
        because code must offset the input index back by 100 **2 times**, but the buggy code does not more that 1

So that's possibly the problem, if you have a buffer size too large, the multiple loops that occur within a single buffer don't offset correctly after the first one.

paulfd:
I think this can be closed now, considering #464

kinwie:
Yes, sorry, just have this checked. The tick is gone. Thanks

PR #462, opened 2020-09-29
Fix the case when multiple loops occur in one buffer

jpcima:
#395 
@kinwie, please check if this helps to fix looping problems

### DIFF
--- a/src/sfizz/MathHelpers.h
+++ b/src/sfizz/MathHelpers.h
@@ -139,6 +139,20 @@ constexpr T clamp(T v, T lo, T hi)
     return max(min(v, hi), lo);
 }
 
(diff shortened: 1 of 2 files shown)


kinwie:
@jpcima No, it still not yet fix the problem

jpcima:
Ok, this one is important regardless, I'll merge it.
Note: there isn't a SIMD acceleration for this one because there isn't a modulo or division in `epi32` variant.

@kinwie I verified the loop end point, the value is correct.
Also as I try reproducing the problem with the organ sample, I don't hear the ticking effect.

kinwie:
I just test the build 1567, the loop still ticking

Issue #237, opened 2020-05-16
Opcode request : loop_crossfade


kinwie:
It's a sfz version 2 opcode for crossfading within sample's loop points. This opcode will be very useful for translating from other sample format into sfz, where its numbers of available free and good quality instruments keep growing these days. With this opcode, then we don't need to destructively edit the sample files, but instead just done at programming level inside sfz, so we can honour any restriction to modify sample files.

kinwie:
Just checked this also. The crossfade is working good. Thank!

PR #464, opened 2020-09-30
Loop xfade

jpcima:
#237
Support loop crossfading
- enforces a default minimum, set to 1ms, should resolve the ticking problem (#395)
- `loop_crossfade` opcode
- subdivide the render block into partitions
  - a partition is a block of frames where the indices increase monotonically, and also alternates between "normal" and "xfading" sections.
  - in normal mode, it's processing as usual
  - in xfading mode, two segments of the source audio are interpolated, and faded together with xfin/xfout curves
     the xfout section is that goes up to `loopend`, and xfin going up to loopstart. the size of these sections is determined by the amount `loop_crossfade`, it is rounded to the nearest frame.

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -588,8 +588,7 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
         partitionStarts = absl::MakeSpan(const_cast<int*>(starts), 1);
         partitionTypes = absl::MakeSpan(const_cast<int*>(types), 1);
         numPartitions = 1;
-    }
-    else {
+    } else {
         for (auto& buf : partitionBuffers) {
             buf = resources.bufferPool.getIndexBuffer(numSamples);
             if (!buf)
@@ -600,11 +599,6 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
         // Note: partitions will be alternance of Normal/Xfade
         //       computed along with index processing below
     }
-
-    // loop crossfade settings
-    constexpr int loopXfadeUseCurves = 2; // 0: linear
-                                          // 1: use curves 5 & 6
-                                          // 2: use S-shaped curve
 
     // index preprocessing for loops
     if (isLooping) {
@@ -699,17 +693,17 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
             {
                 // compute out curve
                 absl::Span<float> xfCurve = xfTemp2->first(ptSize);
-                IF_CONSTEXPR (loopXfadeUseCurves == 2) {
+                IF_CONSTEXPR (config::loopXfadeCurve == 2) {
                     const Curve& xfIn = getSCurve();
                     for (unsigned i = 0; i < ptSize; ++i)
                         xfCurve[i] = xfIn.evalNormalized(1.0f - xfCurvePos[i]);
                 }
-                else IF_CONSTEXPR (loopXfadeUseCurves == 1) {
+                else IF_CONSTEXPR (config::loopXfadeCurve == 1) {
                     const Curve& xfOut = resources.curves.getCurve(6);
                     for (unsigned i = 0; i < ptSize; ++i)
                         xfCurve[i] = xfOut.evalNormalized(xfCurvePos[i]);
                 }
-                else IF_CONSTEXPR (loopXfadeUseCurves == 0) {
+                else IF_CONSTEXPR (config::loopXfadeCurve == 0) {
                     // TODO(jpc) vectorize this
                     for (unsigned i = 0; i < ptSize; ++i)
                         xfCurve[i] = clamp(1.0f - xfCurvePos[i], 0.0f, 1.0f);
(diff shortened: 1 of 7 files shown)


jpcima:
Loop xfade update:
- the xfin/xfout curve model was reported to still produce ticking; this was because of the volume variation at fade center that makes overall volume go up during fade sections and then down in normal sections.
- the linear curve fixes this as been reported "not sounding as good"
- the s-curve (current) is a compromise curve that preserves the 50% at center property, and have a smoothed amplitude curve

see also https://en.wikipedia.org/wiki/Fade_(audio_engineering)#S-curve
s

EDIT more curve possibilities with parameter control
https://gist.github.com/jpcima/cd3d7b605f2391ef21dbb272a5431bbe

src/sfizz/Voice.cpp
```diff
@@ -581,25 +688,112 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
...
+                        xfCurve[i] = xfIn.evalNormalized(xfInCoeff[i]);
+                }
+                else IF_CONSTEXPR (loopXfadeUseCurves == 0) {
+                    // TODO(jpc) vectorize this
+                    for (unsigned i = 0; i < applySize; ++i)
+                        xfCurve[i] = clamp(xfInCoeff[i], 0.0f, 1.0f);
+                }
+                // apply in curve
+                fillInterpolatedWithQuality<true>(
+                    source, xfInBuffer, xfInIndices, *coeffs, xfCurve, quality);
```
paulfd: Should it be `xfInCoeff` here instead of `*coeffs`?

paulfd:
Here are some proposed patches

[Loop xf patches.zip](https://github.com/sfztools/sfizz/files/5332133/Loop.xf.patches.zip)

We could also precompute the loop crossfade step on voice start and use a SIMD helper to fill in the coeffs at
```c++
for (unsigned i = 0; i < ptSize; ++i) {
    float pos = ptIndices[i] + ptCoeffs[i];
    xfCoeff[i] = (pos - loop.xfOutStart) / loop.xfSize;
}
clampAll(xfCoeff, 0.0f, 1.0f);
```

paulfd:
Overall I think it's good, but we can probably improve it a little down the line. I like the idea of overlaying and partitioning. I wonder if we could make an "almost" branchless version that could be easier to vectorize. 

After the next release I plan to work around this to remove some bottlenecks on ARM systems.

paulfd (approved):

src/sfizz/Voice.cpp
jpcima: No, it should be `ptCoeffs->subspan(applyOffset)`.
these other "coeff" tables are indeed confusing, I have to rename these.
They are the position of the fade on the curve.

jpcima:
Branches outside of the frame loop are not usually that significant.
What bottlenecks do you think of?

paulfd:
Same as with the pan lookup, it's better to pool all integer operations into blocks on some ARM platforms apparently. But I'll check exactly where and why so for now let's not prematurely optimize unneeded things :slightly_smiling_face:

src/sfizz/Voice.cpp
```diff
@@ -581,25 +688,113 @@ void sfz::Voice::fillWithData(AudioSpan<float> buffer) noexcept
...
+                    for (unsigned i = 0; i < ptSize; ++i)
+                        xfCurve[i] = xfOut.evalNormalized(xfCurvePos[i]);
+                }
+                else IF_CONSTEXPR (loopXfadeUseCurves == 0) {
+                    // TODO(jpc) vectorize this
+                    for (unsigned i = 0; i < ptSize; ++i)
+                        xfCurve[i] = clamp(1.0f - xfCurvePos[i], 0.0f, 1.0f);
+                }
+                // apply out curve
+                if (0)
```
paulfd: Missed this one; if it's entirely dead code maybe we can just remove it !

jpcima:
Regarding this
> We could also precompute the loop crossfade step on voice start and use a SIMD helper to fill in the coeffs at

The crossfade segment may go up to 1s of signal (by current limits)
That's probably not something we want done in one go while in RT, as this is sure to create spikes in CPU usage.

Then one could argue in favor of making it at sample loading time; in this case it needs one precalculation per (sample × loop_crossfade). This complicates the loader, and it increases the memory usage and work load.

Since these 2 scenarios are not ideal, I select the current solution.
The larger the crossfade set in sfz, the larger is the CPU price to pay for it.
A minimum 1ms makes the xfade process kick in at some regular interval, over a period of ~44 frames (44kHz) at the keycenter. This creates a worst case, in a multiple voice playback, at an instant where all xfade segments happen to pile up over the same interval.

Regarding this worst case, there's a property worth noting: since this rounds `loop_crossfade` to the nearest integer, this means that the 2 source signals are interpolated at equal coeffs. Hence, both have identical interpolation polynomials at `quality>=2`. We could take advantage of this for interpolation both at once, and reducing instruction latency.

paulfd:
I had in mind to compute only the step `1 / loopXfadeSize` and use it as a multiplier instead of a division, or for a `linearRamp` call that spans exactly the size of the needed buffer in the callback. I agree with the analysis of the 2 scenarios, I had the same line of thought :)

jpcima:
On the patch suggestions

All good for me, except `0004-Clamp-the-coeffs-and-vectorize-the-linear-case.patch`.
Evaluating the curve, the clamping operation is already done at curve level, making it redundant.
I certainly don't mind though, if curves can have a unsafe evaluation function that skips the bounds check.

paulfd:
> On the patch suggestions
> 
> All good for me, except `0004-Clamp-the-coeffs-and-vectorize-the-linear-case.patch`.
> Evaluating the curve, the clamping operation is already done at curve level, making it redundant.
> I certainly don't mind though, if curves can have a unsafe evaluation function that skips the bounds check.

No prob, you had a comment to vectorize this one so I obliged but it's fine!

jpcima:
note: we need a "rsubtract1" helper that accepts a single value as LHS.
Then we can properly write the vectorization.

PR #484, opened 2020-10-07
accept samplesPerBlock up to 8192 (inclusive)

atsushieno:
`ASSERT(samplesPerBlock < config::maxBlockSize)` (where `maxBlockSize` is 8192) means 
we cannot set 8192. But we don't want 8191 as the actual maximum size.


jpcima:
Thanks, I've also added your name to the list of contributors.

PR #490, opened 2020-10-12
Fix build on Android 32-bit arm: hardfp is not available

atsushieno:
context: https://github.com/sfztools/sfizz/commit/d783ba6c8715ebb19124d713bf3223c9f4946691#r43170024

This fixes build errors like this on Android 32-bit arm architecture:
`error: lv2/CMakeFiles/sfizz_lv2.dir/sfizz.c.o uses VFP register arguments, output does not`


jpcima:
I just discovered this flag `-mhard-float`.
https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/ARM-Options.html

This appears to create adequate instructions.
Perhaps we can replace `-mfloat-abi` with this instead, if this works on Android?

atsushieno:
It did not work, resulted in the same error.

I'm indeed seeing -mhard-float to have worked like that https://android.googlesource.com/platform/ndk/+/34fe1db112d048a94d4b555d273ee037e7f5a804/tests/device/hard-float/jni/Android.mk , but this test is gone from ndk/master. Probably the situation has changed.

jpcima:
I think it's correct. The only worry is I don't have the toolchain right now to check this generates hard float.
Do you mind to quick check VFP instructions to be present?
eg. `android-[...]-objdump -d build/src/CMakeFiles/sfizz_static.dir/sfizz/SfzFilter.cpp.o | grep f32`

atsushieno:
clang++ indeed generates hard float instructions:

```
/sources/aap-lv2/aap-sfizz/.cxx$ /home/atsushi/Android/Sdk/ndk/21.2.6472646/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-objdump -d ./cmake/debug/armeabi-v7a/src/CMakeFiles/sfizz_static.dir/sfizz/SfzFilter.cpp.o | grep f32
./cmake/debug/armeabi-v7a/src/CMakeFiles/sfizz_static.dir/sfizz/SfzFilter.cpp.o:     file format elf32-littlearm
  50:	eef7 1ac0 	vcvt.f64.f32	d17, s0
  94:	eef7 0ac0 	vcvt.f64.f32	d16, s0
  e2:	eeb7 0be0 	vcvt.f32.f64	s0, d16
<snip>
```

(It is with `-mhard-float`)

jpcima:
Confirmed working on my side

PR #500, opened 2020-10-15
Add a mutex on the loading jobs to protect them in the case of freewheeling

paulfd:
Seemed to create alot of problems for ARM builds which hit the race condition all the time.


jpcima:
Perhaps you forget this one as well
https://github.com/sfztools/sfizz/blob/a5647ab33943302c4f99bd9cd61e5745510bd362/src/sfizz/FilePool.cpp#L480

paulfd:
I did, I moved it around at the last minute stupidly :)

jpcima (approved):

jpcima:
Use `std::mutex`? RT thread is not involved here

PR #508, opened 2020-10-19
Ref-count the global thread pool

jpcima:
Closes the thread pool when the synth is closed, instead of the static destructor.
Appears to fix the deadlock on Windows.


jpcima:
@kinwie, this is expected to take care of the closing issue.
Tried here on windows 7 32-bit.

kinwie:
@jpcima Yes, checked on my side. It is fixed, thanks!

PR #512, opened 2020-10-20
Fix an implementation error of FNV-1a

jpcima:
The algorithm says to process every next byte.
It's not what happens in case the platform defines `char` as signed.


jpcima:
This one is desired for osc dispatching, I merge it now.
A distinct function is available for hashing single bytes.

PR #521, opened 2020-10-23
Clearing concurrency issues

### LINKED ISSUE #519: Concurrency issue: crash on changing instrument
The problem is as follows:
- change instrument
- the DSP invokes `FilePool::triggerGarbageCollection` on the old files
- some of these files are still streaming, their memory block is freed and they write invalid memory
- crash with [report.txt.gz](https://github.com/sfztools/sfizz/files/5420012/report.txt.gz)

very easily reproduced with a timer that changes loaded instrument every tick
[stress.patch.gz](https://github.com/sfztools/sfizz/files/5420016/stress.patch.gz)

It's difficult to find a quick fix to this one, it's more than tempting to revert the loader to pre-0.5.0.

paulfd:
Closes #519 

What hurt was a mix of many things. My current analysis was that:
- The file pool was cleared but not the garbage to collect nor the last used files
- The background loaders could still be running while clearing the file pool
- The lock was released between clearing and loading a new file; no voices were running, but the garbage collection could be triggered
- There was no check in the garbage collection method if the "id" of the sample to check was indeed in the `preloadedFiles` data structure, thereby leading to UB.

The fix was to:
- wait for background loading to finish before clearing the `FilePool`
- clear all the `FilePool` structures (garbage, lastUsedFiles and preloadedFiles) under the callback lock in the `Synth::clear()` method.

I also added a couple sanity checks, and merged the FilePool mutex on garbage and used files into one.

I would ask for your pov @jpcima on the use of the callback lock in the loading functions of the synth. I wonder if it's a good idea to release the lock between `Synth::clear()` and the rest of the methods. We could remove the mutex hold in `clear()` and handle it only in the `loadSfz...()` ones: `clear()` is not in the public API and only called by the loading functions.

Still needs testing on various platforms to see if everything works as expected...

### DIFF
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -428,7 +428,7 @@ void sfz::FilePool::loadingJob(QueuedFileData data) noexcept
 
     data.data->status = FileData::Status::Done;
 
-    std::lock_guard<SpinMutex> guard { lastUsedMutex };
+    std::lock_guard<SpinMutex> guard { garbageAndLastUsedMutex };
     if (absl::c_find(lastUsedFiles, *id) == lastUsedFiles.end())
         lastUsedFiles.push_back(*id);
 }
@@ -514,10 +514,7 @@ void sfz::FilePool::dispatchingJob() noexcept
 void sfz::FilePool::garbageJob() noexcept
 {
     while (semGarbageBarrier.wait(), garbageFlag) {
-        std::lock_guard<SpinMutex> guard { garbageMutex };
-        for (auto& g: garbageToCollect)
-            g.reset();
-
+        std::lock_guard<SpinMutex> guard { garbageAndLastUsedMutex };
         garbageToCollect.clear();
     }
 }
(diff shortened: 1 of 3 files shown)


paulfd:
Seems to work OK on windows with VSTHost and Reaper x64. @kinwie if you want to double check.

jpcima:
It seems to work robustly here, let's merge it.

> We could remove the mutex hold in clear() and handle it only in the loadSfz...() ones:

Agreed.

src/sfizz/FilePool.cpp
```diff
@@ -587,17 +587,24 @@ void sfz::FilePool::setRamLoading(bool loadInRam) noexcept
...
     swapAndPopAll(lastUsedFiles, [&](const FileId& id) {
         if (garbageToCollect.size() == garbageToCollect.capacity())
            return false;
 
-        auto& data = preloadedFiles[id];
+        auto it = preloadedFiles.find(id);
+        if (it == preloadedFiles.end()) {
+            // Getting here means that the preloadedFiles got changed (probably cleared)
+            // while the lastUsedFiles were untouched.
+            ASSERTFALSE;
```
jpcima, line 603: We should not keep this assert, given that this condition is possible, is that correct?

paulfd: Well the `if` handles the error gracefully if this happens in Release: the file id is assumed invalid and remove from the list of recent files. Assuming the clearing is done properly it should not happen though. I'd keep the assertion as a way to catch an unexpected code path signaling a deeper error.

PR #525, opened 2020-10-24
Better installation rules for macOS

jpcima:


jpcima:
@alcomposer this merged PR adds effortless `make install` for macOS, as requested

PR #482, opened 2020-10-07
Audio file reader without sndfile

jpcima:
Add option `SFIZZ_USE_SNDFILE`, which permits to use dr libraries as a replacement when `OFF`.

PR #527, opened 2020-10-27
CI reorganization

redtide:
Based on a recent discussion, this PR moves the macOS build to Appveyor to access its built packages (artifacts) and removes the unneeded resulting packages already created in OBS, but keeping the builds for testing in Travis-CI. Stages was also removed.

The Appveyor configuration was renamed to `.appveyor.yml`, so this requires to change the related configuration on the UI on their website. To make it clean I splitted it using some additional scripts in the scripts/appveyor directory in a similar way done for Travis.

There are also some minimal changes in gitignore and in the packages naming convention.

The deployment result can be seen [here](https://github.com/redtide/sfizz/releases/tag/0.5.1-test-ci).


paulfd:
Nice thanks !

I'm in favor of deprecating some of the builds. The mingw ones are still gcc errors and compilations, and we have msvc builds with appveyor. Testing the library build and the static plugin build is nice but could probably be only amd, since there's an arm64 test build already, and the MOD build tests the armv7. What's the opinion?

jpcima:
Ok mostly, except:
- add `scripts/appveyor` to gitattributes, exclude from exporting
- do not enforce `CMAKE_PREFIX_PATH` in our build files? it's likely better passed as an option to cmake.
  (if using Brew then prefix is /usr/local, but if MacPorts then it's /opt/local, etc)

jpcima:
> I'm in favor of deprecating some of the builds. The mingw ones are still gcc errors and compilations, and we have msvc builds with appveyor. 

The mingw builds are useful for me, to have local windows builds, so I'll like to keep it afloat as possible.
Also note also that make-based builds accept gcc only.

> Testing the library build and the static plugin build is nice but could probably be only amd

I think to put the build process of plugins, libs, as well as test processes all together in one.
Also, we can take advantage of some configurations to run some generic plugin tests. (like the Tracktion pluginval, Apple auval)
Imo, the ARM tests can be valuable in some particular cases, maybe in the future: numeric tests, concurrency tests;
we have some experience that some concurrency problems tend to arrive on ARM only, for instance these pre-0.5.1 test runs in Travis that blocked often and only in the ARM64 context.

paulfd:
> > I'm in favor of deprecating some of the builds. The mingw ones are still gcc errors and compilations, and we have msvc builds with appveyor.
> 
> The mingw builds are useful for me, to have local windows builds, so I'll like to keep it afloat as possible.
> Also note also that make-based builds accept gcc only.

OK so let's keep them. What I meant for gcc is that all linux builds are gcc so we already have the quirks and errors of gcc :)

> > Testing the library build and the static plugin build is nice but could probably be only amd
> 
> I think to put the build process of plugins, libs, as well as test processes all together in one.
> Also, we can take advantage of some configurations to run some generic plugin tests. (like the Tracktion pluginval, Apple auval)
> Imo, the ARM tests can be valuable in some particular cases, maybe in the future: numeric tests, concurrency tests;
> we have some experience that some concurrency problems tend to arrive on ARM only, for instance these pre-0.5.1 test runs in Travis that blocked often and only in the ARM64 context.

Yes I was arguing to keep the ARM tests but drop the other ARM builds. Merging them if possible could help the build speed. We could also run tests on the appveyor builds maybe?

redtide:
> Ok mostly, except:
> 
>     * add `scripts/appveyor` to gitattributes, exclude from exporting
I forgot about the gitattributes for the source builds, my bad
>     * do not enforce `CMAKE_PREFIX_PATH` in our build files? it's likely better passed as an option to cmake.
>       (if using Brew then prefix is /usr/local, but if MacPorts then it's /opt/local, etc)
that was the only workaround I found to fix the problem, sure it's better to pass it as parameter during configuration

jpcima (approved):

paulfd:
I'll rebase the branches once I have the tests working in appveyor.

One thing is strange: https://ci.appveyor.com/project/SFZTools/sfizz/builds/36003468
In this case the win builds failed but the overall build appears as working? Is there an option somewhere to say that if any build fails in the matrix it fails overall? There seems to be a `fast_finish: true` option to fail on the first failure...

For the Travis build, would it be OK then if I merge the "library" builds with the "tests" builds? We're not deploying the libraries anymore and the targets that are built are mostly the same between these 2 jobs.

jpcima:
> In this case the win builds failed but the overall build appears as working?

Possibly an effect of the `allow_failures` setting in appveyor?
Why exactly is it set in the first place?

> For the Travis build, would it be OK then if I merge the "library" builds with the "tests" builds?

Yes, and the plugins too.

paulfd:
For the plugins it tests the "static" path and thus needs a reconfigure, maybe having a separate job is okay there?

paulfd (approved):

jpcima:
wasn't sndfile fine as private dependency before?

paulfd:
It's in `FileMetadata.h`, which get included up to `Synth.h`, featuring prominently in the test suite. I think on linux it's mostly silent because the file is in the system includes, but on windows it has to be public.

Probably the encapsulated version with the dr_libs of #482 will remove the need for this.

jpcima:
Understandable, but the projects which are including sfizz as a subdirectory of theirs shouldn't see sndfile.
Perhaps, a target `sfizz-internal` is a better idea.

jpcima:
@paulfd  I did this rebase wrong on top of dr-libs and lost your work, do you mind to repost the commits from your local copy?

paulfd:
sure

PR #528, opened 2020-10-28
Let the flex eg free-run when set as the ampeg

paulfd:
This matches the behavior of the "normal" ampeg.


paulfd:
After tests, this is not what sforzando does. The `loop_mode=one_shot` is ignored in this case. I'd however say that this is a reasonable feature; if you don't want the envelope to free-run, don't put one_shot there.

jpcima (approved):

PR #532, opened 2020-10-30
VST state file search

jpcima:
This adds a mechanism lo let VST find files on a different environment, after the session has been shared/moved/etc.
It will establish a list of directories where SFZ files are expected to be stored.

When a file is missing after state loading, the plugin is going to run a background scan, synchronously but shared between multiple instances.
This will collect the paths into a map [filename]->[list of paths] with lower-cased keys.

On scan completion, the file path is looked up into the index, yielding a list of candidates.
Then, we study the absolute path by comparing directory components in reverse order with the candidate, yielding a score according to the number of matched components.

This process applies to sfz files as well as tuning files.

### DIFF
--- a/vst/CMakeLists.txt
+++ b/vst/CMakeLists.txt
@@ -18,14 +18,20 @@ set(VSTPLUGIN_SOURCES
     SfizzVstEditor.cpp
     SfizzVstState.cpp
     VstPluginFactory.cpp
-    X11RunLoop.cpp)
+    X11RunLoop.cpp
+    NativeHelpers.cpp)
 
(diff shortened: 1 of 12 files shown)


jpcima:
There is other potential for optimization from here.
- [x] speed: ~~from experience `fts.h` is much faster for recursive file scanning on non-Windows, but the comparison was to boost::filesystem, not ghc~~
- [x] memory: use a hierarchical file-trie structure such that directory components of absolute paths are not duplicated when stored

paulfd (approved):

PR #533, opened 2020-10-30
Refactor of Synth and Voice into a pimpl pattern

paulfd:
- Introduced a "VoiceList" `std::vector` proxy to hold the voices. This new structure belongs in the synth and hold the `getVoiceById` method. Note that since Voices are "pimpled", `VoiceList` holds `Voice` objects instead of `std::unique_ptr<Voice>` ones. This require custom moves in the Voice, in particular to handle the intrusive `SisterVoiceRing`.
- Updated the code style to have `_` suffixes to private variables.
- The `ccInitialValues` structure was basically unused, so for now the `MidiState` is the holder of the ground truth of CC values. Upon reset, CC 7, 10 and 11 are forced to their default values. Should we also reset all the others?

I'm testing on platforms now, but @jpcima maybe you can check that in principle you agree with the technical direction and choices in there. I might continue to refactor parts of it as I go.


jpcima:
It's agreed, just some remarks I can think of
- it might be interesting to move some polyphony logic out of `Synth` if possible and inside the `VoiceList`. (potentially upgrade it into a `VoiceManager`)
- the CC initial values are not unused, I will have interest in these for UI interaction, because the knobs of the CC tab need be associated with a default value.
- I don't have a strong opinion regarding the CC reset. Let's do like aria?

paulfd:
> It's agreed, just some remarks I can think of
> 
>     * it might be interesting to move some polyphony logic out of `Synth` if possible and inside the `VoiceList`. (potentially upgrade it into a `VoiceManager`)

Ok I'll think about it more !

>     * the CC initial values are not unused, I will have interest in these for UI interaction, because the knobs of the CC tab need be associated with a default value.

OK so it's more to know where to reset the value if asked by the UI? In this case, my bad, I'll revert it.

Any idea about the clang tidy check failing? I'm a bit at a loss, there's no argument to this constructor..
```
/home/travis/build/sfztools/sfizz/src/sfizz/Synth.cpp:381:17: error: Function taking 2 arguments is called with fewer (1) [clang-analyzer-core.CallAndMessage,-warnings-as-errors]

    impl_.reset(new Impl);
```

paulfd:
Maybe just review that the structure is okay. I "mostly" changed the way the polyphony checks and voice stealing are handled, but these had a hefty test coverage.

jpcima:
The structure looks very reasonable.
There is some code-style remarks, which I'll rather note in the proper issue.
Go ahead and merge if you wish, this is sure to save me some future conflicts with another piece of work.

jpcima:
Thanks. This had a problem of mismatches `class/struct Region`, I fixed this with a commit in develop.

PR #540, opened 2020-11-02
Add OSC and plugin-side communication

jpcima:
This implements the OSC message protocol.
- serialization routines in `sfizz_message.h` + tests
- sfizz client API for sending to synth and receiving
- a skeleton of a message matcher by hashing
- a VST implementation over FIFO buffer
- a LV2 implementation over custom binary atoms (not osc.lv2 extension, that's for another time)
- communicating the messages with UI in both directions

not implemented:
- actual messages that will interest us (cf. `SynthMessaging.cpp`)

### DIFF
--- a/lv2/sfizz.c
+++ b/lv2/sfizz.c
@@ -148,6 +148,7 @@ typedef struct
     LV2_URID sfizz_log_status_uri;
     LV2_URID sfizz_check_modification_uri;
     LV2_URID sfizz_active_voices_uri;
+    LV2_URID sfizz_osc_blob_uri;
     LV2_URID time_position_uri;
     LV2_URID time_bar_uri;
     LV2_URID time_bar_beat_uri;
@@ -158,6 +159,7 @@ typedef struct
 
     // Sfizz related data
     sfizz_synth_t *synth;
+    sfizz_client_t *client;
     bool expect_nominal_block_length;
     char sfz_file_path[MAX_PATH_SIZE];
     char scala_file_path[MAX_PATH_SIZE];
@@ -181,6 +183,9 @@ typedef struct
 
     // Paths
     char bundle_path[MAX_BUNDLE_PATH_SIZE];
+
+    // OSC
+    uint8_t osc_temp[OSC_TEMP_SIZE];
 } sfizz_plugin_t;
 
 enum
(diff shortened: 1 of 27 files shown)


lv2/sfizz.c
```diff
@@ -423,6 +429,28 @@ sfizz_lv2_update_timeinfo(sfizz_plugin_t *self, int delay, int updates)
...
+    if (osc_size > OSC_TEMP_SIZE)
+        return;
+
+    bool write_ok =
+        lv2_atom_forge_frame_time(&self->forge, 0) &&
+        lv2_atom_forge_atom(&self->forge, osc_size, self->sfizz_osc_blob_uri) &&
+        lv2_atom_forge_raw(&self->forge, osc_temp, osc_size);
+    lv2_atom_forge_pad(&self->forge, osc_size);
+
+    (void)write_ok;
```
paulfd, line 451: Why not testing this?

paulfd (approved):

paulfd (approved):
Ok it looks good ! I'll give it a try soon to replace/add introspection (e.g. remove the getRegionView type of comments for the tests, it'd be nice to be able to test directly the public API for this).

lv2/sfizz.c
jpcima: It's false when the port buffer is full. There is no recovery to take if this happens, at best we could log this as a warning.

jpcima:
Ok, the plan was to do this by request and reply in a form like this.
- getting: `/region/1/sample` (no args) -> `/region/1/sample` (arg: "s")
- setting: `/region/1/sample` (arg: "s")

Also there is possibility of handling a wildcard message for querying all at once
- `/region/*/sample` -> `/region/1/sample`, `/region/[...]/sample`, `/region/N/sample` (arg: "s")

There exists a "OSC best practices" pdf somewhere, which has some guidelines to follow in this regard.

src/sfizz/SynthMessaging.cpp
```diff
@@ -0,0 +1,81 @@
...
+    while (unsigned char c = *path++) {
+        if (!absl::ascii_isdigit(c))
+            h = hashByte(c, h);
+        else {
+            h = hashByte('&', h);
+            while (absl::ascii_isdigit(*path))
+                ++path;
+        }
+    }
+    h = hashByte(',');
```
paulfd: Should be `h = hashByte(',', h);`

PR #547, opened 2020-11-12
Add API to process CC events received by automation

jpcima:
This API allows to receive CC values without having them interpreted as MIDI events.
For instance, this API would permit to turn a UI knob attached to the CC 121, without triggering an avalanche of reset events.


paulfd:
I wonder if then we'd also "ignore" the volume and pan midi defaults then?

jpcima:
> I wonder if then we'd also "ignore" the volume and pan midi defaults then?

No, think of it really in context of automation.
It's not a cause of problems to automate the CC7/10 by either MIDI or knob.

paulfd (approved):

PR #552, opened 2020-11-14
Set up the code-signing key pair for macOS

jpcima:
Sets up the code-signing identity for macOS.
Note that the secrets are not available when on PR, for security reasons.
We enable an exception on Appveyor to permit it, when the PR comes from the same repo.


jpcima:
Attaching CA and Signer certificate to load in user's keychain
[public.zip](https://github.com/sfztools/sfizz/files/5541037/public.zip)

(Signer probably not required, but just in case)

PR #559, opened 2020-11-22
Opcode parsing and defaults

paulfd:
This reworks the opcode parsing, with 3 main goals:
- Have default constants in a `.cpp` file
- Avoid the overloaded/templated `setXXXfromOpcode` methods which have to be duplicated for different use cases.
- Have a finer "clamping" behavior for out-of-bounds values (i.e. ignore the value or clamp it)

The opcodes defaults are now `OpcodeSpec`s, with the following signature:
```c++
template<class T>
struct OpcodeSpec
{
    T value;
    Range<T> bounds;
    int flags; // can be kIgnoreOOB, kCanBeNote, kEnforceLowerBound, kEnforceUpperBound
};
```
These signatures are in `Default.cpp` for all relevant opcodes. Some opcodes share a spec, most notably the `key` one that spans `hikey`, `lokey`, `pitch_keycenter`, `amp_keycenter`, ...

Some defaults still have to live outside of this paradigm, for example default ranges for crossfades and enums.
I also kept the `readBooleanFromOpcode` for now.
I also moved the effect defaults into `Defaults.h`.

The "new" way to parse an opcode is now to use the member function `read(OpcodeSpec<T> spec)`, either as
```c++
if (auto value = opcode.read(Default::loopRange))
   loopRange.setEnd(*value);
```
or with the single-line shortcut for simple plain values that could be "replaced" by the new opcode.
```c++
delay = opcode.read(Default::delay).value_or(delay);
```
The member method properly dispatches between `int`/`float` specs, and handles the opcode as specified by the flags.

This builds on top of #555. I still need to activate some of the tests, and I will add tests and OSC interfaces for the effects also.


jpcima:
This looks fantastic, but now all builds are failing at generating the editor UI.
An idea why that is? I will check on my side.

paulfd:
No idea I'll check later 😊

Nov 22, 2020 20:21:51 JP Cimalando <notifications@github.com>:

jpcima:
The reason of this problem is that the final commit would remove the generated file.
As consequence, the target directory would be empty, git does not record this directory, generating fails because of the missing directory.

src/sfizz/Region.h
```diff
@@ -43,17 +43,7 @@ class RegionSet;
...
-    {
-        ccSwitched.set();
-
-        gainToEffect.reserve(5); // sufficient room for main and fx1-4
-        gainToEffect.push_back(1.0); // contribute 100% into the main bus
-
-        // Default amplitude release
-        amplitudeEG.release = Default::ampegRelease;
-    }
+    Region(int regionNumber, const MidiState& midiState, absl::string_view defaultPath = "");
```
jpcima, line 46: If a control header sets `default_path` after this region is built, this region's `defaultPath` turns into a dangling pointer.

src/sfizz/Opcode.cpp
```diff
@@ -117,6 +117,116 @@ OpcodeCategory Opcode::identifyCategory(absl::string_view name)
...
+    if (returnedValue < static_cast<int64_t>(spec.bounds.getStart())) {
+        if (spec.flags & kEnforceLowerBound)
+            return spec.bounds.getStart();
+
+        if (spec.flags & kIgnoreOOB)
+            return {};
+    }
+
+    T castValue = static_cast<T>(returnedValue);
+    if ((castValue != returnedValue) & kIgnoreOOB)
```
jpcima: This conditional is looking suspicious.
```suggestion
    if (castValue != returnedValue && (spec.flags & kIgnoreOOB))
```

I'm not sure what it's trying to accomplish though.
do we have to repair values in case of an overflow?

jpcima:
~~The flag `kIgnoreOOB` feels like an anomaly.
It's better imho `kEnforceBounds = kEnforceLowerBound|kEnforceUpperBound`, and not enforcing bounds being default.
This makes also the code more straightforward, the conditionals `kIgnoreOOB` can go altogether.~~

jpcima:
Nevermind I understood it wrong. I take it that `kIgnoreOOB` must be for enumerated values.
It raises some questions though, why is it set some parameters which are not enumerated?
(eg. `compThreshold`, `compRatio`)

src/sfizz/Opcode.cpp
paulfd: Yes it's wrong. I'd say in case of an overflow we should probably ignore the input.

paulfd:
No actually it's more the "ignoring vs clamping" question when you get a value out of bounds, so not for enumerated. For example what to do if you get `offset=40 offset=-1`? Do you keep offset at 40 or "clamp" the -1 to zero?

src/sfizz/Region.h
paulfd: This is wrong sure, I thought this was an std::string at some point (and it probably should be).

jpcima:
> Do you keep offset at 40 or "clamp" the -1 to zero?

Imho a maximum of the opcode reading process should be encoded into the spec.
That includes "-1" special values, textual enumerations, normalizations to %, and so on.

paulfd:
I'll try to put in more.

Regarding the out-of-bounds situation, it seems sforzando has varying behaviors. For example, negative values for `ampeg_attack` will cause the value to be reset to the default one. Negative values for `pitch_keycenter` seem to screw up everything however. I think I'll remove the `kIgnoreOOB` flag, and in case the value is out of an enforced bound, then the read method returns `nullopt`. In the current parsing most of the time this will leave the value unchanged rather than reset to default, it seems better like this.

paulfd:
Hi,

A bit tired of this PR tbh, but here goes. Now the `OpcodeSpec` handles everything in the background. The prototype is as follows:
```c++
template<class T>
struct OpcodeSpec
{
    T defaultInputValue;
    Range<T> bounds;
    int flags;
    /**
     * @brief Normalizes an input as needed for the spec
     *
     * @param input
     * @return T
     */
    T normalizeInput(T input) const { .... }

    operator T() const { return normalizeInput(defaultInputValue); }
};
```

Flags can be
```c++
enum OpcodeFlags : int {
    kCanBeNote = 1,
    kEnforceLowerBound = 1 << 1,
    kEnforceUpperBound = 1 << 2,
    kNormalizePercent = 1 << 3,
    kNormalizeMidi = 1 << 4,
    kNormalizeBend = 1 << 5,
    kWrapPhase = 1 << 6,
    kDb2Mag = 1 << 7,
};
```

In usage for parsing, opcodes are initialized as
```
int64_t offsetRandom { Default::offsetRandom }; // offset_random
```
and parsed as
```
crossfadeVelCurve = opcode.read(Default::crossfadeCurve);
```
or if you need an unset optional as default
```
downKeyswitch = opcode.readOptional(Default::key);
```
The `read` method basically calls `readOptional`, and `readOptional` sends back `nullopt` if the value is e.g. out of bounds. The `read` method then sends back the default value if this happens. This matches ARIA's behavior. All the switching and normalization appears behind the scene. I also tried to tidy up the effects and put their defaults with the others as much as possible.

Some cleanups, commenting and testing to do still.

src/sfizz/Region.h
jpcima: There's 2 possible ways to fix this.
- should the region reference the path known to the synth, or
- copy the value at the time of instantiation?

ie. does a `default_path` opcode take effect on all regions, or just on the regions which appears after it?
```
<region> // A
<control> default_path=[..]
<region> // B
```

paulfd: I corrected it by copying it in the current state, to be on the safe side. To answer your question, it's only the regions that come after. If we wished to be efficient we could have a vector of "defaultPaths" in the current sfz file, with this string view referencing the proper one.

jpcima (approved):

PR #561, opened 2020-11-23
Interpolation using simde

jpcima:
It's a beginning of neon support implemented with simde.
This implements the interpolators.

### DIFF
--- a/src/sfizz/Interpolators.hpp
+++ b/src/sfizz/Interpolators.hpp
@@ -8,6 +8,10 @@
 #include "WindowedSinc.h"
 #include "MathHelpers.h"
 #include "SIMDConfig.h"
+#include <simde/simde-features.h>
+#if SIMDE_NATURAL_VECTOR_SIZE_GE(128)
+#include <simde/x86/sse.h>
+#endif
 
 namespace sfz {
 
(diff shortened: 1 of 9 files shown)


paulfd (approved):

jpcima:
Are you able to test this and benchmark? last one is for a NEON-only optimization.

src/sfizz/Interpolators.hpp
```diff
@@ -48,25 +52,17 @@ class Interpolator<kInterpolatorLinear, R>
 //------------------------------------------------------------------------------
 // Hermite 3rd order, SSE specialization
 
-#if SFIZZ_HAVE_SSE
+#if SIMDE_NATURAL_VECTOR_SIZE_GE(128)
```
paulfd, line 57: This should be true for all archs we support, but just in case maybe we can have a generic fallback instantiation for scalar floats (just adding an `#else` with the template instantiation)

paulfd (approved):

src/sfizz/Interpolators.hpp
jpcima: Wasm lacks this support.
It has already a scalar fallback anyway. I wrote it as a reference.

paulfd:
Yes that's true I'm a bit rusty, it'll get instantiated already.

Dec 26, 2020 15:43:04 JP Cimalando <notifications@github.com>:

PR #566, opened 2020-11-25
Implementation of "hint_min_samplerate" and semi-piecewise interpolation

PythonBlue:
This pull request implements optional voice oversampling via the ARIA-specific opcode "hint_min_samplerate".

Also implemented a custom interpolation mode that, when combined with said oversampling, can sound convincing to alias-free piecewise interpolation. To activate this interpolation, one needs to set the "sample_quality" to 0.

Third significant change was allowing the resonance to go as low as "-3" as I felt 0 sounded a little too sudden of a cutoff to me. This much can be ommitted from the main repo if necessary.


jpcima:
Some quick remarks
- `sample_quality` 0 is not implemented. It's just a clause to add in Voice.cpp `fillInterpolatedWithQuality` with the quality `kInterpolatorLinear`. This would be good as a PR of its own. It will be a SFZ extension.
- I've got a feeling that the IIR technique would be inefficient at high rates. Perhaps at low rates we can depend on the half-band technique (as exists now);
  upper rates could use Polyphase FIR technique? then these frames that will be discarded by decimating can be skipped altogether, and also FIR is not autodepent of its output, which opens the possibilities of SIMD implementation

jpcima:
Perhaps, this has to do with this method you are implementing?
https://www.gearslutz.com/board/electronic-music-instruments-and-electronic-music-production/750045-synclavier-emulation.html
https://www.gearslutz.com/board/showpost.php?p=8127446

PythonBlue:
It does, yes, and it is the same premise. Basically, my method of interpolation is meant to get close to the desired sound of the imaging frequencies, while controlling the obvious aliasing that happens when said frequencies exceed the Nyquist.

jpcima:
Some more remarks
- the term `pow()` is repeated identically twice in the interpolation routine, this should be preferably avoided
- the interpolation is preferably rewritten as `y1+(y2-y1)*mu` instead of `y1*(1-mu)+y2*mu`, this should save an operation
- the math function `pow()` is preferably avoided in the resample loop. Take a fast-path for the case `mod=1`.
A 2d-table may be used, polynomials or other approximations. It's something to benchmark.
(eg. https://martin.ankerl.com/2012/01/25/optimized-approximative-pow-in-c-and-cpp/)

In the end I'm favorable to have such feature in the engine, but possibly should be provided and parameterized as some extension opcode.

(As I understand the method overall, it's a system that degrades the quality of the antialiasing filter in relation with the resample factor. I have put the "lofi" equation in my recent response analysis tool, attached here below)
[response.py.gz](https://github.com/sfztools/sfizz/files/5978575/response.py.gz)

PythonBlue:
> Some more remarks
> 
>     * the term `pow()` is repeated identically twice in the interpolation routine, this should be preferably avoided
> 
>     * the interpolation is preferably rewritten as `y1+(y2-y1)*mu` instead of `y1*(1-mu)+y2*mu`, this should save an operation
> 
>     * the math function `pow()` is preferably avoided in the resample loop. Take a fast-path for the case `mod=1`.
>       A 2d-table may be used, polynomials or other approximations. It's something to benchmark.
>       (eg. https://martin.ankerl.com/2012/01/25/optimized-approximative-pow-in-c-and-cpp/)
> 
> 
> In the end I'm favorable to have such feature in the engine, but possibly should be provided and parameterized as some extension opcode.
> 
> (As I understand the method overall, it's a system that degrades the quality of the antialiasing filter in relation with the resample factor. I have put the "lofi" equation in my recent response analysis tool, attached here below)
> [response.py.gz](https://github.com/sfztools/sfizz/files/5978575/response.py.gz)

Bullet #1: Has the pow() function been used twice? I'll need to double check to confirm, though be assured that changing to that extent was unintentional.

Bullet #2: Pretty sure the part of the interpolation code that resembled linear was unchanged from the pure linear equivalent when I first edited the code, but we can rewrite it anyway.

Bullet #3: For use of pow() at all, I'll admit I'm inexperienced with polynomials and some other complex forms of math, so this was the best I could do with the knowledge I do have. My custom interpolation, to help you understand the theory, was to make it sound like an old-school hardware sampler (hence the name in the code) while technically avoiding the aliasing by keeping the added frequencies below the Nyquist, which is also why the power changes depending on the playback pitch. It's a similar concept to what was done with chipsounds in ARIA, but it isn't dependent on a certain bitrate. If you can find another method that's simpler, or at least is more CPU efficient, I'm willing to take a look into it.

Thank you for the critiques.

PythonBlue:
Just wanted to give an update for this pull request: created an additional branch called "eg_shapes", which, needless to say, implements support for shapes for the non-flex EG's, though as of today, it also redoes the cc-based modulation of filter cutoff to support curves. I would make a separate pull request for that much, but I fear that may be a bit redundant.

paulfd:
Let's revisit this resampling for the next point version. I'd like that we make it have zero cost for the "normal" path if at all possible.

Filter cutoffs are built on the modulation matrix, so they should support curves. Do you have a non-working example?

PythonBlue:
> Let's revisit this resampling for the next point version. I'd like that we make it have zero cost for the "normal" path if at all possible.
> 
> Filter cutoffs are built on the modulation matrix, so they should support curves. Do you have a non-working example?

Fair enough about resampling.

As for the filter cutoff issue, I can easily name one case that doesn't sound correct in the current main fork: any SFZ file that makes use of the "cutoff&_curvecc&" opcode. That addition was made on my end out of some admittedly selfish impatience. For example's sake, one is attached, even if the samples themselves aren't included for copyright reasons:
[https://www.dropbox.com/s/5ac757n9mma5hnx/112-Movie%20Layer%20.sfz?dl=0](
https://www.dropbox.com/s/5ac757n9mma5hnx/112-Movie%20Layer%20.sfz?dl=0)

paulfd:
Do you mind if I extract the part about the cutoff curves in separate PRs?

PythonBlue:
> Do you mind if I extract the part about the cutoff curves in separate PRs?

Not at all. I should've updated this pull request anyway to say that "hint_min_samplerate" was eventually dropped in my fork due to the complexity of updating with the newer commits. Feel free to borrow and refine my code for the attack/decay/release_shape opcodes as well if you wish.

PR #588, opened 2020-12-20
Possible corrections to CheckIPO.cmake script

redtide:
- add a warning when building for armv7 instead disable silently
- better condition expression syntax

I wonder if it could be useful also to rename `ENABLE_LTO` with a `SFIZZ_` prefix as all other options.

### DIFF
--- a/cmake/CheckIPO.cmake
+++ b/cmake/CheckIPO.cmake
@@ -1,36 +1,34 @@
-# Added in CMake 3.9
-
 if(CMAKE_VERSION VERSION_LESS 3.9)
     message(WARNING "\nIPO checks are only available on CMake 3.9 and later.")
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL armv7l)
+    message(WARNING "\nLTO not enabled for armv7l.")
+else()
+    set(LTO_CHECK_OK ON)
+endif()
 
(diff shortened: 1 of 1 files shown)


jpcima:
Regarding the armv7 special case:

Perhaps it's best to define some platform-specific defaults for our options, and LTO is one of the candidates for that.
eg. it's ARMv7, LTO is default to OFF, otherwise ON
Then it's not needed to force it off, even in case the user explicitly requests it.

The same could apply for instance, to building JACK: set it default to ON on Linux, and OFF on other platform.

paulfd:
I like jpcima's idea, although I do agree that the silent ignore is not good.

IIRC `ENABLE_LTO` is not really a sfizz specific option, it's a cmake one.

redtide:
`ENABLE_LTO` is a custom option defined in the main `CMakeLists.txt` file, used as a flag in `cmake/CheckIPO.cmake`.
IIRC at the time we wrote that cmake module, the idea was to keep it generic for any project, but since it has also a `SFIZZ_ENABLE_LTO_IF_NEEDED` function defined in it we could rename that flag too (if not the function removing the `SFIZZ` bit).

paulfd:
Ok I thought not. In this case yes a size prefix would be better !

Dec 26, 2020 15:16:34 redtide <notifications@github.com>:

PR #599, opened 2021-01-21
Make tuning frequency editable by wheel

jpcima:
This allows the `SValueMenu` to be edited by wheel.
Tuning frequency can be adjusted by increments of `0.1`.


ghost:
Brilliant!

ghost:
Ah, spoke too soon. I just tried and the value does not update in real time. Only on clicking on the box does the value show as updated.

jpcima:
Ok I probably forgot to `invalid()` the component after value change.
Fix in 2a82e50

ghost:
I can confirm that all is working as expected now. Thanks!

PR #609, opened 2021-01-31
Windowed sinc interpolation

jpcima:
This adds the sample quality 3 to 10.
See also: https://github.com/sfztools/sfizz/wiki/Resampling

There may be room for improvement, by adjusting Kaiser beta and table size.

Adds the Bessel I0 math function, required when C++ < 17.

### DIFF
--- a/benchmarks/BM_interpolators.cpp
+++ b/benchmarks/BM_interpolators.cpp
@@ -13,6 +13,11 @@
 
 class Interpolators : public benchmark::Fixture {
 public:
+    Interpolators()
+    {
+        sfz::initializeInterpolators();
+    }
+
     void SetUp(const ::benchmark::State& state)
     {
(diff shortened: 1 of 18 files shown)


jpcima:
Notifying @kinwie for `sample_quality` 3-10, implemented in this PR

kinwie:
@jpcima Wow, this is great implementation! I've checked the CPU jumps for each quality and it seems very optimized :+1: 
I even saw that sample_quality=3 (Sinc-8) seems use a bit less CPU than 2 (Hermite). And tested the sound of sample_quality=10, it sounds so fat and CPU also efficient. Superb!!

jpcima:
Great thanks, I merge this once.
On my machine Sinc 8-point is just a little slower than Hermite is, but this can depend on machine.

PR #617, opened 2021-02-01
Thread safety

jpcima:
It's a revamp of thread-safety in the entire library and the clients.
This PR removes all lock guards from the API, and passes the responsibility to the caller.
This breaks the library interface, because any thread-safety guarantees are gone. (bump the so number?)

The rationale is as follows
- as of currently, some library API are not mutex-protected but should have been; it's a source of threading problems, for instance in our LV2, which would let the sfizz library do all locking.
- at library level, a mutex would be taken for every message sent, which is much over-locking for nothing, and has possibly a performance penalty to it. we don't want to lock 100 times to send 100 messages
- also at library level, care should have been taken such that a function holding a lock must not call another which would take it. the only clean solution seems the *recursive mutex*, which doesn't exist in spin-lock variant, and possibly also has some added cost

As conclusion, in-library locking is a ton of complication, and it's broken, so I don't find a reason to insist keeping it.

This PR updates VST, LV2, JACK for client-side locking.
SpinMutex is organized as a distinct library such that the clients can access it.

At some time later, we must specify in documentation the threading restriction; the library is a 2-actor "controller thread/audio thread", it must tell which thread can call which function, and which of controller-side functions are safe to call while the audio runs.


jpcima:
Most likely fixes recently reported crashes

PR #618, opened 2021-02-02
Respect region end opcode

### LINKED ISSUE #616: SFZv1 end= opcode ignored
`<region> sample=test.wav offset=1000 end=2000`
Expectation is that sample playback stops at sample index 2000. From my tests, it looks like `end=` is ignored and sample is still played until the end of file.

dwuertz:
Limit sample playback according to "end" opcode in non-looping modes. Currently, it is ignored.
Fixes #616.
Unsure if the opcode needs also special hangling when looping?


jpcima:
I think it to be correct, I will review in more depth later.

PR #628, opened 2021-02-04
Refactor VST with UpdateHandler

jpcima:
This refactor rewrites VST such that Controller<->Editor is implemented with different strategy.
Needs more testing. When merged, it facilitates implementation of future automation.

Previously, it wrote a mutex-protected zone waiting UI to pick up the changes.
Instead, it uses the `UpdateHandler` provided in SDK.

By this model, the controller has a set of `update` objects, that editor can subscribe to.
Parameters are such objects, but we create also special objects for transmitting non-parameter data.
(the OSC messages, the SFZ file path, etc)

None of this is very evident from VST docs, but it works like this: an update can be performed in 2 ways.
- asynchronous, by `deferUpdates()`, which runs on the UI thread
- synchronous, by `change()`, which runs on the calling thread

A thing, it's I don't think we can push many OSC message by this system, as in some kind of queue, so I process them in a one-by-one fashion in synchronous mode. It's likely not efficient but it doesn't seem there is a choice. It is improvable by packing multiple OSC messages into the update object.


jpcima:
I ran the folllowing checklist. It also passed `pluginval` and `vstvalidator`.
- [x] editing host parameter affects UI control
- [x] editing UI control affecs host parameter
- [x] choosing new file affects UI controls
- [x] choosing new file affects UI paths
- [x] choosing new file affects UI playing state
- [x] save+load reloads files
- [x] save+load updates host parameters
- [x] save+load reloads UI controls
- [x] save+load reloads UI paths
- [x] save+load updates UI playing state

PR #642, opened 2021-02-17
Deprecate the internal oversampling factor

### LINKED ISSUE #638: audible artefacts and loud pops when oversampling in VCV Rack
Just gave the project *(the SFZ player, vst3) a spin in VCV Rack on windows, using the Host XL module. It may have been this machine, so I'll report more after further testing. Used some of the instruments in Edge-blown Aerophones from the vcsl SFZ library. Ocarina, and Pipe Organ, and found the issue when using oversampling. If this is a vcv rack issue it's fine, but if it's an issue with the project here i'd be happy to keep testing the issue. I have several other hosts as well, but like to use vcv rack for impromptu stuff.

paulfd:
Closes #638 

- Removed it completely from the library
- Kept the API for future work.


paulfd:
Ho I didn't see #640. Hm, I guess we can keep the Oversampler class even if we don't do an oversampling factor. At least you don't need to be constrained to the way the FilePool used it.

jpcima:
The work of #640 is intented for the lo-fi feature which has been requested.
If you recall, the request was a combination of nearest interpolation with a very high oversampling, around 100x.
#566 does it with `lp6p`, but the work aims to replace it with HIIR with ability of higher factors.

jpcima (approved):

PR #654, opened 2021-02-23
Add support for count and loop_count

paulfd:
Note that the `count` opcode apparently counts the number of restarts. It also sets `loop_mode` to `one_shot`.

### DIFF
--- a/src/sfizz/SynthMessaging.cpp
+++ b/src/sfizz/SynthMessaging.cpp
@@ -188,6 +188,8 @@ void sfz::Synth::dispatchMessage(Client& client, int delay, const char* path, co
             GET_REGION_OR_BREAK(indices[0])
             if (region.sampleCount)
                 client.receive<'h'>(delay, path, *region.sampleCount);
+            else
+                client.receive<'N'>(delay, path, {});
         } break;
 
         MATCH("/region&/loop_range", "") {
@@ -224,6 +226,14 @@ void sfz::Synth::dispatchMessage(Client& client, int delay, const char* path, co
         MATCH("/region&/loop_crossfade", "") {
             GET_REGION_OR_BREAK(indices[0])
             client.receive<'f'>(delay, path, region.loopCrossfade);
+        } break;
+
+        MATCH("/region&/loop_count", "") {
+            GET_REGION_OR_BREAK(indices[0])
+            if (region.loopCount)
+                client.receive<'h'>(delay, path, *region.loopCount);
+            else
+                client.receive<'N'>(delay, path, {});
         } break;
 
         MATCH("/region&/group", "") {
--- a/src/sfizz/SynthMessaging.cpp
+++ b/src/sfizz/SynthMessaging.cpp
@@ -186,7 +186,8 @@ void sfz::Synth::dispatchMessage(Client& client, int delay, const char* path, co
 
         MATCH("/region&/count", "") {
             GET_REGION_OR_BREAK(indices[0])
-            client.receive<'h'>(delay, path, region.sampleCount);
+            if (region.sampleCount)
+                client.receive<'h'>(delay, path, *region.sampleCount);
         } break;
 
         MATCH("/region&/loop_range", "") {
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -987,7 +987,6 @@ void Voice::Impl::fillWithData(AudioSpan<float> buffer) noexcept
     };
 
     if (loopContinuous) {
-
         for (unsigned i = 0; i < numSamples; ++i) {
             int wrappedIndex = (*indices)[i] - loop.size * blockRestarts;
             if (wrappedIndex > loop.end) {
@@ -1000,9 +999,8 @@ void Voice::Impl::fillWithData(AudioSpan<float> buffer) noexcept
             addPartitionIfNecessary(i, wrappedIndex, wrapped);
 
             // Release if we reached the loop count
-            if (wrapped && region_->loopCount && loop_.restarts >= *region_->loopCount && !released()) {
+            if (wrapped && region_->loopCount && loop_.restarts >= *region_->loopCount && !released())
                 release(i);
-            }
         }
     } else if (loopSustain) {
         unsigned i = 0;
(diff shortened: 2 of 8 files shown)


kinwie:
I think the description on the original sfz v1 specs is a little ambiguous.
But the implementation of `count` in sfz.dll, sforzando and Cakewalk DropZone almost the same, except for the 0 and 1 behavior.
The region plays as many times as the count number

paulfd:
@kinwie should be OK now !

kinwie:
@paulfd Tested the new build for this. The `count` is fine now.

However, I encounter an issue for `loop_count`
The last portion of the sample after the loop is not played even I keep holding the key
Just in this simple example :
```
<region>
sample=snare_looped.wav
key=60
loop_count=3
```
After the loop, the snare will be cutoff. In Rapture and Sforzando, after the loop, it will play until the end of the sample.
Again, I know there is no detailed description for this opcode :))

paulfd:
Okay I didn't know that, so I guess it's not over.

Loop sustain has a similar behavior right?

Feb 27, 2021 12:37:15 kinwie <notifications@github.com>:

kinwie:
Yes, in loop_sustain, the last portion of the sample after the loop, will be played after releasing the key and it controlled by ampeg_release

src/sfizz/SynthMessaging.cpp
```diff
@@ -165,7 +165,8 @@ void sfz::Synth::dispatchMessage(Client& client, int delay, const char* path, co
 
         MATCH("/region&/count", "") {
             GET_REGION_OR_BREAK(indices[0])
-            client.receive<'h'>(delay, path, region.sampleCount);
+            if (region.sampleCount)
+                client.receive<'h'>(delay, path, *region.sampleCount);
```
jpcima, line 190: can we still answer the request, without argument?

src/sfizz/Voice.cpp
```diff
@@ -924,12 +929,22 @@ void Voice::Impl::fillWithData(AudioSpan<float> buffer) noexcept
...
-
-            // wrap indices post loop-entry around the loop segment
-            int wrappedIndex = (index <= loop.end) ? index :
-                (loop.start + (index - loop.start) % loop.size);
+            int index = (*indices)[i] - loop.size * blockRestarts;
+            int wrappedIndex = index;
+            // wrap indices post loop-entry around the loop segment and increment the counter if necessary
+            if (index > loop.end) {
+                if (region_->loopCount && loop_.restarts >= *region_->loopCount) {
+                    wrappedIndex = loop.end;
```
jpcima: What about the `sampleEnd` portion after loop end?
this can never get played?

src/sfizz/SynthMessaging.cpp
paulfd: Added !

src/sfizz/Voice.cpp
paulfd: Yeah actually I revised it all and also included this part about loop_sustain to handle your comment. Now I find that it behaves differently than sforzando, but we can launch the squad to decide the best behavior.

paulfd:
Okay so I revised all of this code to handle both the `count`, `loop_count`, and `loop_mode` with the differences between `loop_sustain` and `loop_continuous`.

I distinguished 3 cases:
- `loop_continuous`: only loop on the sample. Upon release, the sample will continue to loop. If `loop_count` is also set, the voice will release itself after `loop_count`s loops but still continue to loop during the release phase.
- `loop_sustain`: only loop on the sample until release, then play as if it were a normal sample. If `loop_count` is also set, the voice will release itself after `loop_count`s loops and play the part after the loop.
- `one_shot`: well you ignore `loop_count` in this case but `count` is taken into account. In all `loop_` cases `count` is ignored.

If there are loop points in the sample file and no `loop_mode` in the SFZ file, we set `loop_mode` to `loop_continuous` and apply the above behavior. Maybe if `loop_count` is present setting the loop mode to `loop_sustain` would make more sense? It looks like it's what sforzando does, if `loop_count` is present the `loop_mode` is set to `loop_sustain` by default.

I created a test file with a kick, a looping hi-hat and a crash to test stuff if various players if you like.
[loop_count.zip](https://github.com/sfztools/sfizz/files/6090414/loop_count.zip)

kinwie:
I play your loop_count_test file in sforzando, DropZone and Rapture to compare to the current fix for loop_count.
Both sforzando and DropZone behave similar. I think in both, if `loop_count` is present `loop_mode` is set to `loop_continuous`, not loop_sustain.

From your test file, the loop_sustain is already fine.
Key d4 and f4 works similar like sforzando and DropZone.

Key e4 and g4 still not yet correct, imo, and I think the problem lies only at the `loop_count` behavior.
The behavior is, `loop_count` doesn't depend to the note-off message, unlike loop_sustain.

So,
- When a note is held as long as the sample or longer, it will play looped N times +1 and continue to the end portion of the sample (_loop-end point to end point_) and at full volume
- If the note released shorter than the sample and there is an ampeg_release, it will play looped N times +1 and continue to the end portion of the sample as long as value of ampeg_release (fade-out)

src/sfizz/Voice.cpp
```diff
@@ -918,63 +956,85 @@ void Voice::Impl::fillWithData(AudioSpan<float> buffer) noexcept
...
-            int(region_->trueSampleEnd(resources_.filePool.getOversamplingFactor())),
-            int(currentPromise_->information.end),
-            int(source.getNumFrames()))
-            - 1;
 
+        while (i < numSamples) { // In case we released within the block, continue as if it were a one-shot
+            (*indices)[i] -= loop.size * blockRestarts;
+            if ((*indices)[i] >= sampleEnd) {
+                fill<int>(indices->subspan(i), sampleEnd);
+                fill<float>(coeffs->subspan(i), 1.0f);
```
jpcima: Reduce the value of coeff into domain [0;1[
```suggestion
                fill<float>(coeffs->subspan(i), 0x1.fffffep-1);
```

kinwie:
I've checked the "Fine tune the looping behavior" build. The `loop_count` is now working fine :+1:

jpcima (approved):
OK to merge if it works

PR #680, opened 2021-03-07
Import foreign instruments

jpcima:
This adds support at plugin level for importing instrument formats other than SFZ.
As an example, it permits to load a sound file in the plugin.

When the importer has support for this file format (by file extension), it creates a SFZ string and makes sfizz load that.

The SFZ string for audio files is a simple `<region>sample=sound.wav`.

One can take the AudioFile code of this PR as a basis to add new formats.


jpcima:
Adding a series of MinGW fixes related to mismatched C++ standards from different components.
`fs::path` resolves to different types depending on C++ standard which is used for building.

PR #688, opened 2021-03-09
Support KDialog in VSTGUI

redtide:


jpcima:
It needs updating `askQuestion` too in the NativeHelpers.

PR #691, opened 2021-03-10
Don't use VSTGUI::CNewFileSelector::setDefaultExtension

redtide:
This fixes a wrong file extension filter in the file dialog under Windows.
It seems that the last added extension is selected as default one due to some wrong selection index. So use the All Files filter as workaround.

### DIFF
--- a/plugins/editor/src/editor/Editor.cpp
+++ b/plugins/editor/src/editor/Editor.cpp
@@ -1067,7 +1067,7 @@ void Editor::Impl::chooseSfzFile()
     SharedPointer<CNewFileSelector> fs = owned(CNewFileSelector::create(frame_));
 
     fs->setTitle("Load SFZ file");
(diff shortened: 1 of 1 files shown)


plugins/editor/src/editor/Editor.cpp
```diff
@@ -1067,9 +1067,9 @@ void Editor::Impl::chooseSfzFile()
     SharedPointer<CNewFileSelector> fs = owned(CNewFileSelector::create(frame_));
 
     fs->setTitle("Load SFZ file");
-    fs->setDefaultExtension(CFileExtension("SFZ", "sfz"));
 
     // also add extensions of importable files
+    fs->addFileExtension(CFileExtension("SFZ", "sfz"));
```
jpcima: SFZ is not among "importable files" as indicated in the comment.
Can you move it back up a few lines?

PR #693, opened 2021-03-10
Implement LFOs v1

jpcima:
This implements LFO v1 mostly.
Some opcodes still not done. (that's the depthcc, also to implement in v2)
Based on the large refactor PR.

Note that amplfo controls Volume not Amplitude. (outputs additive decibels)
All 3 LFO have behavior matched to ARIA.

EDIT(1) correct an implementation bug which would introduce a regression in v2 LFO


jpcima:
Merging this because it passes tests, checked against ARIA in various cases, and for avoidance of future conflicts following the refactor.

PR #703, opened 2021-03-13
Linear smoother

jpcima:
Implements the linear smoother, with SSE optimization, faster by 10x than the previous one.
This matches ARIA near perfectly. (except smoothcc still limited to max 100)


jpcima:
Old implementation is still there if one needs it.
In my system, `_mm_store_ps` vs `_mm_storeu_ps` makes next to no difference.

PR #710, opened 2021-03-17
Reset the smoothers

### LINKED ISSUE #382: Reset the smoothccs
In ARIA, when the playhead is repositioned, the smoothcc states are expected to be reset.

For example if a CC automation is manipulating the pitch
(127, 0, ..., 0)
and the track is played in a loop, the filter will smooth down, but it will not smooth up when the track loops to beginning.

paulfd:
Closes: #382 

In the end I added some code to check if the playhead moved. *If* the playhead moved and we're rendering with transport is playing, the smoothers are reset *and* the CC state is flushed. 

It seemed to match sforzando behavior in Reaper on Windows. It's not entirely what we discussed because upon looping you will have a discrete jump and no smoothing. The alternative is not as straightforward though, because looping seems to always involve splitting buffers in Reaper (as expected), which in turns mean you have a non-deterministic behavior of the smoother. To give an example, assuming a block size of 256:
- Upon looping the first time, the buffer ends exactly at the loop point. Rendering the new buffer, the host sends the CC state at delay 0, which "replaces" the previous state at loop end and no smoothing occurs.
- Upon looping the second time, loop point is mid buffer. The host splits the buffer and sends a CC state at delay 128. Now, a short smoothing will occur between delay 0 and 128.

All of this can be discussed and tweaked for sure. I'll try to compare more with the DAWs I have at hand, which is mainly Ardour now...

I also corrected some things in the BeatClock, @jpcima you might want to validate it. I'll comment them inline. There's also a small change in the MidiState which is actually a bug correction I think.

### DIFF
--- a/src/sfizz/BeatClock.cpp
+++ b/src/sfizz/BeatClock.cpp
@@ -157,6 +157,12 @@ absl::Span<const float> BeatClock::getRunningBeatPosition()
     return absl::MakeConstSpan(runningBeatPosition_.data(), currentCycleFrames_);
 }
 
+float BeatClock::getLastBeatPosition() const
+{
+    double beats = lastClientPos_.toBeats(timeSig_);
+    return static_cast<float>(beats);
+}
+
 absl::Span<const int> BeatClock::getRunningBeatsPerBar()
 {
     fillBufferUpTo(currentCycleFrames_);
@@ -175,31 +181,34 @@ void BeatClock::fillBufferUpTo(unsigned delay)
     for (unsigned i = fillIdx; i < delay; ++i)
         beatsPerBarData[i] = sig.beatsPerBar;
 
-    if (!isPlaying_) {
-        if (fillIdx < delay) {
-            fill(absl::MakeSpan(&beatNumberData[fillIdx], delay - fillIdx), 0);
-            fill(absl::MakeSpan(&beatNumberPosition[fillIdx], delay - fillIdx), 0.0f);
-        }
-        currentCycleFill_ = fillIdx;
-        return;
-    }
-
     BBT clientPos = lastClientPos_;
-    const double beatsPerFrame = beatsPerSecond_ * samplePeriod_;
-
     const BBT hostPos = lastHostPos_;
     bool mustApplyHostPos = mustApplyHostPos_;
 
-    for (; fillIdx < delay; ++fillIdx) {
-        clientPos = BBT::fromBeats(sig, clientPos.toBeats(sig) + beatsPerFrame);
+    if (!isPlaying_) {
         clientPos = mustApplyHostPos ? hostPos : clientPos;
         mustApplyHostPos = false;
 
-        // quantization to nearest for prevention of rounding errors
-        double beats = clientPos.toBeats(sig);
-        beatNumberData[fillIdx] = dequantize<int>(quantize(beats));
-        beatNumberPosition[fillIdx] = static_cast<float>(beats);
+        if (fillIdx < delay) {
+            double beats = clientPos.toBeats(sig);
+            // quantization to nearest for prevention of rounding errors
+            fill(absl::MakeSpan(&beatNumberData[fillIdx], delay - fillIdx), dequantize<int>(quantize(beats)));
+            fill(absl::MakeSpan(&beatNumberPosition[fillIdx], delay - fillIdx), static_cast<float>(beats));
+            fillIdx = delay;
+        }
+    } else {
+        for (; fillIdx < delay; ++fillIdx) {
+            clientPos = BBT::fromBeats(sig, clientPos.toBeats(sig) + getBeatsPerFrame());
+            clientPos = mustApplyHostPos ? hostPos : clientPos;
+            mustApplyHostPos = false;
+
+            // quantization to nearest for prevention of rounding errors
+            double beats = clientPos.toBeats(sig);
+            beatNumberData[fillIdx] = dequantize<int>(quantize(beats));
+            beatNumberPosition[fillIdx] = static_cast<float>(beats);
+        }
     }
+
 
     currentCycleFill_ = fillIdx;
     lastClientPos_ = clientPos;
--- a/src/sfizz/BeatClock.cpp
+++ b/src/sfizz/BeatClock.cpp
@@ -157,10 +157,9 @@ absl::Span<const float> BeatClock::getRunningBeatPosition()
     return absl::MakeConstSpan(runningBeatPosition_.data(), currentCycleFrames_);
 }
 
-float BeatClock::getLastBeatPosition() const
+double BeatClock::getLastBeatPosition() const
 {
-    double beats = lastClientPos_.toBeats(timeSig_);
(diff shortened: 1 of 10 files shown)


src/sfizz/BeatClock.cpp
```diff
@@ -175,32 +181,35 @@ void BeatClock::fillBufferUpTo(unsigned delay)
     for (unsigned i = fillIdx; i < delay; ++i)
         beatsPerBarData[i] = sig.beatsPerBar;
 
-    if (!isPlaying_) {
-        if (fillIdx < delay) {
-            fill(absl::MakeSpan(&beatNumberData[fillIdx], delay - fillIdx), 0);
-            fill(absl::MakeSpan(&beatNumberPosition[fillIdx], delay - fillIdx), 0.0f);
-        }
-        currentCycleFill_ = fillIdx;
```
paulfd, line 183: I changed this so that `currentCycleFill_` is set to `delay` even if the transport is not playing. It seemed more appropriate.

src/sfizz/BeatClock.cpp
```diff
@@ -175,32 +181,35 @@ void BeatClock::fillBufferUpTo(unsigned delay)
...
-
     BBT clientPos = lastClientPos_;
-    const double beatsPerFrame = beatsPerSecond_ * samplePeriod_;
-
     const BBT hostPos = lastHostPos_;
     bool mustApplyHostPos = mustApplyHostPos_;
 
-    for (; fillIdx < delay; ++fillIdx) {
-        clientPos = BBT::fromBeats(sig, clientPos.toBeats(sig) + beatsPerFrame);
+    if (!isPlaying_) {
```
paulfd, line 187: Now if we're not playing, the beat position vector is set to the last known position *or* the host position if it was sent. Before it was initialized to zero. I think it makes more sense like this but I'll let you validate it.

src/sfizz/MidiState.cpp
```diff
@@ -113,7 +119,7 @@ float sfz::MidiState::getLastVelocity() const noexcept
 
 void sfz::MidiState::insertEventInVector(EventVector& events, int delay, float value)
 {
-    const auto insertionPoint = absl::c_upper_bound(events, delay, MidiEventDelayComparator {});
+    const auto insertionPoint = absl::c_lower_bound(events, delay, MidiEventDelayComparator {});
```
paulfd, line 122: Considering there's a test below to check if the `delay` in the new event is the same as the `insertionPoint` one, I think I meant to use `lower_bound` here. If we have two events with the same delay in a block, the last received one is supposed to overwrite the previous one.

src/sfizz/modulations/sources/Controller.cpp
```diff
@@ -58,6 +75,7 @@ void ControllerSource::init(const ModKey& sourceKey, NumericId<Voice> voiceId, u
     if (p.smooth > 0) {
         Smoother s;
         s.setSmoothing(p.smooth, impl_->sampleRate_);
+        s.reset(impl_->getLastTransformedValue(p.cc, p.curve));
```
paulfd, line 78: I added this, I think this function can be call with a "non-zero" cc state, in which case the smoother should start at the current cc value I think.

src/sfizz/Synth.cpp
```diff
@@ -1296,7 +1303,16 @@ void Synth::timePosition(int delay, int bar, double barBeat)
...
 
-    impl.resources_.beatClock.setTimePosition(delay, BBT(bar, barBeat));
+    const auto newPosition = BBT(bar, barBeat);
+    const auto newBeatPosition = newPosition.toBeats(impl.resources_.beatClock.getTimeSignature());
+    const auto currentBeatPosition = impl.resources_.beatClock.getLastBeatPosition();
+    const auto positionDifference = std::abs(newBeatPosition - currentBeatPosition);
+    const auto threshold = 2 * static_cast<float>(impl.resources_.beatClock.getBeatsPerFrame());
+
+    if (positionDifference > threshold)
+        impl.playheadMoved_ = true;
```
paulfd, line 1318: This logic is not super pretty although it does make some sense. There may be a better way to assess whether the playhead moved.

jpcima: Perhaps reset them only at T=0 ?
It seems working in Reaper to check for bar=0 && beat=0 && playing=on.

src/sfizz/BeatClock.cpp
jpcima: This implementation certainly makes sense.
For completeness, the host position should probably be also applied in the non-playing case.
It's imaginable that play head can be moved also while stopped.

src/sfizz/BeatClock.h
```diff
@@ -134,6 +139,19 @@ class BeatClock {
      * @brief Get the time signature numerator for each frame of the current cycle.
      */
     absl::Span<const int> getRunningBeatsPerBar();
+
+    /**
+     * @brief Get the last BeatPosition
+     *
+     * @return float
+     */
+    float getLastBeatPosition() const;
```
jpcima: The unit is double. Nothing so important, only the initial implementation techniques have given me paranoia about the precision :) (not probably that important after switching to quantization)
As a policy, I picked double for all time keeping, except float in buffers.

src/sfizz/modulations/sources/Controller.cpp
jpcima: Imo, the smoother would rather keep the last target value given to it, and have something along the lines of `Smoother::resetToTarget`. The linear one already has this.
What do you think?

src/sfizz/BeatClock.cpp
paulfd: I think it is now, if you look below this line you have `clientPos = mustApplyHostPos ? hostPos : clientPos;`

src/sfizz/modulations/sources/Controller.cpp
paulfd: Sure, why not ! I'll add it.

src/sfizz/BeatClock.h
paulfd: OK

src/sfizz/Synth.cpp
paulfd: Yeah I was talking about the detection of the moved playhead. For the reset conditions there are many variations. It seems sforzando does not reset only on playhead moved but I'm not sure it's so natural.

paulfd: In Alex's example if you have the ability to render/bounce a region, only resetting at T=0 would smooth if the region you render does not start on 0.

src/sfizz/modulations/sources/Controller.cpp
paulfd: OK actually not a good idea, because usually when the playhead moves the host sends new CC value. You don't want to reset to the previous target, you want to reset to the new CC state so you need to query it fresh.

I did add the `resetToTarget` methods though, might keep them.

jpcima (approved):

PR #712, opened 2021-03-18
Timed recheck files

### LINKED ISSUE #689: MacOS Logic Pro Auto-Reload Bug
Using latest sfizz with MacOS Logic Pro, the auto-reload feature of sfizz works sporadically.

Sometime it will reload when I save the `file.sfz`- and other times I will need to close/open sfizz plugin. It also works sometimes when I load another SFZ script- and come back to the one in question.

jpcima:
Expected to fix #689, using semaphore timedwait
To test on various platforms.
@alcomposer


AlexMitchellMus:
Tested on MacOS Logic AU & Ardour VST3. Works as intended. 
1000ms feels fine.
But for seeing visual feedback when editing range/keyswitch etc 500ms or even 250ms may be better?

PR #721, opened 2021-03-19
Remove opcode limits

jpcima:
A work in progress


jpcima:
Most limits are lifted, except key and fx for now.
The former has a bit more complicated logic involved and I've not checked each single case.

Some opcodes remain oob-defaulted, it's the cases of enums and identifiers (for example cc numbers).

Ok for review

paulfd:
Wouldn't it make sense to just have permissive bounds as default?

src/sfizz/Range.h
```diff
@@ -32,6 +32,12 @@ class Range {
     {
 
     }
+
+    constexpr Range(const Range<Type, !Checked>& other)
```
paulfd, line 36: Why the `!Checked` here?

src/sfizz/Range.h
```diff
@@ -32,6 +32,12 @@ class Range {
...
+        : Range(other.getStart(), other.getEnd())
+    {
+    }
+
     constexpr Type getStart() const noexcept { return _start; }
```
paulfd, line 41: In the rest of the file all the `if (Checked) ...` could use the `IF_CONSTEXPR` macro for C++17, I think we have it.

paulfd (approved):

jpcima:
>Wouldn't it make sense to just have permissive bounds as default?

This would be the best idea starting from scratch, but I thought this easier for transition purposes.

src/sfizz/Range.h
jpcima: To allow `UncheckedRange` to be converted to `Range` and vice-versa

PR #723, opened 2021-03-19
Free-run the envelope if the sustain value is low enough

### LINKED ISSUE #676: No sound and CPU maxed with CC 64 sustain
- Send CC 64 value 127 to turn on sustain
- Play some notes
- Instrument goes silent
- Voice count never decreases

Problem reproducible with NoBudgetBand guitar, eg. `Guitar/LesPaul/lp_full.sfz`
http://www.bandshed.net/sounds/sfz/nbb_guitars.zip

paulfd:
Value taken from Sforzando

Closes #676


jpcima:
This is very near -50 dB, which hints it's likely correct :)

PR #733, opened 2021-03-21
Actually perform a CC when receiving the UI message

### LINKED ISSUE #684: Moving CC4 of Hihats from UI doesn't change the sound
I found an issue when testing some drums instruments that use CC4 for Hihats articulation switching (closed to open) :
_Moving CC4 from UI doesn't change the sound_

While moving from savihost CC, it works fine.

Example kit here, in case needed for checking :
https://github.com/sfzinstruments/WilkinsonAudio.NakedDrums

ND_sfizz

paulfd:
Also try to avoid updating the ui CC when the control is editing

Closes #684

### DIFF
--- a/plugins/editor/src/editor/Editor.cpp
+++ b/plugins/editor/src/editor/Editor.cpp
@@ -29,6 +29,7 @@
 #include <cstdarg>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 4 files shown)


jpcima:
Made obsolete by automation support in #786

PR #734, opened 2021-03-22
Adsr smooth transitions

### LINKED ISSUE #373: Noise print or denormal occur when ampeg_release not reached

jpcima:
Adds a linear smoothing in ADSR steps: Decay->Sustain, Release->0
Expected to fix #373
@kinwie


jpcima:
I'm not sure yet if unit test updates make total sense, but they pass.

paulfd (approved):

PR #747, opened 2021-03-25
Extended ccs

### LINKED ISSUE #481: Extended CC
https://sfzformat.com/extensions/midi_ccs

- [ ] 128: pitch bend, [-1:+1] per cycle
- [x] 129: channel aftertouch, ?
- [ ] 130: polyphonic aftertouch, ?
- [ ] 131: note on velocity, per voice [0:1]
- [ ] 132: note off velocity, ?
- [ ] 133: keyboard note number, per voice [0:1] normalized
- [ ] 134: keyboard note gate, per voice [0:1]
- [ ] 135: unipolar random, per cycle? [0:1]
- [ ] 136: bipolar random, per cycle? [-1:+1]
- [ ] 137: alternate, per cycle? [0:1]

paulfd:
Closes #481

### DIFF
--- a/src/sfizz/modulations/sources/Controller.cpp
+++ b/src/sfizz/modulations/sources/Controller.cpp
@@ -19,13 +19,15 @@ struct ControllerSource::Impl {
     float getLastTransformedValue(uint16_t cc, uint8_t curve) const noexcept;
     double sampleRate_ = config::defaultSampleRate;
     Resources* res_ = nullptr;
+    VoiceManager* voiceManager_ = nullptr;
     absl::flat_hash_map<ModKey, Smoother> smoother_;
 };
 
-ControllerSource::ControllerSource(Resources& res)
+ControllerSource::ControllerSource(Resources& res, VoiceManager& manager)
     : impl_(new Impl)
 {
     impl_->res_ = &res;
+    impl_->voiceManager_ = &manager;
 }
 
 ControllerSource::~ControllerSource()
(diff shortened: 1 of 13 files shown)


jpcima:
Do extended ccs not take in account smoothcc and stepcc value?

paulfd:
stepcc are not in yet, smoothcc should be configured already (for the smoother it's just a cc).

src/sfizz/modulations/sources/Controller.cpp
```diff
@@ -85,27 +87,105 @@ void ControllerSource::init(const ModKey& sourceKey, NumericId<Voice> voiceId, u
...
     const Resources& res = *impl_->res_;
     const Curve& curve = res.curves.getCurve(p.curve);
     const MidiState& ms = res.midiState;
-    const EventVector& events = ms.getCCEvents(p.cc);
+    bool canShortcut = false;
 
     auto transformValue = [p, &curve](float x) {
-        return curve.evalNormalized(x);
+        // Otherwise it clamps the bipolar values
+        if (p.curve != 0)
```
paulfd: This is not entirely satisfying but it's needed because the default curve (and all curves) "clamp" their inputs.

jpcima (approved):

src/sfizz/modulations/sources/Controller.cpp
jpcima: Should bipolar curves not be evaluated centered at 0.5 ?

paulfd: After some tests, it seems the curve is ignored for at least cc128 (bend) in Sforzando:
`<region> sample=*sine pitch_cc128=200 pitch_curvecc128=3` does not invert the polarity. In fact setting any predefined curve does nothing.

paulfd: `stepcc` does work though.

paulfd: Actually curve seems to be ignored in sforzando for extended ccs. I matched this behavior.

jpcima: It's correct, but we could still have it? it seems like a feature that may be useful.
Also Cakewalk does not have it either.

PR #750, opened 2021-03-25
Make the region stateless

jpcima:
It's work in progress in order to move the state information into `Layer`.

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -6,6 +6,7 @@
 
 #include "Region.h"
 #include "Opcode.h"
+#include "MidiState.h"
 #include "MathHelpers.h"
 #include "utility/SwapAndPop.h"
 #include "utility/StringViewHelpers.h"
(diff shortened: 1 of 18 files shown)


paulfd:
I'll check the details later, thanks ! For a general discussion, my understanding was that having more compact data structures to iterate over for state changes over regions was interesting for performance reasons. Here we actually have one more indirection. Could we make the `Layer` data structure in the synth as `vector<Layer>` instead of `vector<LayerPtr>`? Similarly, maybe it could be nice to copy some key elements from the region to the layer (regions do not mutate anyway) to avoid checking the region to update the state of the layer (i.e. key/velocity ranges, etc).

src/sfizz/Region.cpp
```diff
@@ -1697,7 +1511,7 @@ float sfz::Region::getBasePitchVariation(float noteNumber, float velocity) const
     return centsFactor(pitchVariationInCents);
 }
 
-float sfz::Region::getBaseVolumedB(int noteNumber) const noexcept
+float sfz::Region::getBaseVolumedB(const MidiState& midiState, int noteNumber) const noexcept
```
paulfd, line 1514: These could probably be entirely free functions or set within the layer which has the midistate dependency injected already?

jpcima: Either is valid to me, I think it can move without problems, just some needed unit test updates.

PR #752, opened 2021-03-29
Add CC modifiers to end, loop_start, loop_end

paulfd:
The end cc modifiers are computed at voice start like the offset.
The loop cc modifiers are updated in each block.

Should we consider adding a specific value for "max" range (right now it's in samples, but we could choose to have `end_cc1=max` which means the CC spans the whole sample range, or also `end_cc1=-max`).


jpcima:
>we could choose to have end_cc1=max which means the CC spans the whole sample range

IIRC there are some opcodes which accept -1, which possibly include `end`,that  ARIA seems to just cast it to an unsigned; effectively leading to the same effect.

src/sfizz/Region.cpp
```diff
@@ -1751,22 +1769,37 @@ float sfz::Region::getDelay() const noexcept
...
-uint32_t sfz::Region::trueSampleEnd(Oversampling factor) const noexcept
+uint32_t sfz::Region::getSampleEnd(Oversampling factor) const noexcept
 {
-    if (sampleEnd <= 0)
-        return 0;
+    int64_t end = sampleEnd;
+    for (const auto& mod: endCC)
+        end += static_cast<int64_t>(mod.data * midiState.getCCValue(mod.cc));
+
+    end = max(int64_t { 0 }, end);
```
jpcima: not clamped in both directions?

paulfd: You mean at sample end?

jpcima: Yes, if needed and not done already somewhere else

paulfd:
Not sure that `-1` is really an option here since technically you could have `loop_end=SAMPLE_END loop_end_cc1=-SAMPLE_END`, for whatever value `SAMPLE_END` is. In this case, the cc will move the loop end point backwards from the sample end, which can be a desirable behavior.

jpcima (approved):

PR #753, opened 2021-03-30
Add lfoN_fade_oncc, lfoN_delay_oncc, lfoN_phase_oncc

paulfd:
The phase modifier is actually an  "audio rate" modifier with curve, step and smoothcc options.
The v2 spec only had oncc, which would limit it to a set value with the LFO starts, but having it at audio rate makes sense if you want to tweak the phase while the LFO is running.

I'll add some tests.


paulfd:
Should be ok now !

src/sfizz/LFO.cpp
```diff
@@ -21,33 +19,30 @@
 namespace sfz {
 
 struct LFO::Impl {
-    explicit Impl(BufferPool& bufferPool, BeatClock* beatClock, ModMatrix* modMatrix)
-        : bufferPool_(bufferPool),
-          beatClock_(beatClock),
-          modMatrix_(modMatrix),
+    explicit Impl(Resources& resources)
+        : resources_(resources),
```
jpcima, line 23: This needs to update the demos programs.

src/sfizz/Region.cpp
```diff
@@ -1172,12 +1173,27 @@ bool sfz::Region::parseLFOOpcodeV2(const Opcode& opcode)
...
+        lfo.delayCC[opcode.parameters.back()] = opcode.read(Default::lfoDelayMod);
+        break;
     case hash("lfo&_fade"):
         lfo.fade = opcode.read(Default::lfoFade);
         break;
+    case hash("lfo&_fade_oncc&"):
+        if (opcode.parameters.back() > config::numCCs)
+            return false;
+
+        lfo.fadeCC[opcode.parameters.back()] = opcode.read(Default::lfoFadeMod);
```
jpcima, line 1195: That one possibly is dynamic, is it checked it in ARIA?
also, the fade possibly to change to linear.

paulfd: Huh, no I didn't but since there are no `smooth/step` version in sfzformat I assumed it was a straightforward one (like offset and the like). I'll check.

paulfd: > also, the fade possibly to change to linear.

What do you mean?

jpcima: just checked: linear and non-dynamic

src/sfizz/LFO.cpp
```diff
@@ -70,16 +65,25 @@ void LFO::start(unsigned triggerDelay)
...
+    for (const auto& mod: desc.delayCC)
+        delay += mod.data * state.getCCValue(mod.cc);
+
     size_t delayFrames = (delay > 0) ? static_cast<size_t>(std::ceil(sampleRate * delay)) : 0u;
     impl.delayFramesLeft_ = triggerDelay + delayFrames;
 
-    impl.fadePosition_ = (desc.fade > 0) ? 0.0f : 1.0f;
+    float fade = desc.fade;
+    for (const auto& mod: desc.fadeCC)
+        fade += mod.data * state.getCCValue(mod.cc);
```
jpcima, line 83: Can we get the well-timed value with respect to `triggerDelay` ? same also above

src/sfizz/Region.cpp
jpcima: > What do you mean?

Nvm misremembered, I thought this one was still using exponential.

jpcima (approved):

PR #758, opened 2021-03-31
Normalize percent values everywhere

jpcima:
Normalize all %, replace "percent-multiplicative" in modulations with regular multiplicative.
It simplifies by reducing the number of special cases in MM.

It helps a coming MM work which needs multiplicative source depths mods.


jpcima:
Affected opcodes are checked as working, and so are defaults.
(note: `ampeg_start` seems not working in aria)

PR #764, opened 2021-04-01
Support polyphonic aftertouch basics

### LINKED ISSUE #92: Support for Polyphonic Aftertouch
I would be interested in seeing Polyphonic Aftertouch as well. Devices such as the Roli Seaboard support it, as do any devices that support MPE. (The list is growing every day).

I understand that this would be a long term goal.

_Originally posted by @alcomposer in https://github.com/sfztools/sfizz/issues/7#issuecomment-561137996_

paulfd:
Closes #92

### DIFF
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -64,6 +64,7 @@ Synth::Impl::Impl()
     genFlexEnvelope_.reset(new FlexEnvelopeSource(voiceManager_));
     genADSREnvelope_.reset(new ADSREnvelopeSource(voiceManager_, resources_.midiState));
     genChannelAftertouch_.reset(new ChannelAftertouchSource(voiceManager_, resources_.midiState));
+    genPolyAftertouch_.reset(new PolyAftertouchSource(voiceManager_, resources_.midiState));
 }
 
 Synth::Impl::~Impl()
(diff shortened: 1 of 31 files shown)


src/sfizz/Synth.cpp
```diff
@@ -1343,6 +1344,26 @@ void Synth::hdAftertouch(int delay, float normAftertouch) noexcept
...
+{
+    Impl& impl = *impl_;
+    ScopedTiming logger { impl.dispatchDuration_, ScopedTiming::Operation::addToDuration };
+
+    impl.resources_.midiState.polyAftertouchEvent(delay, noteNumber, normAftertouch);
+
+    for (auto& voice : impl.voiceManager_)
+        voice.registerPolyAftertouch(delay, noteNumber, normAftertouch);
+
+    // Note information is lost on this CC
```
jpcima, line 1357: >    // Note information is lost on this CC

What is the reason for not keeping it?

paulfd: Well there is note + value on the polyphonic aftertouch event, and only value in the CC event, so we cannot "translate" it entirely.

It's only to start regions on polyphonic aftertouch using `on_lo/hicc130` and activate regions using `lo/hicc130` though, envelopes like e.g. `pan_cc130` will be per-voice and use the note information---see the polyphonicAftertouch section in `Controller.cpp`.

PR #769, opened 2021-04-03
image opcode

redtide:
Support for the [image](https://sfzformat.com/opcodes/image) opcode under the `<control>` header with image formats supported by [stb_image](https://github.com/nothings/stb#stb) 2.26 with scaling where needed.

### DIFF
--- a/plugins/editor/src/editor/Editor.cpp
+++ b/plugins/editor/src/editor/Editor.cpp
@@ -6,11 +6,11 @@
 
 #include "Editor.h"
 #include "EditorController.h"
-#include "EditorLibs.h"
 #include "EditIds.h"
 #include "GUIComponents.h"
 #include "GUIHelpers.h"
 #include "GUIPiano.h"
+#include "ImageHelpers.h"
 #include "NativeHelpers.h"
 #include "BitArray.h"
 #include "plugin/MessageUtils.h"
@@ -39,11 +39,6 @@ using namespace VSTGUI;
 
 const int Editor::viewWidth { 800 };
 const int Editor::viewHeight { 475 };
-
-struct image_deleter {
-    void operator()(unsigned char* x) const noexcept { stbi_image_free(x); }
-};
-typedef std::unique_ptr<unsigned char[], image_deleter> image_u;
 
 struct Editor::Impl : EditorController::Receiver, IControlListener {
     EditorController* ctrl_ = nullptr;
@@ -242,8 +237,6 @@ struct Editor::Impl : EditorController::Receiver, IControlListener {
         const char* keyName = keyNames[key % 12];
         return std::string(keyName) + ' ' + std::to_string(octave);
     }
-
-    static SharedPointer<CBitmap> loadAnyFormatImage(const fs::path& filePath);
 };
 
 Editor::Editor(EditorController& ctrl)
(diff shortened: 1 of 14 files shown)


plugins/editor/src/editor/Editor.cpp
```diff
@@ -18,6 +18,10 @@
 #include <absl/strings/ascii.h>
 #include <absl/strings/numbers.h>
 #include <ghc/fs_std.hpp>
+
+#define STB_IMAGE_IMPLEMENTATION
+#include <stb_image/stb_image.h>
```
jpcima: Compile stb_image implementation in a separate cpp file, to save compile time in Editor.
(example in `external/st_audiofile/src/st_audiofile_libs.c`)

plugins/editor/src/editor/Editor.cpp
```diff
@@ -39,6 +43,11 @@ using namespace VSTGUI;
 const int Editor::viewWidth { 800 };
 const int Editor::viewHeight { 475 };
 
+struct image_deleter {
+    void operator()(unsigned char* x) const noexcept { stbi_image_free(x); }
+};
+typedef std::unique_ptr<unsigned char, image_deleter> image_u;
```
jpcima:
```suggestion
typedef std::unique_ptr<unsigned char[], image_deleter> image_u;
```

plugins/editor/src/editor/Editor.cpp
```diff
@@ -311,6 +324,7 @@ void Editor::Impl::uiReceiveValue(EditId id, const EditValue& v)
             sendQueuedOSC("/key/slots", "", nullptr);
             sendQueuedOSC("/sw/last/slots", "", nullptr);
             sendQueuedOSC("/cc/slots", "", nullptr);
+            sendQueuedOSC("/image", "", nullptr);
```
jpcima, line 319: To add also in the other places of the same file, which send this same OSC sequence

plugins/editor/src/editor/Editor.cpp
```diff
@@ -1563,6 +1579,51 @@ void Editor::Impl::updateSWLastLabel(unsigned sw, const char* label)
         updateKeyswitchNameLabel();
 }
 
+void Editor::Impl::updateBackgroundImage(const char* filepath)
+{
+    const fs::path sfzFilePath = fs::u8path(currentSfzFile_);
+    const fs::path sfzDirPath = sfzFilePath.parent_path();
+    const fs::path imagePath = sfzDirPath / filepath;
```
jpcima: Some security in case of some MinGW which use ANSI path.
```suggestion
    const fs::path imagePath = sfzDirPath / fs::u8path(filepath);
```

plugins/editor/src/editor/Editor.cpp
```diff
@@ -1563,6 +1579,51 @@ void Editor::Impl::updateSWLastLabel(unsigned sw, const char* label)
...
+
+    int width, height, channels;
+    image_u image {
+        stbi_load(imagePath.c_str(), &width, &height, &channels, STBI_rgb_alpha)
```
jpcima: Has problems with Unicode paths on Windows OS, because `stbi_load` calls `fopen` (which uses ANSI encoding).
On windows, prefer `_wfopen` with `imagePath.wstring().c_str()` together with `stbi_load_file`.

plugins/editor/src/editor/Editor.cpp
```diff
@@ -1563,6 +1579,51 @@ void Editor::Impl::updateSWLastLabel(unsigned sw, const char* label)
...
+    unsigned char* imageData = image.get();
+
+    if (imageData) {
+        SharedPointer<CBitmap> bitmap = makeOwned<CBitmap>(width, height);
+        CBitmapPixelAccess* accessor = CBitmapPixelAccess::create(bitmap.get());
+
+        if (accessor) {
+            CColor c;
+            do {
+                accessor->getColor(c);
```
jpcima: Remove `getColor`, create a `CColor` out of the 4 components.

src/sfizz/Synth.cpp
```diff
@@ -93,6 +93,7 @@ void Synth::Impl::onParseFullBlock(const std::string& header, const std::vector<
         break;
     case hash("control"):
         defaultPath_ = ""; // Always reset on a new control header
+        image_ = "";
```
jpcima: It probably shouldn't reset on new `<control>` header.

jpcima (changes requested):

plugins/editor/src/editor/Editor.cpp
```diff
@@ -1563,6 +1579,51 @@ void Editor::Impl::updateSWLastLabel(unsigned sw, const char* label)
...
+{
+    const fs::path sfzFilePath = fs::u8path(currentSfzFile_);
+    const fs::path sfzDirPath = sfzFilePath.parent_path();
+    const fs::path imagePath = sfzDirPath / filepath;
+
+    int width, height, channels;
+    image_u image {
+        stbi_load(imagePath.c_str(), &width, &height, &channels, STBI_rgb_alpha)
+    };
+    unsigned char* imageData = image.get();
```
jpcima: Not needed after changing `image_u` as indicated, can address `image` directly

PR #771, opened 2021-04-03
Out-of-order API update

paulfd:
Updated the API documentation to mention that we require delay-ordered events.
Updated tests mostly to call on `renderBlock()` before testing things like number of active voices and such, just in case we reorder the events within the library later on.


paulfd:
@jpcima @redtide maybe just check that the API doc is OK !

jpcima (approved):

redtide (approved):

PR #780, opened 2021-04-05
Update abseil and ghc::filesystem

paulfd:


jpcima:
In abseil release notes, there was an indication about removal of some modules.
Makefile needs changing accordingly.

paulfd:
Yes I'll update it :D

PR #794, opened 2021-04-07
Fix Atom port corruption in LV2

jpcima:
Protects against corruption in case of overflowing the port buffer


jpcima:
The large number of parameter manages to overflow the buffer, allocate larger ports for safety

PR #803, opened 2021-04-09
Ordered event processing for VST

### LINKED ISSUE #58: Have VST process events and parameters in frame order
The VST parameters don't currently arrive in proper order.
It asks for an efficient way to sort the parameter points as they arrive from multiple queues.

### LINKED ISSUE #768: Have VST process events and parameters in frame order
The VST parameters don't currently arrive in proper order.
It asks for an efficient way to sort the parameter points as they arrive from multiple queues.

jpcima:
Fixes sfztools/sfizz-ui#58
This adds ordering of messages for VST3. It's expected to be efficient.
The memory usage is reduced by dividing the audio block in fixed subdivisions.

### DIFF
--- a/plugins/vst/CMakeLists.txt
+++ b/plugins/vst/CMakeLists.txt
@@ -38,6 +38,8 @@ add_library(sfizz-vst3-core STATIC
     SfizzVstUpdates.h
     SfizzVstUpdates.cpp
     SfizzVstIDs.h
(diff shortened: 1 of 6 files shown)


jpcima:
- [x] tested for correct frame times with subdivisions
- [x] tested working in small subdivisions
- [x] tested working remainders
- [x] fuzzed

PR #834, opened 2021-04-15
Direct OSC updates into the regions

jpcima:
Removes lock+vector system, replacing with direct update.


jpcima:
@rghvdberg, it's expected to fix the problem occurring in Ardour that you reported

PR #836, opened 2021-04-15
Api updates

jpcima:
Update API for beats per minute, arguments of type `int`


paulfd:
Should we add asserts or runtime checks for the int values (e.g. ignore OOB values)?

jpcima:
It's ok if parameter values go out of their bounds.
Anyway it goes to curves and these are no longer bounded.

paulfd:
They're also propagated to voices and regions though.

PR #855, opened 2021-04-24
Convert discrete CC conditions without gaps 

### LINKED ISSUE #852: Low-resolution range opcodes (like `lo/hicc`) have a "gap" when automated from the UI
Assume 2 groups `locc=0 hicc=64` and `locc=65 hicc=127`. You could put the UI knob in between 64 and 65 and neither group would play.

paulfd:
Something like `hicc1=13` is actually translated to float as `13.999999999/127` instead of `13/127`.

Closes #852


paulfd:
I dragged in #854 with it but it's OK.

jpcima (approved):

PR #864, opened 2021-04-27
Thread-safety for receiving parameters

jpcima:
This implements a thread-safety measure in the editor, because the Juce-based Carla host sends us the parameter updates in the wrong thread.
This regards the parameters changes of the level meters, performed by the DSP.
If this workaround is not implemented, this messes with the VSTGUI data, most likely the tracking of update rectangles, creating memory corruption.

### DIFF
--- a/plugins/vst/SfizzVstEditor.cpp
+++ b/plugins/vst/SfizzVstEditor.cpp
@@ -18,6 +18,7 @@
 #include "X11RunLoop.h"
 #endif
 #include <ghc/fs_std.hpp>
+#include <atomic>
 
 using namespace VSTGUI;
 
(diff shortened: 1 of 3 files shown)


falkTX:
Curious about this, only an issue with carla vst3 hosting?
Does the issue come from juce or carla? any other vst3 host that has this issue too?

jpcima:
I can't tell yet about Tracktion, I can give this a check.
We meet this problem since we have a code sends audio levels through parameters.
https://github.com/sfztools/sfizz/blob/7ca9c15a28bc7559a5b6f6e179e8f27ba4a737d5/plugins/vst/SfizzVstProcessor.cpp#L299-L305

jpcima:
@falkTX confirmation that Tracktion Waveform has the problem too. Reaper does not.

PR #866, opened 2021-04-28
VST thread safety+

jpcima:
This work permits to communicate DSP->Controller forcing it to use the controller thread, which was not the case in Reaper previously.
The message processing is expected to become thread-safe, taking advantage of work from the previous PR.
It's also expected to fix the standalone.

Notes
- [x] fixme, stuck notes on the VK


jpcima:
Note on crashes still occurring under this patch
`IMessage` and `IAttributeList`  have custom implementation in the JUCE host.
These are backed by an internal buffer, and this seens to cause corruption when `allocateMessage` is called from more than one thread.
It appears the solution is to create the message at the last moment, just before sending.

Issue #871, opened 2021-04-30
Controller reset in plugins, as result of automation


jpcima:
Problem occurring in Ardour with VST.
The CC 121 parameter is automated during an initialization step.
This resets sfizz's CC to zero values, including volumes, and the instrument is silent.

There are really 2 problems here
- [x] sfizz's reset function should not send zero, but instead the initial values according to the SFZ file
- [x] the VST cannot properly restore its settings

I vaguely recall about some VST documentation, which suggests that there shouldn't be a parameter automation happening as a result of another. I deduce that VST should not be processing reset-all-controllers at all.
In LV2, where the plugin's CC are host-controlled, it's probably safe to assume the same.

PR #872, opened 2021-04-30
Ensure plugin CC does not Reset all controllers

jpcima:
#871
This prevents reset-all-controllers which confuses the plugin state


jpcima:
This should fix a save/restore problem using Ardour

PR #896, opened 2021-05-20
Fill gaps on opcodes related to velocity and aftertouch triggers

paulfd:
Much like with the CC situation before, there could be situations were an instrument defined as
```
<region> ... lovel=0 hivel=50
<region> ... lovel=51 hivel=127
```
would not play any note sent with a high resolution velocity of `50.5f/127.0f`.

This impacts the crossfade gain calculations, so tests will fail and I have to check with Sforzando at least to not completely screw up the volumes on crossfades.


paulfd:
Ok seems good now!

jpcima (approved):

PR #898, opened 2021-05-21
Add the ability to cancel the release on the basic ADSR Envelope

### LINKED ISSUE #885: Sustain pedal realism
Hi there, Ever since I got into Linux audio for playing the piano with my MIDI keyboard, I noticed a unique, subtle trait of LinuxSampler together with SFZ soundfonts that made it feel distinctly more natural. With a sustain pedal, LinuxSampler allows for sustaining the residual resonance even after releasing all the keys. So basically, on a real piano, you hit some notes fairly hard, completely release them all, and then before their resonance dies out immediately hit the sustain pedal, and the residual resonance will continue for quite a while. LinuxSampler faithfully reproduces this behavior (~~only with SFZ and GIG soundfonts, the SF2 spec apparently doesn't support it~~ _EDIT: Actually it works with SFZ, GIG, and SF2_), and to my knowledge it's the only free SFZ sampler that does this. Sfizz does not. Even the Sforzando player for Windows which is generally regarded as the standard for free SFZ players doesn't do this. Apart from LinuxSampler, the only thing on my system that also has proper sustain behavior is the commercial Pianoteq app. But the problem is that LinuxSampler doesn't seem to exist as a Linux VST (only for Windows apparently?) so it's hard to use it with other DAWs and synthesizer apps. Would it be possible to implement this behavior on Sfizz? Thanks a lot!

paulfd:
Closes: #885 

- [x] Flex EGs
- [x] Global setting
- [x] Tests

I got bitten by the LV2 midi automation since the sustain pedal is basically ignored by default in all hosts atm...


paulfd:
I'm missing the UI but I'd like to streamline a bit the settings page in a separate PR if everyone is OK.

kinwie:
Yes

src/sfizz/ADSREnvelope.cpp
```diff
@@ -122,7 +122,8 @@ void ADSREnvelope::getBlock(absl::Span<Float> output) noexcept
                 break;
             }
             while (count < size) {
-                currentValue = std::max(sustain, currentValue + transitionDelta);
+                if (currentValue > sustain)
```
jpcima, line 125: Can you move the branch out the loop? that is
```
if()
  while () { .. }
else
  while () { ... }
```

paulfd: I wonder if actually it wouldn't just be
```c++
if (currentValue > sustain)
    currentValue += transitionDelta;
```

paulfd: (Note that there are branches in std::max, at least in libstdc++)

PR #917, opened 2021-06-07
Fix group cutting behavior with CC triggered samples (such as sustain pedals)

praashie:
While trying out the [Salamander Grand Piano](http://freepats.zenvoid.org/Piano/acoustic-grand-piano.html), I noticed that the sustain pedal sample was not stopped upon releasing the pedal. The SFZ file does include an opcode for cutting the activation sample:
```xml
//======================
//pedalAction

<group> group=1 hikey=-1 lokey=-1 on_locc64=126 on_hicc64=127 off_by=2 volume=-20

<region> sample=samples/pedalD1.flac lorand=0 hirand=0.5
<region> sample=samples/pedalD2.flac lorand=0.5 hirand=1

<group> group=2 hikey=-1 lokey=-1 on_locc64=0 on_hicc64=1 volume=-19

<region> sample=samples/pedalU1.flac lorand=0 hirand=0.5
<region> sample=samples/pedalU2.flac lorand=0.5 hirand=1
```

This commit fixes this problem.

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -1648,7 +1648,9 @@ bool Voice::checkOffGroup(const Region* other, int delay, int noteNumber) noexce
     if (impl.released())
         return false;
 
-    if (impl.triggerEvent_.type == TriggerEventType::NoteOn
+    if (
+        (impl.triggerEvent_.type == TriggerEventType::NoteOn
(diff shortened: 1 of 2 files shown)


paulfd:
Thanks alot for the catch and the patch! I have removed the newline and added a test case.

PR #924, opened 2021-06-22
lv2: A6 workaround, single output atom port

jpcima:
Previously, some `patch:Set` messages were directed to a distinct atom
port, when it was for notification purposes rather than automation.

Ardour does not support this, and the ports get assigned the same
buffer, producing corruption.

This commit returns to single-port behavior, and instead replaces
`patch:Set` notification messages with `sfizz:Notify` custom objects;
the UI processes these messages all the same, and they does not trigger
automation in the host.


paulfd:
Should we change the LV2 version number given that the ports have changed?

jpcima:
>Should we change the LV2 version number given that the ports have changed?

Yes, do you prefer now or before a release?

PR #931, opened 2021-06-26
Dynamic updates

paulfd:
I was working on this too! So I just merge the test I made and the dynamic update.

### DIFF
--- a/src/sfizz/FlexEnvelope.cpp
+++ b/src/sfizz/FlexEnvelope.cpp
@@ -57,6 +57,7 @@ struct FlexEnvelope::Impl {
     void process(absl::Span<float> out);
     bool advanceToStage(unsigned stageNumber);
     bool advanceToNextStage();
+    void updateCurrentTimeAndLevel();
 };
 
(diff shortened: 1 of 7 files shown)


src/sfizz/FlexEnvelope.cpp
```diff
@@ -172,6 +163,10 @@ void FlexEnvelope::Impl::process(absl::Span<float> out)
     const FlexEGDescription& desc = *desc_;
     size_t numFrames = out.size();
     const float samplePeriod = samplePeriod_;
+
+    if (desc.dynamic)
+        updateCurrentTimeAndLevel();
```
jpcima, line 168: In my experience, `egN_dynamic` operates at audio rate.
It never feels right to update at buffer intervals, because this is not predictable.
It's better IMO having 2 versions of process, one for static and one for dynamic.

jpcima:
OK for merging except `dynamic`

paulfd:
Dynamic is the point of the PR so..

PR #935, opened 2021-06-28
Store files to load and their offset and defer loading

### LINKED ISSUE #912: Bug: Pianobook Autumn Piano / monolithic libraries not loading
Attempting to open the library results in indefinite loading. The plugin doesn't stop to load, so I have to shut down the DAW.

This is the library: https://www.pianobook.co.uk/library/autumn-piano/

I believe that specifically those libraries made by Christian Henson use a less common method: monolithic audio file with no separate samples per note, but one file containing all the notes with specific timestamps being referenced in the mapping. There seems to be a similar issue when loading the Decent Sampler version on that platform: 
https://www.pianobook.co.uk/forums/topic/having-trouble-with-christian-hensons-samples-in-decent-sampler-format/

VST 3 version on Windows, Ableton 11

paulfd:
Closes: #912

Better behavior when using a file multiple times with different offsets. This does not solve the high memory usage for monolithic libraries which will basically load the full sample in memory; for this I'll track another issue.

I'll keep adding to this PR to try and debug what's going on with this library: it showed other issues (volume is high and polyphony too).

### DIFF
--- a/src/sfizz/Layer.cpp
+++ b/src/sfizz/Layer.cpp
@@ -129,7 +129,7 @@ bool Layer::registerNoteOff(int noteNumber, float velocity, float randValue) noe
     return false;
 }
 
-bool Layer::registerCC(int ccNumber, float ccValue) noexcept
+bool Layer::registerCC(int ccNumber, float ccValue, bool dontTrigger) noexcept
 {
     const Region& region = region_;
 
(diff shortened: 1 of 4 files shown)


paulfd:
~Ok it seems fine, but the pedal up/down noises are missing for some reason.~

PR #945, opened 2021-07-06
CC input

### LINKED ISSUE #882: FR: CC Keyboard entry
Currently CC controls can only be changed via mouse.

There should be a way to enter in values via keyboard, as sometimes (depending on what the CC is connected to internally) a mouse adjustment may not be precise enough, especially for a specific value.

paulfd:
Closes: #882 

- Add an EditValue in the knob
- Shade the knobs when editing
- Right-click on the knob allows to change from low resolution to high resolution CC

@jpcima I'm having a remaining issue with mostly (?) the Pan knob (second knob in the page). When you try to move it it jumps. This doesn't happen with other knobs and for the life of me I can't really find why atm.

Edit: ^ does not bug out on Windows....

### DIFF
--- a/plugins/editor/src/editor/Editor.cpp
+++ b/plugins/editor/src/editor/Editor.cpp
@@ -162,11 +162,11 @@ struct Editor::Impl : EditorController::Receiver,
     SharedPointer<CBitmap> backgroundBitmap_;
     SharedPointer<CBitmap> defaultBackgroundBitmap_;
 
-    CControl* getSecondaryCCControl(unsigned cc)
+    SKnobCCBox* getSecondaryCCKnob(unsigned cc)
     {
         switch (cc) {
-        case 7: return volumeCCKnob_ ? volumeCCKnob_->getControl() : nullptr;
-        case 10: return panCCKnob_ ? panCCKnob_->getControl() : nullptr;
+        case 7: return volumeCCKnob_ ? volumeCCKnob_ : nullptr;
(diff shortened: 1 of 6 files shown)


AlexMitchellMus:
**I understand this is a WIP**
I have tested this feature, here are a few UX thoughts:

1) Right click menu is fantastic, works really well for the hi/lo res mode select.
2) There needs to be a visual way to distinguish between hi/lo res mode. I suggest a badge in the `CC number` space. Because right now, a musician has no way to know if `value 1` is `0-1` or `0-127`.
3) It's too easy to mistakenly enter the keyboard entry mode for knobs. Possibly use double click, or alternatively place it into the right click menu `Edit Value`, as how Surge achieves this: 
surge_edit_value
4) I'm not sure we need the grey highlight behind the knob. Right now it doesn't feel integrated.

AlexMitchellMus:
Two notes from 63c1af9:

* Ardour Linux doesn't default to linear mouse knob mode
* Data entry for value should be limited to numeric only

paulfd:
- Ardour forces all UIs to the relative circular knob mode
- Invalid input result is ignored, is that a big issue? Seems OK to me like this...

AlexMitchellMus:
Ardour doesn't use rotary mouse interaction knob mode itself... I think it would be best to make it standard.

Is this something on the Ardour end?

paulfd:
Ardour should be fixed iirc, so I'll merge this.

PR #946, opened 2021-07-09
Don't draw a white keyboard if all keys are used

### LINKED ISSUE #900: max keyranges not showing color on the keyboard
This :
```
<region>
sample=*sine
```

Or this :
```
<region>
sample=*sine
lokey=0
hikey=127
```

not showing the color range on the keyboard ui

Change it to :
lokey=1 hikey=127, or
lokey=0 hikey=126
does shown the color range

paulfd:
Closes: #900


jpcima (approved):
If it's preferred this way, no problem

PR #948, opened 2021-07-16
Use glib's find in path

### LINKED ISSUE #60: Automatically detect the zenity bin path required by VSTGUI X11::FileSelector
Currently there is a `zenity` hard-coded path in both [VSTGUI] and editor's [NativeHelpers].
It would be possible to auto-detect them by [searching in PATH] for systems like NixOS.
See sfztools/sfizz#505.

[VSTGUI]: https://github.com/sfztools/vstgui/blob/f4fe0d0b1e53443342ba732a0e715ce5c96b49df/vstgui/lib/platform/linux/x11fileselector.cpp#L20
[NativeHelpers]: https://github.com/sfztools/sfizz/blob/develop/plugins/editor/src/editor/NativeHelpers.cpp#L115
[searching in PATH]: https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-find-program-in-path

paulfd:
Unlike the similar VSTGUI change, we already include gio in sfizz so we can use @redtide 's proposed function.
Closes: sfztools/sfizz-ui#60 (along with the proposed VSTGUI patch on our fork)
I believe nothing calls this code atm, I have to try it out manually.

### DIFF
--- a/plugins/editor/src/editor/NativeHelpers.cpp
+++ b/plugins/editor/src/editor/NativeHelpers.cpp
@@ -198,12 +198,21 @@ static std::vector<char *> createForkEnviron()
     return newEnv;
 }
 
-static constexpr char zenityPath[] = "/usr/bin/zenity";
+static const std::string zenityPath = [] {
+    auto glibPath = g_find_program_in_path("zenity");
+    if (glibPath) {
+        std::string s { glibPath };
+        g_free(glibPath);
+        return s;
+    } else {
+        return std::string("/usr/bin/zenity");
+    }
+}();
 
 bool askQuestion(const char *text)
 {
     char *argv[] = {
-        const_cast<char *>(zenityPath),
+        const_cast<char *>(zenityPath.c_str()),
         const_cast<char *>("--question"),
         const_cast<char *>("--text"),
         const_cast<char *>(text),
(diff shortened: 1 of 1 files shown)


redtide:
I forgot some details about this but the code should be also changed in the same way upstream on VSTGUI.

paulfd:
Yes I have an equivalent PR in our fork.

Jul 25, 2021 12:21:46 redtide ***@***.***>:

plugins/editor/src/editor/NativeHelpers.cpp
```diff
@@ -198,12 +198,16 @@ static std::vector<char *> createForkEnviron()
     return newEnv;
 }
 
-static constexpr char zenityPath[] = "/usr/bin/zenity";
+const char* zenityPath()
+{
+    static const char* zenityPath = g_find_program_in_path("zenity");
```
jpcima: Release glib's path with `g_free` after done.
other: zenityPath can be null if not found; fallback to the hardcoded value?

paulfd: Better like this?

jpcima: It's good

PR #954, opened 2021-07-20
Parse oscillator=auto and only have implicit wavetables if the file is short AND wavetable tags are in

### LINKED ISSUE #916: Short audio files don't get played completely
I am creating a mapping for some drum kit library, and noticed that very short files don't get played out correctly. This happened under Linux, against JACK backend, using sfizz-git, branch "development", commit ID ce2170c88d.

I have attached a ZIP archive [testdata.zip](https://github.com/sfztools/sfizz/files/6604471/testdata.zip) with an example .sfz and 3 short files of pink noise to show the effect.
The 0.2s and 0.1s duration files get played out correctly; the 0.05s file is only played partially.
This is also seen in the enclosed screenshot from a jack_capture run of that session.
truncated_audio

Even though this might look like a "synthetic" test, there are in fact very short (hihat) samples that have roughly the same duration.
I assume this is not some kind of automatic "noise gate" kicking in?
If more info is needed, let me know. Thanks.

paulfd:
Closes: #916


src/sfizz/Synth.cpp
```diff
@@ -638,7 +638,7 @@ void Synth::Impl::finalizeSfzLoad()
                     region.sampleId.reset(new FileId("*silence"));
                     region.hasWavetableSample = false;
                 } else {
-                    region.hasWavetableSample |= true;
+                    region.hasWavetableSample &= true;
```
jpcima: It's a useless statement.

jpcima (approved):

PR #961, opened 2021-07-24
Remember the last keyswitch used in the states

### LINKED ISSUE #939: set articulation not being recalled in SFZ patches using keyswitches
I tried it using the Vibraphone - Keyswitch SFZ patch from the VCSL (v 1.2.2) library by Versilian. I played a MIDI sequence switching the articulation to "Hard Mallets". When I reload the project in my DAW, the set articulation will not be recalled. The keyswitch setting is not remembered and I get the initial setting of the patch.

Is this intended behaviour? Sure, I could add an extra note to my MIDI sequence telling Sfizz which keyswitch articulation to use, but find it preferable if the articulation that was last being used could be also recalled.

paulfd:
Closes: #939


jpcima:
Rebased on develop

PR #972, opened 2021-08-17
Fix `off_by` behavior with long-releasing notes

### LINKED ISSUE #971: Mod Duo choke groups dont work (8.0-0)
Hello, Im running 8.0-0 on my Mod Duo X (not sure if this applies to other versions of sfizz, only use it on my Mod Duo X) and it is completely ignoring choke groups.

praashie:
This resolves #971. New test case included.

### DIFF
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -1666,7 +1666,7 @@ bool Voice::checkOffGroup(const Region* other, int delay, int noteNumber) noexce
     if (region == nullptr || other == nullptr)
         return false;
 
-    if (impl.released())
+    if (impl.released() && region->offMode == OffMode::normal)
         return false;
 
     if ((impl.triggerEvent_.type == TriggerEventType::NoteOn
--- a/src/sfizz/Voice.cpp
+++ b/src/sfizz/Voice.cpp
@@ -1666,7 +1666,7 @@ bool Voice::checkOffGroup(const Region* other, int delay, int noteNumber) noexce
     if (region == nullptr || other == nullptr)
         return false;
 
-    if (impl.released() && region->offMode == OffMode::normal)
+    if (impl.offed_)
         return false;
 
     if ((impl.triggerEvent_.type == TriggerEventType::NoteOn
--- a/tests/SynthT.cpp
+++ b/tests/SynthT.cpp
@@ -1568,6 +1568,26 @@ TEST_CASE("[Synth] Off by alone and repeated")
     REQUIRE( numPlayingVoices(synth) == 3 );
 }
 
+TEST_CASE("[Synth] Off by with staccato notes")
+{
+    sfz::Synth synth;
+    sfz::AudioBuffer<float> buffer { 2, static_cast<unsigned>(synth.getSamplesPerBlock()) };
+
+    synth.loadSfzString(fs::current_path(), R"(
+        <region> group=1 off_by=1 sample=*sine ampeg_release=2
+    )");
+    synth.noteOn(0, 60, 85);
+    synth.renderBlock(buffer);
+    REQUIRE( numPlayingVoices(synth) == 1 );
+    synth.noteOff(0, 60, 85);
+    synth.renderBlock(buffer);
+    REQUIRE( numPlayingVoices(synth) == 0 );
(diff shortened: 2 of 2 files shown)


paulfd:
What do you think about testing if the voice was already "offed"? Simply with
```
if (impl.offed_)
...
```
I tried it and it passes your test. This check used to be there so that we don't "off" a voice twice, but this boolean didn't exist then, so I used the release state.

However, and unrelated to this PR, I should check what happens with release samples and long release tails on the attack sample.

praashie:
The problem of the same note not muting itself is still present. Your `impl.offed_` version does fix the problem with different notes played in staccato.

I looked more into the code base and I found the following test:

https://github.com/praashie/sfizz/blob/e4b4dcd6553edcf94e81c7bddbb38dda35be3023/tests/SynthT.cpp#L1621-L1638

This test came from commit 863a08c42, but the behavior is [not mentioned in sfzformat.com](https://sfzformat.com/opcodes/group) . Is sfizz intended to replicate this quirky (but not clearly specified) behavior?

paulfd:
Apparently this was sfz.dll's behavior, although indeed sforzando does not behave like this. I'm guessing the idea was to allow buildup when playing the same note repeatedly but still choking different notes (probably for some drum thing).

Note that sfzformat is a (fairly complete by now) reverse engineering work-in-progress. Many quirks like this are yet to be documented, and usually we discover them on the fly.

For this case in the end, one solution could be to add a `hint_...` switch to enable the sfz.dll behavior, but keep sforzando's by default. When these 2 disagree it's a bit annoying.

kinwie:
> 
> 
> The problem of the same note not muting itself is still present. Your `impl.offed_` version does fix the problem with different notes played in staccato.
> 
> I looked more into the code base and I found the following test:
> 
> https://github.com/praashie/sfizz/blob/e4b4dcd6553edcf94e81c7bddbb38dda35be3023/tests/SynthT.cpp#L1621-L1638
> 
> This test came from commit [863a08c](https://github.com/sfztools/sfizz/commit/863a08c421d563b11ae425a3ece59f89b8dc1c30), but the behavior is [not mentioned in sfzformat.com](https://sfzformat.com/opcodes/group) . Is sfizz intended to replicate this quirky (but not clearly specified) behavior?

Logically, group=1 with off_by=1 muting itself is a true condition and that's what happened in sforzando. But it will has issue in practical usage, for example playing back auto-converted sfz, for example sfz that converted from sf2 with Polyphone, because afaik auto-converted program can only detect and assign the same number for `group` and `off_by` opcodes for the "Exclusive Group" setting. So then, the region will become monophonic that cut itself when played in sforzando. Iirc, even sforzando's own sf2-converter also omitted this exclusive group setting, so a dropped sf2 (e.g a drumkit hihat) won't choke. Another issue, if the stereo samples splitted as two mono regions (L-R), only 1 channel will be sounded. The other one will be muted because of this same `group` and `off_by` number behaviour.

So, what sfz.dll and DimPro/Rapture does is more nicer imho. They don't choke itself. Most general samplers also suffer from this issue and they fix it million years ago, e.g. NN-XT added a "Group Mono" button to solved it in version 4. Kontakt add the "Exclusive Group" to the Voice Group in version 1.5, etc, etc. Here I quote what Toontrack said about Polyphonic Muting, from dfh Superior manual.pdf from 2004 (yeah! It's a known issue since a long long time ago) :

Untitled

With the "Not self-muting" behaviour will make sfizz better imho. This behaviour not documented in sfz, well...lot of things are still mystery in sfz!

To get the same identical monophonic result of :
```
group=1
off_by=1
```
We can simply just use :
```
group=1
polyphony=1
```
Sfz is just a text file and you can advise people to just edit and change the `off_by=N` to `polyphony=1`. It just that simple...

kinwie:
@paulfd 
Too many `hint` :))

paulfd:
I find these arguments reasonable. I'm not developing instruments but at least this way we have the two behaviors, and it matches the original sfz. I can merge as is which already fixes one of the problems!

kinwie:
Thanks @paulfd .
I'm so afraid that this great original sfz.dll behavior gone again...
You've perfected it with fixing the CC usage which was a limitation in cakewalk players

PR #973, opened 2021-08-20
Implement an internal CLI for loading soundfonts and setting engine parameters.

jofemodo:
BTW, add a new command line argument (--num_voices).

### DIFF
--- a/clients/jack_client.cpp
+++ b/clients/jack_client.cpp
@@ -151,7 +151,8 @@ static void done(int sig)
 
 ABSL_FLAG(std::string, client_name, "sfizz", "Jack client name");
 ABSL_FLAG(std::string, oversampling, "1x", "Internal oversampling factor (value values are x1, x2, x4, x8)");
-ABSL_FLAG(uint32_t, preload_size, 8192, "Preloaded value");
+ABSL_FLAG(uint32_t, preload_size, 8192, "Preloaded size");
+ABSL_FLAG(uint32_t, num_voices, 32, "Num of voices");
 ABSL_FLAG(bool, state, false, "Output the synth state in the jack loop");
 
 int main(int argc, char** argv)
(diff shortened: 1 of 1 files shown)


clients/jack_client.cpp
```diff
@@ -149,30 +153,146 @@ static void done(int sig)
...
+        std::cout << "===========================" << '\n';
+        std::cout << "Import format: " << importFormat << '\n';
+    }
+    // std::cout << std::flush;
+
+    return true;
+}
+
+
+std::vector<std::string> string_tokenize(std::string str) {
```
paulfd: `str` could be const ref here

clients/jack_client.cpp
```diff
@@ -149,30 +153,146 @@ static void done(int sig)
...
+
+        if (kw=="load_instrument") {
+            //args.erase(std::remove(str.begin(), str.end(), '\"'), str.end());
+            load_instrument(tokens[0].c_str());
+        }
+        else if (kw=="set_oversampling") {
+            try {
+                synth.setOversamplingFactor(stoi(args));
+            } catch (...) {
+                std::cout << "ERROR: Can't set oversampling!\n";
```
paulfd, line 246: None of these throw, however they're not thread-safe so we need a mutex in the callback. You can check the `SfizzVstProcessor.cpp` file for an example, especially the `doBackgroundWork()` and `process()` functions. Basically you need to fully lock in this `cli_thread_proc()` method, and use `try_lock()` in the real-time thread (and exit if you can't take the lock).

clients/jack_client.cpp
jofemodo: Will fix ...

clients/jack_client.cpp
jofemodo: Done! I've been testing and the "random breaks" when loading soundfonts seems to be fixed.
Thanks!

paulfd: You may still have breaks for big soundfonts. If you want to avoid them you could somehow fade out the current loaded file before loading a new one in, but none of our plugins does this currently.

jofemodo: Could you elaborate a little bit what you mean for "fade out" the "currently loaded soundfont"? ;-)
Do you mean ending active notes before loading the new soundfont?

paulfd:
I made a change and did some formatting but I cannot push to your branch. Would you mind allowing it? Otherwise I'll do another PR from my fork.

jofemodo:
You should allowed to write in my fork now.

Thanks!

jofemodo:
Hi @paulfd !

Is there some problem with merging this? Can i do something? We are preparing a new stable version and we would like to install sfizz from your deb packages, so we can keep updated easily.

Thanks!

jofemodo:
BTW, I sent you and invitation to write on my fork but it's pending yet.

paulfd:
Hi ! Sorry I'm too taken in RL things, and I missed the invitation which is now expired. You can resend it, but I think there's a checkbox in the right column of the pull request interface that says "Allow edits by maintainers" which would allow me to write on your branch without having write access on the whole repository. Then I can push my changes and if you're OK with them we can merge :)

jofemodo:
Hi @paulfd !

I didn't find this check anywhere. Perhaps it's only available when creating the PR?
Anyway, i resend the invitation ;-)

paulfd:
We can let the CI run although there are no "tests" of the program. In the meantime maybe check my changes ! Once the command-line interface is stable on your side maybe it'd be good to write down some tests so that we don't accidentally regress.

jofemodo (commented):
Everything looks OK now ;-)
I will write some tests ASAP.

jofemodo:
BTW, are the tests required for merging this PR?

paulfd:
No no

Issue #207, opened 2020-05-03
Routing of effects as a graph


jpcima:
Some notes on implementing a future effect graph

See the pages 351-352. It has an example of using it for routing.
This need the effects `buffer`, `mixer` and the buses `auxN`.

- a `buffer` serves to send its input to a designated destination bus
  - `buffer_source` names the input. It is set to 0 to indicate the buffer effect's input signal
  - `buffer_destination` names a bus (eg. `auxN`) where to add the source signal

- a `mixer` serves to add signals from a pair of buses
  - `source_1` typically 0, it indicates the mixer effect's input signal
  - `source_2` a secondary bus to add in the signal (eg. `auxN`)

---
To implement this, we need to compute an order over all effects.
It's a dependency graph, and needs computation once after the sfz load.
The purpose is to compute the chains which produce `auxN` (buffer) prior to computing another effect that requires it (mixer).

---
**Idea:** introduce some additional I/O ports to the Sfizz plugin.
If we extend this model to designate `inN` / `outN` as buses mapped to extra plugin ports, we can have some abilities
- to use Sfizz as an effect processor only (effect bus receives `inN`, writes `outN`)
- to route distinct effect chains into distinct audio ports

One could imagine an example like the following, which routes signal unchanged into the 2nd stereo output.
```
<effect>
bus=fx2
fx2tomain=0
type=buffer
buffer_source=0
buffer_destination=out2
```
Maybe `bus=out2` as a simplified way of writing?

AlexMitchellMus:
So your suggesting the final example could be written using new opcodes:
```
<effect>
bus=out2
```

jpcima:
Yes, that would be the idea. This wouldn't have an equivalent simplified form for input though.
Implementation-wise, this would not make it a big difference.
This still sticks reasonably to the Cakewalk effect specification + the book addition.

paulfd:
Yeah I had something similar in mind in order to transparently handle both the multiple output cases and the effect inserts.

PR #1002, opened 2021-10-29
Multiple outputs

### LINKED ISSUE #9: [FR] Add support for multiple (stereo) output channels
Relevant opcode:

* https://sfzformat.com/opcodes/output

The main use case is SFZ files for drum sets, where one would want to route different drum sounds to different output channels, so they can be routed to different track in the DAW, e.g. for treating them with different effects or do custom panning.

paulfd:
Closes: #9 

As a first step, this creates identical copies of the effect buses described in the SFZ file on each output. The effect buses also then provide the temporary buffers that are mixed onto the actual output. The render methods have been updated to actually handle more than 1 stereo channel as an output. If there are more outputs in the SFZ file than in the buffer-to-render, SFZ outputs are "wrapped" onto the available rendering output.


jpcima:
What do you think of the idea from #207 ?
The duplicated multi-out vs the bus output seem conflicting approaches.

paulfd:
Actually I was thinking that the `auxN` buses could use the idea from #207, and the `main` and `fxN` buses would be duplicated.

The main advantage I see with this solution is that it's pretty transparent for users: you put an effect and it appears on all outputs by default, so it kind of does what you'd expect regardless if you have one or more outputs. If you need finer grained effect then you could use aux buses, to be implemented of course.

One alternative would be to specify the output each effect applies to, in which case it's a bit more involved and verbose.

jpcima:
Sure, but on the other hand, it creates difficulty with some future perspectives.
In particular, if one wants to give introspection and edition abilities into a graph-like processor.
I'm thinking of script API in particular.

paulfd:
So you'd rather that we have the `output` opcode on the effects, so that all effects are on the output 0 buses by default but you could put it on another one?

Either way you can introspect 😄

PR #1020, opened 2021-11-10
Building with emscripten

paulfd:
This branch is maintained with a thread-less version of sfizz. The background loader is removed and all files are loaded in RAM.


ephemer:
Hi @paulfd, I have a question about this: if "all files" are loaded in RAM, does that mean uncompressed files or the compressed ones? Would be a big difference for RAM usage of course

PR #1037, opened 2021-11-20
Clamp flex EG levels

### LINKED ISSUE #1032: units for egN_pitch
My first contribution on any github project ever. Be gentle with me!

I'm a musician rather than a developer. Not sure if this is a bug or an issue.

Running sfizz1.1.1-0 inside Carla on Linux Peppermint 10.

I've loaded Unreal Instruments Two Whistles. All the instruments that utilize the egN_pitch opcode are glitchy. I've worked out that while the instruments are coded to use cents as the unit I think sfizz uses semitones. So I just get a pop or a silence where there should be a specific pitch.

I don't remember that being the case earlier this year when I was using an earlier version of sfizz, but I could be mistaken.

There's no entry for units in the sfzformat.com/opcodes listing for this parameter. And I don't currently have any other examples I can point to. But cents would seem to make more sense, in the same way it is used for calibrating LFOs.

Let me know if you need more info.

paulfd:
Sforzando seems to do so too.

Closes: #1032


jpcima:
What happens if using a `egN_curveX` which extends out of ±1.0 range?

paulfd:
IIRC it's ignored, even when you remove the `egN_curveX`.

Issue #91, opened 2020-02-29
Embedded sample data


jpcima:
Dropzone and Rapture SFZv2 supports embedding of samples.

See: https://github.com/sfzformat/sfzformat.github.io/issues/23

Note that parsing this data is necessarily context-dependent; the parser needs the knowledge that it's handling a binary-valued opcode and treat accordingly.

If it were to handle this case, the current parser would need adapting.
- open the `ifstream` with binary mode
- don't replace all the whitespace with `' '` in advance (would corrupt the coding)
- don't look at the next `=` for opcode's end, but rather the `$` final marker as indicated in the reference post

jpcima:
Is the channel sfztool, @rghvdberg has mentioned an idea of a virtual filesystem that can load SFZ files and/or samples. (not sure I got the idea correctly, elaborate as needed)

It's not unlike the sample embedding above, but except this could be extended to SFZ files when you load them by `#include`.

The target of `#include` could refer to a pseudo-file that is registered in sfizz's virtual paths, and this can serve as an entry point to inject software-generated SFZ elements into sfizz, dynamically.
It's a quick, and maybe a little dirty, way of realizing some custom sfz-based synths.

paulfd:
So with #1041 base64 data can be read from the sfz file, only missing AIFF. I'm kind of on the fence about the parsing format of Rapture: it's not really practical to have a binary format within a text file.

However, reading zlibbed/gzipped data encoded as base64 could be nice to gain a little efficiency.

PR #1041, opened 2021-11-21
Parse <sample> with base64 data

paulfd:
Also added the necessary plumbing to load from RAM.
It's missing AIFF, and metadata reading since those API are heavily file-based for now.
Note that this is not exactly what's proposed in #91, since the goal is also to allow to "copy" these embedded samples as printable characters (e.g. in an edit box on the web, or a pastebin).

The syntax is as follows:
```
<region> sample=test.wav
<sample> name=test.wav encoding=base64 data=[... Base64 data ...]
```

Samples like this will be loaded fully in RAM; I made some modifications to the file pool to check both "fully loaded" and preloaded files on requests, which wasn't the case before.
The `encoding=base64` is required to try and keep a compatibility path with the Rapture encoding mentioned in #91 down the line.
When reading an opcode value, the parser now behaves differently if the opcode name is `data`: it will ignore comments since base64 has the `/` sign, and will only stop on EOF, a new `<` character, or a new `#` character.
The underlying base64 decoder ignores whitespace, and can support optional padding with `=`.

Now that I think about it and write it down, it's not going to be compatible as is with #91 because the parser for the latter stops at a different character (`$`). Maybe a different opcode name (e.g. `base64data=` and `data=`) to allow the parser to parse these items differently? @jpcima, your thoughts?

Missing or would-be-nice, possibly for further PRs:
- [x] Changes depending on the above discussion.
- [x] Rewrite the metadata reader to allow to read from memory instead of just files. The way to proceed is probably to have a new provision in the abstract `AudioReader` to read the raw data directly, regardless of the source. This part was not obvious when I refactored this so for the time being only basic metadata will be read mostly for WAV files. Once done, the `path` parameter added to the `AudioReader` constructors is unneeded.
- [ ] Expand the AIFF library with a memory version.
- [ ] Read gzipped/bzipped data, which would reduce a bit the size of the base64 fields.


paulfd:
I changed it to read the `base64data` field in the end, and updated the metadata reader, as well as adding a bunch of tests. Should be all good.

PR #1082, opened 2022-04-25
Oversampling and support for "cutoff_curveccN", "eg_shape", and "egN_freq_lfoX"

PythonBlue:
Reimplemented my pull request for this project at long last. Additional notes:

1. Had to rewrite a few of the test functions as well just for the sake of minimizing the build errors for sfizz_tests, even if those aren't required for the main project.

2. Partially due to note 1, I did my best at minimizing the changes I made in this particular branch for the sake of ease of merging if approved, meaning I ommitted my custom interpolation code otherwise available in my develop branch.

3. For the sake of my own needs, also increased the range of pitchMod under the Defaults.cpp file


src/sfizz/FilterDescription.h
```diff
@@ -15,6 +15,8 @@ namespace sfz
 struct FilterDescription
 {
     float cutoff { Default::filterCutoff };
```
paulfd, line 17: Unless I'm mistaken this is never set by an opcode. Is this on purpose?

src/sfizz/FilterPool.cpp
```diff
@@ -36,6 +36,17 @@ void sfz::FilterHolder::setup(const Region& region, unsigned filterId, int noteN
...
+
+    auto cutoffMod = description->cutoffMod;
+    for (const auto& mod : description->cutoffCC) {
+        const auto& curve = resources.getCurves().getCurve(mod.data.curve);
+        const float value = resources.getMidiState().getCCValue(mod.cc);
+        cutoffMod += curve.evalNormalized(value) * mod.data.modifier;
+    }
+
+    baseCutoff *= centsFactor(cutoffMod);
+    baseCutoff = Default::filterCutoff.bounds.clamp(baseCutoff);
```
paulfd, line 48: I think we can clamp once in the end, would you agree?

src/sfizz/Region.cpp
```diff
@@ -579,13 +579,30 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode, bool cleanOpcode)
             filters[filterIndex].resonance = opcode.read(Default::filterResonance);
         }
         break;
-    case_any_ccN("cutoff&"): // also cutoff_oncc&, cutoff_cc&, cutoff&_cc&
+    case hash("cutoff&_cc&"): case hash("cutoff&_oncc&"):
```
paulfd, line 582: What was the issue with the previous switch?

src/sfizz/Region.cpp
```diff
@@ -579,13 +579,30 @@ bool sfz::Region::parseOpcode(const Opcode& rawOpcode, bool cleanOpcode)
...
         {
-            const auto filterIndex = opcode.parameters.front() - 1;
+            const auto filterIndex = opcode.parameters.empty() ? 0 : (opcode.parameters.front() - 1);
```
paulfd, line 584: the parameters shouldn't be empty here, otherwise the case won't match.

paulfd (commented):
I'm sorry for yet another delay. I'll extract the cutoff CC changes first so I had questions on this, if you don't mind :)

src/sfizz/ADSREnvelope.cpp
```diff
@@ -35,7 +35,7 @@ Float ADSREnvelope::secondsToExpRate(Float timeInSeconds) const noexcept
         return Float(0.0);
 
     timeInSeconds = std::max(Float(25e-3), timeInSeconds);
-    return std::exp(Float(-9.0) / (timeInSeconds * sampleRate));
+    return 1.0 / (sampleRate * timeInSeconds);
```
paulfd, line 38: I'll deal with this part after the cutoff but what was the reason to change this formula?

src/sfizz/ADSREnvelope.cpp
```diff
@@ -113,13 +116,24 @@ void ADSREnvelope::getBlockInternal(absl::Span<Float> output) noexcept
...
+                attackCount = 0;
                 currentState = State::Attack;
+            }
             break;
         case State::Attack:
-            while (count < size && (currentValue += attackStep) < 1)
+            while (count < size && (currentValue) < 1)
+            {
+                if (attackShape <= 0)
+                    currentValue = start + (1 - start) * pow(attackCount, -attackShape + 1);
```
paulfd, line 131: `pow` is pretty hardcore in such a tight loop, I think we could at least extract it before the while.

src/sfizz/FilterDescription.h
PythonBlue: No worries about the delay, first of all.

Admittedly it's been quite a while since I visited this code, myself, so can't confirm or deny this particular code immediately, though knowing what I was trying to do, it was likely necessary the way I saw the code to make my workaround for CC-based cutoff modulation. Feel free to write up a more convenient alternative to this part if you need to.

src/sfizz/FilterPool.cpp
PythonBlue: If you're proposing we consoldate the lines into one, that could work. The main reason I wrote it like that was to make it more legible.

src/sfizz/Region.cpp
PythonBlue: Maybe I misread the macro definition for case_any_ccN, but my recollection is that the way I edited it means it won't get confused with "smoothcc" or "stepcc", which, as you probably know, is still unimplemented at this time.

src/sfizz/Region.cpp
PythonBlue: Fair, though I'm pretty sure that conditional statement was there in the first place as error prevention against the index subtraction later on in the same line, even if it might have been excessive. Omit if needed.

src/sfizz/ADSREnvelope.cpp
PythonBlue: It's part of the process for being able to customize the EG shapes. I realize it means overwriting what you originally had for this formula, and apologies for that. In retrospect, with the time, I could've put in the effort to see about reverting to that if the EG shape is unspecified.

src/sfizz/ADSREnvelope.cpp
PythonBlue: Hmm. Well, since it appears the pow function doesn't make use of any loop-specific variables, at least for this line, critique accepted.

PR #1088, opened 2022-05-31
Correct a bug with dynamic updates on offed EGs

### LINKED ISSUE #2: Missing notes
Not sure if it happens in the JACK client or the SFZ engine, but compared to other sfz players I notice a definitive lack of notes played when using a few SFZs I have around.

This is one such case https://musical-artifacts.com/artifacts/460

Is there a polyphony limit? If so, it seems very low.

paulfd:
Closes: #2 

@falkTX if you want to check that this indeed solves your problem.

### DIFF
--- a/src/sfizz/ADSREnvelope.cpp
+++ b/src/sfizz/ADSREnvelope.cpp
@@ -34,7 +34,7 @@ Float ADSREnvelope::secondsToExpRate(Float timeInSeconds) const noexcept
     if (timeInSeconds <= 0)
         return Float(0.0);
 
-    timeInSeconds = std::max(Float(25e-3), timeInSeconds);
+    timeInSeconds = std::max(Float(Default::offTime), timeInSeconds);
     return std::exp(Float(-9.0) / (timeInSeconds * sampleRate));
 };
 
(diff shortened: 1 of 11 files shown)


falkTX:
@paulfd note stealing is much improved now, quickly using notes in succession triggers things properly, this is great to see thanks!

falkTX:
black midi still makes sfizz go bad, but that might be asking too much haha :sweat_smile: 
can recommend trying that out, https://github.com/x42/midigen.lv2 is a quick way to test that with its included tetris black midi song

paulfd:
I've received the Dwarf, and indeed performance is not amazing 😓 I'll try to improve it when I find the time (and unpile everything else)

falkTX:
Note that on dwarf there is an issue where RT operation stop while allocating memory on unrelated processes.
I was unable to find the reason why, but it is very easy to trigger by running a memory testing tool while dwarf is doing its business. Memory allocations can induce xruns.
This does not happen on duo or duox...

So the performance issues might be related, due to sfizz reallocating things on its background thread.
One possible thing to try is a fixed pool of memory. It will trigger xruns when plugin is first loaded, but then should be stable.

And yeah, I tried many things for priorities and cpu scheduling stuff. My current theory is on memory not being setup in the best way by the bootloader.

PR #1100, opened 2022-07-14
Unclamp the cutoff and add OSC queries for the cutoff modifiers

paulfd:
@PythonBlue I believe the cutoff modifiers were already in actually, can you check? I added your other changes regarding that.


PythonBlue:
Admittedly I'm getting ready for an extended weekend vacation, so I personally may not be able to verify until after Sunday. What I can say, though: if the edits you did make do allow cutoff_curvecc133,  for instance, to work, then I'll trust your own implementation if time is of the essence.

PR #1103, opened 2022-07-20
Add LFO frequency as an EG target

paulfd:
@PythonBlue, if you can check.


PythonBlue:
This implementation on its own works for me! Tested with the VST3 build.

paulfd:
Great, thanks!

PR #1111, opened 2022-07-29
Vstgui update

paulfd:


redtide:
I wonder if we could just keep the upstream vanilla vstgui instead a patched one, and then save patches to apply on the fly temporarily when building. Some distro maintainer might also benefit by using them.
What do you think?

redtide:
VSTGUI & VST3SDK are at v3.7.7, closing.

PR #1124, opened 2022-08-24
Add section suffix to man pages

iv-m:
`man`, at least on some linux distros, ignores files that don't have have the numeric suffix that corresponds to the man page section. This commit adds the suffix to the tools man pages.

With this change applied, `man sfizz_jack` actually works for me.


paulfd:
Thanks!

PR #1129, opened 2022-09-29
Add WavPack audio files support

lucianoiam:
[WavPack](https://github.com/dbry/WavPack) is a lossless audio codec


external/st_audiofile/src/st_audiofile.c
```diff
@@ -449,6 +496,25 @@ uint64_t st_read_s16(st_audio_file* af, int16_t* buffer, uint64_t count)
     case st_audio_file_mp3:
         count = drmp3_read_pcm_frames_s16(af->mp3, count, buffer);
         break;
+    case st_audio_file_wv:
+        {
+            uint32_t channels = af->cache.wv.channels;
+            int32_t* buf_i32 = (int32_t*)malloc(4 * channels * count);
```
paulfd, line 504: May I change this for some C++-y lifetime management version?

lucianoiam: Yes sure, that would be great.

paulfd: Nevermind I forgot this is a C file.

external/st_audiofile/src/st_audiofile.h
```diff
@@ -19,6 +19,8 @@
 #include <wchar.h>
 #endif
 
+#define WAVPACK_MEMORY_ASSUMED_VERSION 5
```
paulfd: Does this need to be in the header file?

external/st_audiofile/src/st_audiofile.c
```diff
@@ -478,6 +544,28 @@ uint64_t st_read_f32(st_audio_file* af, float* buffer, uint64_t count)
...
+            if (!buf_i32) {
+                return 0;
+            }
+            count = channels * WavpackUnpackSamples(af->wv, buf_i32, (uint32_t)count);
+            if (!(af->cache.wv.mode & MODE_FLOAT)) {
+                if (af->cache.wv.bitrate < 32) {
+                    int d = 32 - af->cache.wv.bitrate;
+                    for (uint64_t i = 0; i < count; i++) {
+                        buf_i32[i] <<= d;
```
paulfd, line 563: So basically in wavpack if you use integers less than 32 bit length, you're not gaining any space?

paulfd (approved):
I rebased and added a test (and force-pushed on your branch).

external/st_audiofile/src/st_audiofile.h
paulfd: I moved it for now, will revert if it becomes an issue.

lucianoiam: No reason to have it in the header file, https://github.com/sfztools/sfizz/pull/1129/commits/c3c416496f26b41b27c82d8e1adf750e4e24ed27 looks good.

external/st_audiofile/src/st_audiofile.c
lucianoiam: Acoording to a [comment](https://github.com/dbry/WavPack/blob/ffc72be679dcb13b7e6004060c4e7baf5ba24401/src/unpack_utils.c#L94) in `WavpackUnpackSamples()`, the function always returns 32-bit data, padded when needed. So optimizing sfizz to use less memory when dealing with < 32 bit WV audio does not seem possible.

paulfd: OK np, in memory we store as float anyway. It was just out of curiosity.

PR #1154, opened 2023-04-09
Abseil update LTS 20220623 to LTS 20230125.2

redtide:
Draft PR because it seems latest Abseil causes some drawbacks:

- Requires to raise C++ standard from 11 to 14 in `SfizzConfig.cmake` and minimum CMake version to 3.10 in main CMakeLists.txt
- Breaks MOD build (see Discord chat)
- in C++14 mode (when not building the VST3 which forces C++17) spits a lot of errors about C++17 classes like `string_view` not being in namespace `absl` and/or `std`

this needs some investigation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -210,6 +210,7 @@ jobs:
           ${{runner.workspace}}/build/${{env.install_name}}.pkg
 
   build_for_mod:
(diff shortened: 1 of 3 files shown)


redtide:
Going to make a PR for 20230125.3 instead.

PR #1171, opened 2023-05-28
Remove CXX standard switches on CI to use the default config

paulfd:
@redtide


redtide:
Oops, keeping the CI config in "sync" with the other repo I forgot that this is not needed here :P

PR #1173, opened 2023-06-09
Update Region.cpp to support per-voice CC's

PythonBlue:
Figured out my issue of why I had to deviate so much with filter cutoff CC curves: it turns out the file I changed originally assumed all CC's are per-cycle controllers, whereas most of the extended CC's, including the ones I was working with, are per-voice controllers, so I added conditional statements for if it's one of those extended CC numbers. Should positively affect other targets that make use of complex CC modulation as well.

### DIFF
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -1692,7 +1692,7 @@ bool sfz::Region::processGenericCc(const Opcode& opcode, OpcodeSpec<float> spec,
         auto it = std::find_if(connections.begin(), connections.end(),
             [ccNumber, &target](const Connection& x) -> bool
             {
-                if ((ccNumber > 13 && ccNumber < 138) || ccNumber == 140 || ccNumber == 141)
+                if ((ccNumber > 130 && ccNumber < 138) || ccNumber == 140 || ccNumber == 141)
                     return x.source.id() == ModId::PerVoiceController &&
                         x.source.parameters().cc == ccNumber &&
                         x.target == target;
--- a/src/sfizz/Region.cpp
+++ b/src/sfizz/Region.cpp
@@ -1690,10 +1690,11 @@ bool sfz::Region::processGenericCc(const Opcode& opcode, OpcodeSpec<float> spec,
         // search an existing connection of same CC number and target
         // if it exists, modify, otherwise create
         auto it = std::find_if(connections.begin(), connections.end(),
-            [ccNumber, &target](const Connection& x) -> bool
+            [ccNumber, &target, this](const Connection& x) -> bool
             {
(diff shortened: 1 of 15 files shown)


src/sfizz/Region.cpp
```diff
@@ -1692,6 +1692,10 @@ bool sfz::Region::processGenericCc(const Opcode& opcode, OpcodeSpec<float> spec,
         auto it = std::find_if(connections.begin(), connections.end(),
             [ccNumber, &target](const Connection& x) -> bool
             {
+                if ((ccNumber > 130 && ccNumber < 138) || ccNumber == 140 || ccNumber == 141)
```
paulfd: If we keep it like this maybe it would make sense to add an `isPerVoiceController` function to the extended CC list to replace both this check and the switch below in this method. However we can plan to do that after.

src/sfizz/Region.cpp
```diff
@@ -1692,6 +1692,10 @@ bool sfz::Region::processGenericCc(const Opcode& opcode, OpcodeSpec<float> spec,
...
+                    return x.source.id() == ModId::PerVoiceController &&
+                        x.source.parameters().cc == ccNumber &&
```
paulfd, line 1698: Shouldn't we check the region id here too in this case? If not, I'd suggest just modifying the initial check to `x.source.id() == ModId::PerVoiceController || x.source.id() == ModId::Controller`.

src/sfizz/Region.cpp
```diff
@@ -1851,7 +1855,12 @@ sfz::Region::Connection& sfz::Region::getOrCreateConnection(const ModKey& source
 sfz::Region::Connection* sfz::Region::getConnectionFromCC(int sourceCC, const ModKey& target)
 {
     for (sfz::Region::Connection& conn : connections) {
-        if (conn.source.id() == sfz::ModId::Controller && conn.target == target) {
+        if (conn.source.id() == sfz::ModId::PerVoiceController && conn.target == target) {
```
paulfd: Same as above, it feels like this could be a single `if` statement.

paulfd (commented):
Hey, thanks a lot. Are there any examples of this and the desired behavior? Especially something we could test would be amazing...

src/sfizz/Region.cpp
PythonBlue: Thanks for the critiques, Paul. Insofar as this code and the one with the corresponding suggestion, good thought about the region checking, I suppose. Beyond that technicality with this particular one, I'll let you be the judge over how to format the syntax exactly: I fully admit code optimization is not my forte.

Sorry for my delayed response, btw.

Issue #1077, opened 2022-02-15
Disappeared large number CC


kinwie:
I encountered a weird issue.
I'm using large CC numbers for an instrument.
Two CCs are missing : _Pan Tom1 (CC 453)_ and _Tune Tom2 (CC 474)_, in red circle :
![disappears](https://user-images.githubusercontent.com/65444898/154096025-f8e171e8-05b7-466d-975a-f7496519e44d.png)

But if I change them to small numbers, they appeared :
![appears](https://user-images.githubusercontent.com/65444898/154096085-14546083-343a-44a3-92ca-b15f1f7971d0.png)

And if I only load only the Tom1 or Tom2 sfz file, the large CCs also appeared.

paulfd:
This is unexpected. Can you share the files?

kinwie:
Here's the files :
[test-master.zip](https://github.com/sfztools/sfizz/files/8077456/test-master.zip)

The files came from our private repo. It's a prototype Drumkit sample library project, so it contains a lot of sfz presets.
But basically you just load the "NewDrums.sfz" in the 3rd folder, the Stereo-Kit version to check.

PR #1181, opened 2023-08-08
Add a test for #1077

### LINKED ISSUE #96: Disappeared large number CC
I encountered a weird issue.
I'm using large CC numbers for an instrument.
Two CCs are missing : _Pan Tom1 (CC 453)_ and _Tune Tom2 (CC 474)_, in red circle :
disappears

But if I change them to small numbers, they appeared :
appears

And if I only load only the Tom1 or Tom2 sfz file, the large CCs also appeared.

paulfd:
Closes: sfztools/sfizz-ui#96 

I think this is a UI problem, we can move this issue @redtide


redtide:
At this point I would move it after merge, as you prefer.

PR #1183, opened 2023-08-10
modifications to support univeral x86_64/ARM builds on Mac

essej:
These changes were necessary to build an Intel/Arm universal library on macOS (some simd-related build errors were occurring). Untested on other platforms, but it *should* work. I just loosened the match regex to catch the compound architecture used when specifying  `set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64" CACHE INTERNAL "")` in a parent CMakeLists (or via command line).


paulfd:
Thanks!

PR #1186, opened 2023-08-25
Add support for opcodes lotimer/hitimer

essej:
While not widely supported or well documented, is very useful especially when triggering a note with on_locc/on_hicc to prevent a bunch of retriggers.

The region will trigger as long as the elapsed time from the note-on of the most recently triggered region in the current polyphony group is between lotimer and hitimer amount of seconds. lotimer default 0, hitimer default is max_float.

Example use: to preventing a region from retriggering too quickly, set lotimer=0.1

### DIFF
--- a/src/sfizz/Defaults.cpp
+++ b/src/sfizz/Defaults.cpp
@@ -199,6 +199,8 @@ FloatSpec lofiDecim { 0.0f, {0.0f, 100.0f}, 0 };
 FloatSpec rectify { 0.0f, {0.0f, 100.0f}, 0 };
 UInt32Spec stringsNumber { maxStrings, {0, maxStrings}, 0 };
 BoolSpec sustainCancelsRelease { false, {0, 1}, kEnforceBounds };
+FloatSpec loTimer { 0.0f, {0.0f, float_max}, 0 };
+FloatSpec hiTimer { float_max, {0.0f, float_max}, 0 };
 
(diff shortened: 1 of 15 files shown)


paulfd:
Thanks for the MR! Same as the other, I'd like to add a some form of testing for this, do you have the typical use case that triggered you to make this change?

essej:
Sure, as mentioned in the comment above, my motivating use case was a way to rate-limit triggers from a region set up to be triggered via CC events using on_locc/on_hicc . For example:
```ini
<region>
group=1
on_locc4=0 on_hicc4=0
lotimer=0.1
sample=hihatFoot_OH_MP_1.wav
```

Which is an example where you could have a pedal which is setup in an expression input that sends CC values, to trigger when the value hits 0, but to not trigger again until after 0.1 seconds elapses, and effectively debounces the input which may jitter around zero.

I saw this unimplemented opcode in the spec, which did exactly what I wanted, assuming I decoded the meaning correctly, since it appears to be unimplemented by most other SFZ engines.

I haven't looked into how the tests are setup, should I perhaps add something in RegionTriggersT.cpp to test this opcode, or in some other place?

paulfd:
The implem looks great. I'll add some parsing and messaging tests, and for this one we can test triggering too.

paulfd:
I added the tests, tell me if everything looks reasonable!

PR #1194, opened 2023-10-03
CC Curve support for standard EG's

PythonBlue:
Implemented a crude approach to supporting CC curves by adding the CurveSet resources to the EG's.

The following opcodes should now be supported with this branch:

ampeg_attack_onccN
ampeg_decay_onccN
ampeg_delay_onccN
ampeg_hold_onccN
ampeg_release_onccN
ampeg_sustain_onccN
fileg_attack_onccN
fileg_decay_onccN
fileg_delay_onccN
fileg_hold_onccN
fileg_elease_onccN
fileg_sustain_onccN
pitcheg_attack_onccN
pitcheg_decay_onccN
pitcheg_delay_onccN
pitcheg_hold_onccN
pitcheg_release_onccN
pitcheg_sustain_onccN

Also, apologies for all the constant re-requests: main issue was my failing to realize I copied all of the files I should have for the branch.


paulfd:
That's a great feature thanks, and the implem looks good at a glance. I'll expand the existing tests to account for this capability. Is this actually in the spec?

PythonBlue:
All of these opcodes are already implemented in ARIA, though your question raises an interesting point that only some of them are listed at all on the sfizz website's opcode page (the one for ampeg attack is missing, for instance?).

redtide:
Maybe better here than on Discord:
some opcodes might be missing on sfzformat: the opcode list on sfz.tools uses the "same" data as sfzformat (a copy updated time to time), so if an opcode is missing is just not included in the sfzformat database. Other than that might be that some opcodes marked as unsupported are already implemented, just not keep updated in the sfz.tools database.
So if someone find some error please report in the appropriated website issue tracker, [sfzformat](https://github.com/sfzformat/sfzformat.github.io/issues) or [sfztools](https://github.com/sfztools/sfztools.github.io/issues).

src/sfizz/EGDescription.h
```diff
@@ -66,13 +68,20 @@ struct EGDescription {
...
-    CCMap<float> ccRelease;
+    CCMap<ModifierCurvePair<float>> ccAttack { ModifierCurvePair<float>{ Default::egTime, Default::curveCC } };
+    CCMap<ModifierCurvePair<float>> ccDecay { ModifierCurvePair<float>{ Default::egTime, Default::curveCC } };
+    CCMap<ModifierCurvePair<float>> ccDelay { ModifierCurvePair<float>{ Default::egTime, Default::curveCC } };
+    CCMap<ModifierCurvePair<float>> ccHold { ModifierCurvePair<float>{ Default::egTime, Default::curveCC } };
+    CCMap<ModifierCurvePair<float>> ccRelease { ModifierCurvePair<float>{ Default::egTime, Default::curveCC } };
     CCMap<float> ccStart;
-    CCMap<float> ccSustain;
+    CCMap<ModifierCurvePair<float>> ccSustain { ModifierCurvePair<float>{ Default::egPercent, Default::curveCC } };
+    //CCMap<float> ccAttack;
```
paulfd, line 78: These could probably go away.

paulfd:
Hey @PythonBlue, is this branch in good shape? I can probably fix the tests and look it over shortly if so!

PythonBlue:
Thanks for the reminder about this PR. I THINK I just fixed the tests issue, though I guess we'll find out in a few minutes, but either way, it should be better this way anyway because I realized part of the issue was how the CC Sustain and Start vaues were made.

src/sfizz/EGDescription.h
PythonBlue: If you're referring to what I commented out, fair point. I just kept it in case revertng was desired.

Issue #1086, opened 2022-05-22
sfizz output level too low?


cvde:
I noticed that the output level of a sample played via sfizz is ~3dB lower compared to playing the sample directly. In sforzando it's exactly the same. I wonder if I'm missing something in my sfz file or if this behavior is intentional.

Here is a (minimal) sfz file:
```
<control>
set_cc7=127
<global>
loop_mode=one_shot
<region> sample=sample.wav key=49
```
The attached sample.wav file peaks at -1dB when played directly. It peaks at -4dB when played via sfizz.

[OutputLevelTest.zip](https://github.com/sfztools/sfizz/files/8749117/OutputLevelTest.zip)

PR #1198, opened 2023-10-29
Adjusted the gain compensation to not lose gain due to applied pan stages

cvde:
This fixes the issue (https://github.com/sfztools/sfizz/issues/1086) that the output level of a sample played through sfizz is ~3dB lower compared to playing the sample directly.


paulfd:
Thanks for the MR!

PR #1202, opened 2023-11-01
Fix a bug that st_read_s16 and st_read_f32 reterns the wrong number of frames

### LINKED ISSUE #1201: The wrong number of frames is returned by the st_audiofile function
st_read_s16 and st_read_f32 returns the number of frames multiplied by the number of the WavPack file's channels.

KKQ-KKQ:
fix #1201

### DIFF
--- a/external/st_audiofile/src/st_audiofile.c
+++ b/external/st_audiofile/src/st_audiofile.c
@@ -516,12 +516,13 @@ uint64_t st_read_s16(st_audio_file* af, int16_t* buffer, uint64_t count)
             if (!buf_i32) {
                 return 0;
             }
-            count = channels * WavpackUnpackSamples(af->wv, buf_i32, (uint32_t)count);
+            count = WavpackUnpackSamples(af->wv, buf_i32, (uint32_t)count);
+            uint64_t buf_size = channels * count;
             if (af->cache.wv.mode & MODE_FLOAT) {
-                drwav_f32_to_s16((drwav_int16*)buffer, (float*)buf_i32, (size_t)count);
+                drwav_f32_to_s16((drwav_int16*)buffer, (float*)buf_i32, (size_t)buf_size);
             } else {
                 int d = af->cache.wv.bitrate - 16;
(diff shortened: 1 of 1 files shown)


external/st_audiofile/src/st_audiofile.c
```diff
@@ -566,15 +567,16 @@ uint64_t st_read_f32(st_audio_file* af, float* buffer, uint64_t count)
             if (!buf_i32) {
                 return 0;
             }
-            count = channels * WavpackUnpackSamples(af->wv, buf_i32, (uint32_t)count);
-            if (!(af->cache.wv.mode & MODE_FLOAT)) {
+            count = WavpackUnpackSamples(af->wv, buf_i32, (uint32_t)count);
+            {
```
paulfd, line 571: I guess from the code above the if condition was always true. If so, we can remove the brackets too.

external/st_audiofile/src/st_audiofile.c
```diff
@@ -566,15 +567,16 @@ uint64_t st_read_f32(st_audio_file* af, float* buffer, uint64_t count)
             if (!buf_i32) {
```
paulfd, line 567: For the float case a couple lines above, would we need to consider channel count too?

paulfd (commented):
Thanks for the MR! I just had a couple comments.

external/st_audiofile/src/st_audiofile.c
KKQ-KKQ: I think the count value was multiplied in order to use for loop.
So we don't have to consider the channel for the float case.

external/st_audiofile/src/st_audiofile.c
KKQ-KKQ: Yes, we can.

external/st_audiofile/src/st_audiofile.c
KKQ-KKQ: I think we needs it for drwav_f32_to_s16 function.
The function simply converts f32 to s16. so we have to consider the channel count.

PR #1224, opened 2023-12-26
Use the standard opcode value read for `hint_ram_based`

### LINKED ISSUE #1187: hint_ram_based boolean?
[hint_ram_based](https://github.com/sfztools/sfizz/blob/develop/src/sfizz/Synth.cpp#L441) looks like an exception compared to other boolean opcode types.
(as a side note, [readBoolean](https://github.com/sfztools/sfizz/blob/develop/src/sfizz/Opcode.cpp#L317) has a TODO that might be removed).

paulfd:
@redtide 
Closes: #1187


redtide:
Thank you!

Sylv1o:
Hi, does it change something in the syntaxe to be used ?
hint_ram_based=1 still OK ?

Thanks

redtide:
@Sylv1o: It's OK. Before this `hint_ram_based` used only integer ARIA style boolean, now should use [readBoolean](https://github.com/sfztools/sfizz/blob/develop/src/sfizz/Opcode.cpp#L307-L325) like other boolean opcodes, that supports also CakeWalk' "on/off" (for consistence).

Sylv1o:
Great, thanks for clarifying.

PR #1229, opened 2024-01-15
Always build universal binaries on macOS

paulfd:


redtide:
I think this is not necessary here, I thought we could make st_audiofile a standalone repo with its own CI workflow and do this there instead.

paulfd:
Doing both is also an option!

redtide:
I still have the same idea about not to "abuse" of CI resources (after the Travis experience...). Other than that, builds are quite a lot and takes forever if there is some activity on the repo.
So if you care to set the universal build here, then forget my previous comment.

redtide:
Anyway.. this is strange, disabling `WAVPACK_ENABLE_ASM` was working before, unless it's something different this time. Is it possible that it needs to cache the var in CMake?

redtide:
Erm... my bad, I forgot about [this](https://github.com/sfztools/sfizz-ui/blob/develop/.github/workflows/build.yml#L174), and also confused for some reason that the issue was on the plugins side.

PR #1232, opened 2024-02-01
Improved Note-On performance

KKQ-KKQ:
I found that calculating a hash of the ModKey is the bottleneck.
I added a variable to cache ModKey's hash.

### DIFF
--- a/src/sfizz/utility/StringViewHelpers.h
+++ b/src/sfizz/utility/StringViewHelpers.h
@@ -101,7 +101,12 @@ uint64_t hashNumber(Int i, uint64_t h = Fnv1aBasis)
 {
     static_assert(std::is_arithmetic<Int>::value,
                   "The hashed object must be of arithmetic type");
-    return hash(absl::string_view(reinterpret_cast<const char*>(&i), sizeof(i)), h);
+    union {
+        uint64_t u64;
+        Int i;
+    } un {};
+    un.i = i;
+    return (h ^ un.u64) * Fnv1aPrime;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


KKQ-KKQ:
I found ModKey hash function doesn't process its region.
I added the process.

src/sfizz/utility/StringViewHelpers.h
```diff
@@ -101,7 +101,12 @@ uint64_t hashNumber(Int i, uint64_t h = Fnv1aBasis)
 {
     static_assert(std::is_arithmetic<Int>::value,
                   "The hashed object must be of arithmetic type");
-    return hash(absl::string_view(reinterpret_cast<const char*>(&i), sizeof(i)), h);
+    union {
```
paulfd, line 104: Why was this needed? Some alignment issue?

paulfd:
Thanks a lot for the PR and for catching this! I will take some time to review it better hopefully this WE. I had another issue a long time ago where I also needed to cache the key/id hashes for lookups in the modulation matrix...

src/sfizz/utility/StringViewHelpers.h
KKQ-KKQ: This block is not needed but this results in the better performance.
I changed this block because I found hashing process is costing a lot at first. but finally I found it is not the core issue.
The core issue is that the hash of ModKey region is not being calculated.

paulfd: I see thanks

paulfd:
Would it be okay that I reformat and squash the commits a bit to match the usual code style? I'll keep your authorship.

KKQ-KKQ:
No problem.
I am very new to co-working and I’m still getting used to using git.

paulfd:
No worries, thanks a lot! I mostly put all commits into a single one (which is called "squashing") to make it easier to follow the history.

PR #1235, opened 2024-02-04
C++20 Support

paulfd:
From @KKQ-KKQ

### DIFF
--- a/clients/sfizz_render.cpp
+++ b/clients/sfizz_render.cpp
@@ -3,6 +3,7 @@
 #include "sfizz/MathHelpers.h"
 #include "sfizz/SfzHelpers.h"
 #include "sfizz/SIMDHelpers.h"
+#include "sfizz/utility/U8Strings.h"
 #include "MidiHelpers.h"
 #include <st_audiofile_libs.h>
 #include <cxxopts.hpp>
@@ -165,7 +166,7 @@ int main(int argc, char** argv)
     ERROR_IF(!synth.loadSfzFile(sfzPath), "There was an error loading the SFZ file.");
     LOG_INFO(synth.getNumRegions() << " regions in the SFZ.");
 
-    fmidi_smf_u midiFile { fmidi_smf_file_read(midiPath.u8string().c_str()) };
+    fmidi_smf_u midiFile { fmidi_smf_file_read(u8EncodedString(midiPath).c_str()) };
     ERROR_IF(!midiFile, "Can't read " << midiPath);
 
     const auto* midiInfo = fmidi_smf_get_info(midiFile.get());
--- a/src/sfizz/Synth.cpp
+++ b/src/sfizz/Synth.cpp
@@ -8,6 +8,7 @@
 #include "Config.h"
 #include "utility/Debug.h"
 #include "utility/Macros.h"
+#include "utility/U8Strings.h"
 #include "modulations/ModId.h"
 #include "modulations/ModKey.h"
 #include "modulations/ModMatrix.h"
@@ -702,7 +703,7 @@ void Synth::Impl::finalizeSfzLoad()
     filePool.setRootDirectory(rootDirectory);
 
     // a string representation used for OSC purposes
-    rootPath_ = rootDirectory.u8string();
+    rootPath_ = u8EncodedString(rootDirectory);
 
     size_t currentRegionIndex = 0;
     size_t currentRegionCount = layers_.size();
--- a/external/threadpool/ThreadPool.h
+++ b/external/threadpool/ThreadPool.h
@@ -18,7 +18,11 @@ public:
     ThreadPool(size_t);
     template<class F, class... Args>
     auto enqueue(F&& f, Args&&... args)
+#if __cplusplus >= 201703L
+        -> std::future<typename std::invoke_result<F, Args...>::type>;
+#else
         -> std::future<typename std::result_of<F(Args...)>::type>;
+#endif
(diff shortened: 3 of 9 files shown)


paulfd:
@KKQ-KKQ if you can check that this build with your toolchain thatd be great. I made a couple changes to reduce the number of ifdef and fix some other issues that popped up.

clients/sfizz_render.cpp
```diff
@@ -165,7 +166,7 @@ int main(int argc, char** argv)
     ERROR_IF(!synth.loadSfzFile(sfzPath), "There was an error loading the SFZ file.");
     LOG_INFO(synth.getNumRegions() << " regions in the SFZ.");
 
-    fmidi_smf_u midiFile { fmidi_smf_file_read(midiPath.u8string().c_str()) };
+    fmidi_smf_u midiFile { fmidi_smf_file_read(from_u8string(midiPath.u8string()).c_str()) };
```
KKQ-KKQ: Does this works in Windows environment with non-ASCII characters?
I cannot confirm it because I am with Mac.
Windows uses ANSI, UTF-16, and UTF-8.
ANSI and UTF-8 are different.

If you don't want to support non-ASCII characters, it might be Ok.

src/sfizz/Synth.cpp
```diff
@@ -702,7 +703,7 @@ void Synth::Impl::finalizeSfzLoad()
     filePool.setRootDirectory(rootDirectory);
 
     // a string representation used for OSC purposes
-    rootPath_ = rootDirectory.u8string();
+    rootPath_ = from_u8string(rootDirectory.u8string());
```
KKQ-KKQ, line 707: ~fs::path::string() and fs::path::u8string() use different character codes.
They are incompatible.~
Oh I'm sorry, it's my code, and rootPath is a UTF-8 std::string, so it might be ok.

KKQ-KKQ (commented):
Sorry for my late review.
I commented some days ago but the comments remained "Pending".

I don't think there is always compatibility between `std::string` and `std::u8string` especially in file paths.

KKQ-KKQ:
It might be better that we add functions like `pathToString`, `stringToPath`, `pathToU8String` and `u8StringToPath`.
On Windows, string means ANSI, u8String means UTF-8, wstring means UTF-16 as far as I know.
but std::u8string is sometimes hard to use, so we use std::string as a UTF-8 string container.
It might be better that we should name the string variables with a prefix/postfix 'u8'.

clients/sfizz_render.cpp
KKQ-KKQ: `fmidi_smf_file_read(const char *)` ~uses ANSI, as far as I confirmed.~

paulfd: Hmm are you sure? From [this](https://github.com/sfztools/sfizz/blob/cc33ed160f49ba49ada5d74c9137a634c15fbb2d/external/fmidi/sources/fmidi/fmidi_mini.cpp#L1182) it seems that the POSIX version would use fopen (which will be UTF-8 on linux afaik, not sure about macOS) and has some conversion from utf8 to utf16 to open on Windows. At least running this patch on windows for `sfizz_render` works properly for me.

Did you have any failing scenario in mind?

paulfd:
No worries. We can add some functions if needed, but after reading https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r2.html I felt their proposal was reasonable.

paulfd:
If this patch works for you and your toolchain, we can solve problems as they arise and possibly add functions at that time!

clients/sfizz_render.cpp
KKQ-KKQ: oh I misunderstood. I was wrong.
I'm sorry.
It use UTF-8 even on Windows.

KKQ-KKQ: I read the code on my mac with the vs code and the wrong `#if` block was grayed, I misunderstood.
Sorry for confusion.

KKQ-KKQ: Do we need this change?
In my opinion, it is a little bit complex process.

KKQ-KKQ:
It is hard for me to apply this patch into my local mod without changes because I don't manage my mod with git.
I only checked if this branch works with sfizz-ui and actually it works.

clients/sfizz_render.cpp
```diff
@@ -165,7 +166,7 @@ int main(int argc, char** argv)
     ERROR_IF(!synth.loadSfzFile(sfzPath), "There was an error loading the SFZ file.");
     LOG_INFO(synth.getNumRegions() << " regions in the SFZ.");
 
-    fmidi_smf_u midiFile { fmidi_smf_file_read(midiPath.u8string().c_str()) };
+    fmidi_smf_u midiFile { fmidi_smf_file_read(u8EncodedString(midiPath).c_str()) };
```
KKQ-KKQ, line 169: The process becomes more complex.
expanded process:
```
const std::u8string s = midiPath.u8string();
fmidi_smf_file_read(std::string(s.begin(), s.end()).c_str())
```
In my opinion, the conversion to std::string is waste.
I prefer just `midiPath.u8string().c_str()` but the choice is yours.

paulfd: Yeah, except we at least need to `reinterpret_cast` it to `const char*`. This would avoid one copy here but it's a minor offense, and it makes this method ubiquitous across the codebase where most other conversions are to owning `std::string`.

PR #1236, opened 2024-02-05
Shared FilePool

KKQ-KKQ:
In iOS/iPadOS, an AU plugin's process is shared.
In order to reduce its memory, I made this changes.

I fixed some issues but I cannot reopen the PR because I squashed commits.

### DIFF
--- a/src/sfizz/AudioReader.cpp
+++ b/src/sfizz/AudioReader.cpp
@@ -16,7 +16,7 @@
 namespace sfz {
 
 #if defined(SFIZZ_USE_SNDFILE)
-static bool formatHasFastSeeking(int format)
+static bool formatHasFastSeeking(int format);
 #endif
 
 class BasicSndfileReader : public AudioReader {
--- a/src/sfizz/AudioReader.cpp
+++ b/src/sfizz/AudioReader.cpp
@@ -15,6 +15,10 @@
 
 namespace sfz {
 
+#if defined(SFIZZ_USE_SNDFILE)
+static bool formatHasFastSeeking(int format)
+#endif
+
 class BasicSndfileReader : public AudioReader {
 public:
     explicit BasicSndfileReader(ST_AudioFile handle, std::unique_ptr<MetadataReader> mdReader)
@@ -25,6 +29,8 @@ public:
     int64_t frames() const override;
     unsigned channels() const override;
     unsigned sampleRate() const override;
+    virtual bool seekable() const override;
+    virtual bool seek(uint64_t position) override;
     bool getInstrumentInfo(InstrumentInfo& instrument) override;
     bool getWavetableInfo(WavetableInfo& instrument) override;
 protected:
@@ -50,6 +56,20 @@ unsigned BasicSndfileReader::channels() const
 unsigned BasicSndfileReader::sampleRate() const
 {
     return handle_.get_sample_rate();
+}
+
+bool BasicSndfileReader::seekable() const
+{
+#if defined(SFIZZ_USE_SNDFILE)
+    return formatHasFastSeeking(handle_.get_sndfile_format());
+#else
+    return true;
+#endif
+}
+
+bool BasicSndfileReader::seek(uint64_t position)
+{
+    return handle_.seek(position);
 }
 
 bool BasicSndfileReader::getWavetableInfo(WavetableInfo& wt)
@@ -165,6 +185,7 @@ public:
     explicit ReverseReader(ST_AudioFile handle, std::unique_ptr<MetadataReader> mdReader);
     AudioReaderType type() const override;
     size_t readNextBlock(float* buffer, size_t frames) override;
+    bool seek(uint64_t position) override;
 
 private:
     uint64_t position_ {};
@@ -200,6 +221,16 @@ size_t ReverseReader::readNextBlock(float* buffer, size_t frames)
     return readFrames;
 }
 
+bool ReverseReader::seek(uint64_t position)
+{
+    uint64_t frames = handle_.get_frame_count();
+    if (position <= frames) {
+        position_ = frames - position;
+        return true;
+    }
+    return false;
+}
+
 //------------------------------------------------------------------------------
 
 /**
@@ -210,6 +241,8 @@ public:
     explicit NoSeekReverseReader(ST_AudioFile handle, std::unique_ptr<MetadataReader> mdReader);
     AudioReaderType type() const override;
     size_t readNextBlock(float* buffer, size_t frames) override;
+    bool seekable() const override { return false; }
+    bool seek(uint64_t) override { return false; }
 
 private:
     void readWholeFile();
@@ -316,6 +349,8 @@ public:
     unsigned channels() const override { return 1; }
     unsigned sampleRate() const override { return 44100; }
     size_t readNextBlock(float*, size_t) override { return 0; }
+    bool seekable() const override { return false; }
+    bool seek(uint64_t) override { return false; }
     bool getInstrumentInfo(InstrumentInfo& ) override { return false; }
 private:
     AudioReaderType type_ {};
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -47,8 +47,11 @@
 using namespace std::placeholders;
 
 sfz::FilePool::GlobalObject::GlobalObject(size_t num_threads)
-: threadPool { new ThreadPool(num_threads) }
-{}
+: lastGarbageCollection_ { std::chrono::high_resolution_clock::now() },
+  garbageThread { new std::thread( &GlobalObject::garbageJob, this ) },
+  threadPool { new ThreadPool(num_threads) }
+{
+}
 
 sfz::FilePool::GlobalObject::~GlobalObject()
 {
@@ -56,7 +59,7 @@ sfz::FilePool::GlobalObject::~GlobalObject()
     std::error_code ec;
     semGarbageBarrier.post(ec);
     ASSERT(!ec);
-    garbageThread.join();
+    garbageThread->join();
 }
 
 RTSemaphore sfz::FilePool::GlobalObject::semGarbageBarrier { 0 };
@@ -820,15 +823,22 @@ void sfz::FilePool::stopRender()
 
 void sfz::FilePool::GlobalObject::garbageJob()
 {
-    while (semGarbageBarrier.timed_wait(500), garbageFlag) {
+    while (semGarbageBarrier.timed_wait(sfz::config::fileClearingPeriod * 1000), garbageFlag) {
         if (runningRender != 0) {
             continue;
         }
+        const auto now = std::chrono::high_resolution_clock::now();
+        const auto timeSinceLastCollection =
+        std::chrono::duration_cast<std::chrono::seconds>(now - lastGarbageCollection_);
+
+        if (timeSinceLastCollection.count() < config::fileClearingPeriod) {
+            continue;
+        }
+        lastGarbageCollection_ = now;
 
         {
             std::unique_lock<std::mutex> guard { preloadedFilesMutex, std::try_to_lock };
             if (guard.owns_lock()) {
-                const auto now = std::chrono::high_resolution_clock::now();
                 for (auto it = preloadedFiles.begin(); it != preloadedFiles.end();) {
                     auto copyIt = it++;
                     auto& data = copyIt->second;
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -848,7 +848,7 @@ void sfz::FilePool::GlobalObject::garbageJob()
                         continue;
                     }
 
-                    if (data->availableFrames == 0)
+                    if (data->availableFrames == 0 || data->readerCount != 0 || data->dataReaderCount != 0)
                         continue;
 
                     switch (FileData::Status(data->status)) {
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -127,7 +127,9 @@ bool sfz::FileData::addSecondaryOwner(FilePool* owner)
 {
     {
         std::unique_lock<std::mutex> guard { readyMutex };
-        readyCond.wait(guard, [this]{ return ready; });
+        if (!readyCond.wait_for(guard, std::chrono::seconds(10), [this]{ return ready; })) {
+            return false;
+        }
     }
 
     std::lock_guard<std::mutex> guard2 { ownerMutex };
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -853,11 +853,12 @@ void sfz::FilePool::GlobalObject::garbageJob()
                     if (secondsIdle < config::fileClearingPeriod)
                         continue;
 
+                    FileAudioBuffer garbage;
                     std::unique_lock<SpinMutex> guard2 { data->garbageMutex, std::try_to_lock };
                     if (guard2.owns_lock() && data->readerCount == 0 && data->dataReaderCount == 0) {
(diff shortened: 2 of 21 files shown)


KKQ-KKQ:
I added a stress test and it crashes.

KKQ-KKQ:
@paulfd The running action hangs! Please stop it.

redtide:
I thought it stopped by itself, not running for more than one hour... I canceled it.

KKQ-KKQ:
@redtide @paulfd  I'm sorry but could you stop actions? I will set timeout next time.

paulfd:
Should be stopped!

KKQ-KKQ:
I'm sorry. but the tests finish normally when they run locally on ubuntu 22.04.
And with the older Abseil, the tests passes.

paulfd:
I can revert abseil up to some version. I think I needed some form of the variant, I can double check. Do you know why it hangs with the newer abseil?

redtide:
Sorry but I don't understand why you care about Actions, it's a way for us to test the builds, if they hangs it's an Actions issue IMO, not yours for sure, they should timeout but for some reason they don't?

KKQ-KKQ:
the default timeout is 6 hours.

hangs is my miss, I think. I will change std::condition_variable::wait() to std::condition_variable::wait_for().
but the finish of initializing should be notified to the all receiver without fork, I think. I don't know why.

Anyway with the newer Abseil-cpp, the develop branch tests fails in my environment.

KKQ-KKQ:
I can set timeout by writing build.yml. next time I will set it to 20 minutes.

KKQ-KKQ:
@redtide the time is for fee. in the free plan it is limited to 2000 minutes as I know.

redtide:
I see, thank you, so we must limit that but I'm not sure which timing is best, with macOS step in particular I'm not sure 20 could be OK, it should I guess. TBH I wasn't much aware about setting limits in the workflow file

KKQ-KKQ:
the actual limit is 3000 minutes per month.
I will set timeout next time.

redtide:
I think we should see the average and set the appropriate one, I see it's per job, so some are quick, other not

redtide:
Looking at the process that takes longer (macOS, max 6 minutes) would be OK to set them all to 10 and later maybe we can fine tune each?
It seems a global setting is [not yet implemented](https://github.com/orgs/community/discussions/10690).

redtide:
> Oh, I didn't know we used release candidate versions other than LTS, so at this point maybe it worth to try the current 20240116.0 of 2 weeks ago maybe?

@KKQ-KKQ do you mind to try to update the abseil submodule to abseil/abseil-cpp@4a2c633 and see if it is good, so in which case we can update instead roll back to 20230125.3?

EDIT: I see now that Paul needed some newer version for new code, so IIUC roll back couldn't even be possible, so maybe updating might fix the issue?

BTW some time ago I added manual workflows, so in future, if you want to test your fork before submit a PR you should be able to do it. If not then we have a problem and I need to set it so people can enable actions on their side without the need to submit the PR first.

KKQ-KKQ:
This PR is too buggy!

For better performance, the number of loading thread is 1.
Disk is too slow to access randomly, so sliced loading job is better.

KKQ-KKQ:
I think it becomes more stable. Please check this PR.

src/sfizz/AudioReader.cpp
```diff
@@ -15,6 +15,10 @@
 
 namespace sfz {
 
+#if defined(SFIZZ_USE_SNDFILE)
+static bool formatHasFastSeeking(int format);
```
paulfd, line 19: Is this one defined in libsndfile?

src/sfizz/FilePool.cpp
```diff
@@ -46,27 +46,114 @@
...
+std::shared_ptr<sfz::FilePool::GlobalObject> sfz::FilePool::getGlobalObject()
+{
+    std::lock_guard<std::mutex> lock { globalObjectMutex };
+    
+    std::shared_ptr<GlobalObject> globalObject = globalObjectWeakPtr.lock();
+    if (globalObject)
+        return globalObject;
+
+#if 1
+    constexpr size_t numThreads = 1;
```
paulfd, line 80: That change can be baked in. iirc sforzando also seems to not try and parallelize reads. There are little situations where this would seem worthwile to do so.

src/sfizz/FilePool.h
```diff
@@ -346,40 +417,73 @@ class FilePool {
...
     std::mutex loadingJobsMutex;
     std::vector<std::future<void>> loadingJobs;
     std::thread dispatchThread { &FilePool::dispatchingJob, this };
-    std::thread garbageThread { &FilePool::garbageJob, this };
 
-    SpinMutex garbageAndLastUsedMutex;
-    std::vector<FileId> lastUsedFiles;
-    std::vector<FileAudioBuffer> garbageToCollect;
+public:
+    struct GlobalObject
```
paulfd, line 442: I'm not the biggest fan of this name, it feel a bit vague. I guess that this is somehow an extension of the "global threadpool" which includes both the threadpool and the filepool, but maybe we can find another more precise name (or even just change the name of the per-instance "filepools" to something else).

As I'm reading more I might have other ideas :)

tests/FilePoolT.cpp
```diff
@@ -0,0 +1,327 @@
...
+        void wait()
+        {
+            fileData.addSecondaryOwner(&filePool);
+        }
+    };
+    ProcessData p;
+    std::thread t1([&p] { p.wait(); });
+    std::thread t2([&p] { p.prepare(); });
+    t1.join();
+    t2.join();
```
paulfd, line 49: Where these the reasons for hanging tests? We could also time out on a condition variable shared by the threads. It's a bit more verbose though.

tests/FilePoolT.cpp
```diff
@@ -0,0 +1,327 @@
...
+    CHECK(filePoolGlobalObj->getNumLoadedSamples() == 1);
+
+    synth2->loadSfzFile(fs::current_path() / "tests/TestFiles/looped_regions.sfz");
+
+    CHECK(synth1->getNumPreloadedSamples() == 1);
+    CHECK(synth2->getNumPreloadedSamples() == 1);
+    CHECK(synth3->getNumPreloadedSamples() == 0);
+    CHECK(filePoolGlobalObj->getNumLoadedSamples() == 1);
+
+    synth1->loadSfzFile("");
```
paulfd, line 96: Is this supposed to be here?

src/sfizz/FilePool.cpp
```diff
@@ -617,57 +795,103 @@ void sfz::FilePool::setRamLoading(bool loadInRam) noexcept
...
+
+                    if (data->canRemove()) {
+                        preloadedFiles.erase(copyIt);
+                        continue;
+                    }
+
+                    if (data->availableFrames == 0 || data->readerCount != 0)
+                        continue;
+
+                    switch (FileData::Status(data->status)) {
```
paulfd, line 856: Maybe this could be an `if` statement for now.

tests/FilePoolT.cpp
```diff
@@ -0,0 +1,327 @@
...
+    CHECK(filePoolGlobalObj->getNumLoadedSamples() == 6);
+
+    synth1->loadSfzFile(fs::current_path() / "tests/TestFiles/wavetables.sfz");
+    for (unsigned i = 0; i < 100; ++i) {
+        synth1->renderBlock(buffer);
+        synth2->renderBlock(buffer);
+        synth3->renderBlock(buffer);
+        wait_ms(10);
+    }
+    wait_ms(1100);
```
paulfd, line 153: Why do we need these waits?

tests/FilePoolT.cpp
```diff
@@ -0,0 +1,327 @@
...
+        sfz::AudioBuffer<float> buffer { 2, 256 };
+        sfz::Synth synth {};
+        std::unique_ptr<std::thread> thread {};
+        RTSemaphore semBarrier { 0 };
+        std::atomic<bool> running { true };
+        std::atomic<Status> status { Status::Initialized };
+        std::atomic<bool> flag { true };
+    };
+
+    std::unique_ptr<TestSynthThread> synthThreads[synthCount];
```
paulfd, line 257: Could we make a vector of these?

paulfd (commented):
There's a lot of work here, thanks a lot. Sadly I think I need to find more than an hour here and there to fully sit down and look at all this. I put some random comments but nothing really deep, sorry.

One thing: is it OK if I separate the `seek` part of the audiofile library? It's not much but it'll help this PR to be more focused.

src/sfizz/AudioReader.cpp
KKQ-KKQ: No. it is defined later in the file.

src/sfizz/FilePool.cpp
KKQ-KKQ: Previous:
sfizz loads sample files at parallel threads. once it starts loading, never stops until it finishes.

This PR:
sfizz loads samples files on sliced jobs at a single thread. the files to load is cyclic.

discussion:
parallel disk access is really slow. i received a report that shows me that terrible noises occurs on M1 iPad Pro when the person uses isfizz with other samplers. I guessed it is from disk access and it gets better.

src/sfizz/FilePool.h
KKQ-KKQ: I agree. but I am not good at English. Could you suggest one or more?

tests/FilePoolT.cpp
KKQ-KKQ: I added this for my local tests. I agree for that time out should be implemented.

src/sfizz/FilePool.cpp
KKQ-KKQ: data->status is std::atomic value, so you should define a variable when we use an `if` statement.

tests/FilePoolT.cpp
KKQ-KKQ: For waiting garbage collection.

tests/FilePoolT.cpp
KKQ-KKQ: Yes but I prefer using std::unique_ptr personally.
The memory pointer is stable and faster.

KKQ-KKQ:
They should be "Shared File Pool" and "Sliced Loading Job" when we separate this PR.
I implemented both in this PR. `seek` function is needed by "Sliced Loading Job".

tests/FilePoolT.cpp
KKQ-KKQ: Locally the line 107 was failed before because of too much release.

src/sfizz/FilePool.h
```diff
@@ -346,40 +417,73 @@ class FilePool {
...
+        static RTSemaphore semGarbageBarrier;
+        void garbageJob();
+
+        std::unique_ptr<ThreadPool> threadPool;
+
+        std::mutex loadedFilesMutex;
+        std::mutex preloadedFilesMutex;
+
+        // Preloaded data
+        absl::flat_hash_map<FileId, std::shared_ptr<FileData>> preloadedFiles;
```
paulfd, line 471: Hey, just a quick question: what made you change these to shared pointers? My concern is that this shared pointer is copied in `getFilePromise`, and this copy will most probably take a locking mutex, which I want to avoid on the realtime thread.

src/sfizz/FilePool.cpp
```diff
@@ -97,24 +184,34 @@ sfz::FileAudioBuffer readFromFile(sfz::AudioReader& reader, uint32_t numFrames)
...
-    size_t inputFrameCounter { 0 };
-    size_t outputFrameCounter { 0 };
+    size_t inputFrameCounter { filledFrames };
+    size_t outputFrameCounter { inputFrameCounter };
     bool inputEof = false;
+    bool seekable = reader.seekable();
+    if (seekable)
+        reader.seek(inputFrameCounter);
+
+    int chunkCounter = (freeWheeling || !seekable) ? INT_MAX : static_cast<int>(sfz::config::numChunkForLoadingAtOnce);
```
paulfd, line 208: So this chunking allows for a better scheduling of loading multiple files at once right? Forcing a round robin.

tests/FilePoolT.cpp
paulfd: Got it, thanks!

tests/FilePoolT.cpp
paulfd: OK!

src/sfizz/FilePool.h
paulfd: I will look at the rest. My intuition is that "Filepool" can actually be the Global Object and that we have other names for the filepool reference in each synth.

src/sfizz/FilePool.cpp
paulfd: Yes that's a much better approach. ~~I also think we could do this separately from the rest of the PR.~~ I see you mention this in another comment actually!

src/sfizz/FilePool.cpp
paulfd: I see I missed this, thanks.

src/sfizz/FilePool.h
KKQ-KKQ: `getFilePromise` doesn't lock any mutex
In my PR, FilePool's preloadedFiles is synchronized GlobalObject's one.
std::shared_ptr itself doesn't use mutex because it is not thread-safe when written.

src/sfizz/FilePool.h
KKQ-KKQ: I see. we don't have to change the names, in my opinion.
a FilePool object belongs to a Synth object (resources), FilePool::GlobalObject is a singleton object.

src/sfizz/FilePool.h
KKQ-KKQ: FYI, the performance is much improved when I just changed FileData to std::shared_ptr&lt;FileData&gt;.

KKQ-KKQ:
I describe this PR. I'm sorry for late.

## Shared File Pool:

### In FilePool::preloadFiles(), FilePool::loadFiles(), FilePool::loadFromRam():
Load data from GlobalObject::(pre)loadedFiles when the GlobalObject::(pre)loadedFiles contains the FileId.
Add data to GlobalObject::(pre)loadedFiles when the FileId is newly created.
The stored data became std::shared_ptr in order to share the same data.

## Sliced Loading Job:

### Added status:
FileData::Status::PendingStreaming

### In streamFromFile():
It returns false when the size of the loading reaches a certain size.

### In FilePool::loadingJob():
Only execute when the FileData::status is PendingStreaming.
It queues the job again when streamFromFile() returns false.

## Avoid from loading when the data is full-loaded

### Added status:
FileData::Status::FullLoaded

### In FilePool::preloadFiles(), FilePool::loadFiles(), FilePool::loadFromRam():
Set its status to FullLoaded when full-loaded.

### In FilePool::getFilePromise():
It queue the job only when the status is not FullLoaded.

paulfd:
Thanks! So if that's OK, I will split this work into these separate components so we can integrate and work on them one by one.

paulfd:
> The stored data became std::shared_ptr in order to share the same data.

This was regarding my comment above? I'm still concerned about this but I do understand the underlying need. Let's pick up this discussion on the split pull request.

src/sfizz/FilePool.cpp
KKQ-KKQ: Yes.

KKQ-KKQ (commented):
I am very new to github, so many comments were "pending".

KKQ-KKQ (commented):
I made a mistake in resolving. Sorry about that.

KKQ-KKQ:
I have divided this PR into 4 PRs.
I close this PR temporally.

PR #1239, opened 2024-02-08
Added invoke.hpp to git submodule check.

KKQ-KKQ:
In order to build without `git submodule init`, I added invoke.hpp to the submodule check list.


redtide:
Thank you!

PR #1242, opened 2024-02-10
Fix a ModMatrix bug

paulfd:
We took a reference to a target member, while possibly moving the vector of all targets. Recent changes to abseil's ASAN behavior made this more evident.

Also updated abseil to the correct lts target; the previous one was a release candidate...


paulfd:
Note: remove the commit that changes the `USE_SYSTEM_ABSEIL` flag when merging. @redtide I think we should have either on branches or at least when merging to master an ASAN build with the vendored abseil; we missed this check for this bug.

PR #1243, opened 2024-02-13
Fixed a bug that causes invalid effect output

KKQ-KKQ:
The tempMixSpan didn't reset at a loop.
This cause an issue with the following sfz example.

```
<effect>
bus=fx1 type=fverb reverb_type=mid_hall reverb_dry=80 reverb_wet=20 reverb_input=40 reverb_size=0 reverb_predelay=0.01 reverb_tone=50 reverb_damp=50 fx1tomix=100
<global>
effect1=100
<region>sample=*sine
ampeg_attack=0.02 ampeg_release=0.1
output=0
<region>sample=*saw
ampeg_attack=0.02 ampeg_release=0.1
output=3
```

The 2nd and the 3rd output should be silent but actually it sounds.


paulfd:
Nice catch thanks!

PR #1246, opened 2024-02-14
Added status full loaded into FileData

KKQ-KKQ:
PR 1/4 of #1236 Shared File Pool

Added FileData::Status::FullLoaded.

### DIFF
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -683,8 +683,13 @@ void sfz::FilePool::triggerGarbageCollection() noexcept
             return status == FileData::Status::Preloaded || status == FileData::Status::FullLoaded;
         }
 
-        if (status != FileData::Status::Done)
-            return false;
+        switch (status) {
+            case FileData::Status::Invalid:
+            case FileData::Status::Streaming:
+                return false;
+            default:
+                break;
+        }
 
         if (data.readerCount != 0)
             return false;
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -467,12 +467,12 @@ void sfz::FilePool::setPreloadSize(uint32_t preloadSize) noexcept
     for (auto& preloadedFile : preloadedFiles) {
         auto& fileId = preloadedFile.first;
         auto& fileData = preloadedFile.second;
-        const uint32_t maxOffset = fileData.information.maxOffset;
+        const auto maxOffset = fileData.information.maxOffset;
         fs::path file { rootDirectory / fileId.filename() };
         AudioReaderPtr reader = createAudioReader(file, fileId.isReverse());
-        const uint32_t frames = static_cast<uint32_t>(reader->frames());
-        const uint32_t framesToLoad = min(frames, maxOffset + preloadSize);
-        fileData.preloadedData = readFromFile(*reader, framesToLoad);
+        const auto frames = reader->frames();
+        const auto framesToLoad = min(frames, maxOffset + preloadSize);
+        fileData.preloadedData = readFromFile(*reader, static_cast<uint32_t>(framesToLoad));
         fileData.fullyLoaded = frames == framesToLoad;
     }
 }
--- a/src/sfizz/FilePool.cpp
+++ b/src/sfizz/FilePool.cpp
@@ -332,11 +332,14 @@ bool sfz::FilePool::preloadFile(const FileId& fileId, uint32_t maxOffset) noexce
 
     const auto existingFile = preloadedFiles.find(fileId);
     if (existingFile != preloadedFiles.end()) {
-        if (framesToLoad > existingFile->second.preloadedData.getNumFrames()) {
+        auto& fileData = existingFile->second;
+        if (framesToLoad > fileData.preloadedData.getNumFrames()) {
             preloadedFiles[fileId].information.maxOffset = maxOffset;
             preloadedFiles[fileId].preloadedData = readFromFile(*reader, framesToLoad);
+            if (fileData.status == FileData::Status::Preloaded)
+                fileData.status = frames == framesToLoad ? FileData::Status::FullLoaded : FileData::Status::Preloaded;
         }
-        existingFile->second.preloadCallCount++;
+        fileData.preloadCallCount++;
     } else {
         fileInformation->sampleRate = static_cast<double>(reader->sampleRate());
         auto insertedPair = preloadedFiles.insert_or_assign(fileId, {
@@ -344,7 +347,7 @@ bool sfz::FilePool::preloadFile(const FileId& fileId, uint32_t maxOffset) noexce
             *fileInformation
         });
 
-        insertedPair.first->second.status = FileData::Status::Preloaded;
+        insertedPair.first->second.status = frames == framesToLoad ? FileData::Status::FullLoaded : FileData::Status::Preloaded;
         insertedPair.first->second.preloadCallCount++;
     }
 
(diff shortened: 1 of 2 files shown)


src/sfizz/FilePool.h
```diff
@@ -67,7 +67,7 @@ struct FileInformation {
 // Strict C++11 disallows member initialization if aggregate initialization is to be used...
 struct FileData
 {
-    enum class Status { Invalid, Preloaded, Streaming, Done };
+    enum class Status { Invalid, Preloaded, Streaming, Done, FullLoaded };
```
paulfd: Let's call this `FullyLoaded`.

src/sfizz/FilePool.cpp
```diff
@@ -509,7 +532,10 @@ void sfz::FilePool::loadingJob(const QueuedFileData& data) noexcept
 
     streamFromFile(*reader, data.data->fileData, &data.data->availableFrames);
 
-    data.data->status = FileData::Status::Done;
+    currentStatus = data.data->status.load();
+    if (currentStatus == FileData::Status::Streaming) {
```
paulfd: Given the `compare_exchange` below, shouldn't this be `while (currentStatus != FileData::Status::Done)`?

src/sfizz/FilePool.cpp
```diff
@@ -458,10 +466,25 @@ void sfz::FilePool::setPreloadSize(uint32_t preloadSize) noexcept
 
     // Update all the preloaded sizes
     for (auto& preloadedFile : preloadedFiles) {
-        const auto maxOffset = preloadedFile.second.information.maxOffset;
-        fs::path file { rootDirectory / preloadedFile.first.filename() };
-        AudioReaderPtr reader = createAudioReader(file, preloadedFile.first.isReverse());
-        preloadedFile.second.preloadedData = readFromFile(*reader, preloadSize + maxOffset);
+        auto& fileId = preloadedFile.first;
+        auto& fileData = preloadedFile.second;
+        const uint32_t maxOffset = fileData.information.maxOffset;
```
paulfd: Maybe we need to make the `maxOffset` in `information` a `uint32_t`. I don't recall why I put an `int64_t`, maybe for full generality and to better support integer arithmetics since you tend to add or remove offsets around. Was there a reason for you to force `uint32_t` here?

src/sfizz/FilePool.cpp
```diff
@@ -458,10 +466,25 @@ void sfz::FilePool::setPreloadSize(uint32_t preloadSize) noexcept
...
+
+        FileData::Status status = fileData.status;
+        bool fullLoaded = frames == framesToLoad;
+        if (fullLoaded && status != FileData::Status::FullLoaded) {
+            // Forced overwrite status
+            fileData.status = FileData::Status::FullLoaded;
+        }
+        else if (!fullLoaded && status == FileData::Status::FullLoaded) {
+            // Exchange status
+            fileData.status.compare_exchange_strong(status, FileData::Status::Preloaded);
```
paulfd: Same comment as below re: `compare_exchange` without a while loop.

src/sfizz/FilePool.cpp
```diff
@@ -619,10 +645,13 @@ void sfz::FilePool::setRamLoading(bool loadInRam) noexcept
...
             fs::path file { rootDirectory / preloadedFile.first.filename() };
             AudioReaderPtr reader = createAudioReader(file, preloadedFile.first.isReverse());
-            preloadedFile.second.preloadedData = readFromFile(
+            auto& fileData = preloadedFile.second;
+            fileData.preloadedData = readFromFile(
                 *reader,
-                preloadedFile.second.information.end
+                fileData.information.end
             );
+            // Initially set status
```
paulfd: We can remove this comment I think.

src/sfizz/FilePool.cpp
```diff
@@ -648,21 +677,33 @@ void sfz::FilePool::triggerGarbageCollection() noexcept
...
 
         if (data.readerCount != 0)
             return false;
 
+        auto status = data.status.load();
+        if (data.availableFrames == 0) {
+            return status == FileData::Status::Preloaded || status == FileData::Status::FullLoaded;
+        }
+
+        switch(status) {
```
paulfd: Previously you mentioned the switch was needed because `status` was an atomic, but here it's not, right? Would it be okay to make this an `if`? It's just so much clearer to me.

paulfd (changes requested):
2 questions:
- If we move to having a single background thread for loading, this logic and the atomics would get much easier, wouldn't it?
- It seems we're so close to not needing the `loadedFiles`. The only issue I see is if someone calls `setRamLoading(true)`, and then `setRamLoading(false)`, which would erase the `FullyLoaded` states from before.

paulfd (changes requested):
Edit: duplicate comment

paulfd (commented):
And thanks for the PR as always :)

src/sfizz/FilePool.h
KKQ-KKQ: It might be better that we have a boolean fullyLoaded variable on FileData instead of adding it's status enum.

src/sfizz/FilePool.cpp
KKQ-KKQ: status should not be changed to Done when the status is changed to FullLoaded.
but it is a little bit dangerous in the following situation:
1. fileData starts streaming
2. status is changed to FullLoaded (e.g. setRamLoading(true))
3. status is changed to Preloaded (e.g. setRamLoading(false))
4. fileData finishes streaming

I will consider more.

src/sfizz/FilePool.cpp
KKQ-KKQ: No. I will fix this.

src/sfizz/FilePool.cpp
KKQ-KKQ: I don't care either way, I will change this if you prefer `if`.

src/sfizz/FilePool.cpp
KKQ-KKQ: In this process, the status should not be overwritten if the status is changed at the moment.

src/sfizz/FilePool.cpp
KKQ-KKQ: I added comments on every status change.

KKQ-KKQ:
> 2 questions:
> 
> * If we move to having a single background thread for loading, this logic and the atomics would get much easier, wouldn't it?

If we moved to having a single background thread for loading and having shared file pool, an instance have to wait until another instance finishes loading.

> * It seems we're so close to not needing the `loadedFiles`. The only issue I see is if someone calls `setRamLoading(true)`, and then `setRamLoading(false)`, which would erase the `FullyLoaded` states from before.

`loadedFiles` is for smaller files and for wave-oscillators and the samples in it should not be released even when `preloadSize` is changed.
I think we needs the `loadedFiles`.

KKQ-KKQ:
> 2 questions:
> 
> * If we move to having a single background thread for loading, this logic and the atomics would get much easier, wouldn't it?

Oh do you mean `finalizeSfzLoad()` just queue loading? I think it might be more complex, but worth consider it.

KKQ-KKQ:
I think the processes become more safe.

src/sfizz/FilePool.cpp
paulfd: Thanks!

paulfd:
Looks good, I'll try to stress test it a bit today or tomorrow thanks!

paulfd:
@KKQ-KKQ I'm sorry I've been way too swamped at work. I'm merging this without extensive testing so you can proceed. I'll test out things as they come when I can...

すみません ! 🙇

KKQ-KKQ:
> @KKQ-KKQ I'm sorry I've been way too swamped at work. I'm merging this without extensive testing so you can proceed. I'll test out things as they come when I can...
> 
> すみません ! 🙇

No problem! Thank you!

PR #1248, opened 2024-02-23
Update atomic_queue to latest version

taylordotfish:
This PR updates [atomic\_queue](https://github.com/max0x7ba/atomic_queue) to the latest stable version, v1.6.1, which includes support for additional architectures, including RISC-V (fixes #1023) and PowerPC, allowing sfizz to be built on those architectures.

### DIFF
--- a/external/atomic_queue/include/atomic_queue/atomic_queue.h
+++ b/external/atomic_queue/include/atomic_queue/atomic_queue.h
@@ -29,12 +29,14 @@ namespace details {
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 template<size_t elements_per_cache_line> struct GetCacheLineIndexBits { static int constexpr value = 0; };
-template<> struct GetCacheLineIndexBits<64> { static int constexpr value = 6; };
-template<> struct GetCacheLineIndexBits<32> { static int constexpr value = 5; };
-template<> struct GetCacheLineIndexBits<16> { static int constexpr value = 4; };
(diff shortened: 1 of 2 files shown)


external/atomic_queue/include/atomic_queue/atomic_queue.h
```diff
@@ -76,8 +75,8 @@ constexpr T& map(T* elements, unsigned index) noexcept {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-// Implement a "bit-twiddling hack" for finding the next power of 2
-// in either 32 bits or 64 bits in C++11 compatible constexpr functions
+// Implement a "bit-twiddling hack" for finding the next power of 2 in either 32 bits or 64 bits
+// in C++11 compatible constexpr functions. The library no longer maintains C++11 compatibility.
```
paulfd, line 79: My concern was this. I'll check but one of my personal targets used to have a C++11 limitation, which I think is lifted now.

paulfd: That being said, as far as I remember, the only issue was some multi-line `constexpr` functions which are just cosmetics. We could very well rewrite these as C++11 compatible constructs, which is what I did at the time but Maxim did not want to constrain his library in the end.

paulfd:
As explained in the comments I'll check my other toolchains as soon as I can. If we move forward, we also don't need to vendor this anymore.

If it happens that the toolchain breaks, then I'll merge the compatibility bits and relevant changes in the vendored library so it'll still work out.

Thanks a lot for the PR!

PR #1253, opened 2024-02-28
feat: add pause support for riscv

XieJiSS:
This PR supersedes https://github.com/sfztools/sfizz/pull/1023

### DIFF
--- a/external/atomic_queue/include/atomic_queue/defs.h
+++ b/external/atomic_queue/include/atomic_queue/defs.h
@@ -31,6 +31,12 @@ static inline void spin_loop_pause() noexcept {
      defined(__ARM_ARCH_8A__) || \
      defined(__aarch64__))
     asm volatile ("yield" ::: "memory");
+#elif defined(__riscv)
+#if defined(__riscv_zihintpause)
+    asm volatile ("pause" ::: "memory");
+#else
+    /* Encoding of the pause instruction, will be treated as nop if not supported by hardware */
+    asm volatile (".4byte 0x100000F");
 #else
     asm volatile ("nop" ::: "memory");
 #endif
--- a/src/external/cpuid/platform/src/platform/config.hpp
+++ b/src/external/cpuid/platform/src/platform/config.hpp
@@ -55,6 +55,10 @@
 #define PLATFORM_MIPS 1
 #define PLATFORM_CLANG_MIPS 1
 #define PLATFORM_GCC_COMPATIBLE_MIPS 1
+#elif defined(__riscv)
+#define PLATFORM_RISCV 1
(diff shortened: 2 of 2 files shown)


redtide:
Thank you! I wonder if these changes could be done upstream to cpuid and atomic_queue instead have to maintain 2 patches just for 2 simple changes?

external/atomic_queue/include/atomic_queue/defs.h
```diff
@@ -31,6 +31,12 @@ static inline void spin_loop_pause() noexcept {
      defined(__ARM_ARCH_8A__) || \
      defined(__aarch64__))
     asm volatile ("yield" ::: "memory");
+#elif defined(__riscv)
```
taylordotfish, line 34: This is still within the `#elif defined(__arm__) || defined(__aarch64__)` above, so it will never actually apply. Also, the upstream atomic\_queue [already added RISC-V support](https://github.com/max0x7ba/atomic_queue/blob/083f3e5d666101e14008fc2084af4de4836c8bc5/include/atomic_queue/defs.h#L56) with something that looks potentially equivalent to what you've added here.

XieJiSS: Yes the upstream version is indeed the same thing as this one, despite that the .insn i pseudo-instruction is not supported until llvm >= 14 AFAIK

So basically if we use upstream's implementation then we'll be dropping support of some relatively older (but actually not that old) compilers. Which I think is okay because most RISC-V users know how to upgrade their compiler, but I'd like to know @redtide's opinion

redtide: I think it's up to @paulfd to decide on this, IMO I would not care much about "old" compilers and use upstream to not to have to maintain patched stuff, so it would be nice to me if the upstream atomic_queue could be used as is and patch cpuid instead.

paulfd: Let's update to the latest cpuid. I'll make a note of it.

redtide: > Let's update to the latest cpuid. I'll make a note of it.

There is #1249 on the way but I wasn't precise: with "patch cpuid instead" I meant "patch cpuid __upstream__ ([platform repo](https://github.com/steinwurf/platform)) instead", so @XieJiSS, could you please propose your cpuid changes to them?

XieJiSS: Sure, I'll propose to them hopefully in the next few days

redtide: Thank you very much!

Issue #1266, opened 2024-05-27
Sample JACK client (sfizz_jack) goes into infinite loop on EOF


acli:
sfizz-1.2.3 compiled with gcc 11. If I start sfizz_jack and then hit ctrl-d to try to end the program, sfizz_jack instead goes into an infinite loop and I’ll have to hit ctrl-c to stop it.

Example (ctrl-d was pressed on line 54, i.e., immediately after I got a prompt):

    [...]
        42 Positional arguments: /opt/sfz/data/VCSL-1.2.2-RC/Chordophones/Zithers/Grand Piano, Kawai - Legacy.sfz,
        43 Instrument loaded: /opt/sfz/data/VCSL-1.2.2-RC/Chordophones/Zithers/Grand Piano, Kawai - Legacy.sfz
        44 ===========================
        45 Total:
        46         Masters: 0
        47         Groups: 3
        48         Regions: 126
        49         Curves: 7
        50         PreloadedSamples: 126
        51 ===========================
        52 Unknown opcodes:
        53 
        54 > 
        55 > 
        56 > 
        57 > 
        58 > 
    [...]
     11039 > 
     11040 > 
     11041 > 
     11042 > 
     11043 > ^C> 
     11044 > 
     11045 > 
     11046 > Signal received
     11047 Closing...

acli:
Just tried this on current git version. Same behaviour and can also be reproduced by redirecting stdin from /dev/null.

GeoffreyPlitt:
@acli I love the potential of this project, but I'm not sure how active development is now. Have you had any success with getting a response from the developers?

GeoffreyPlitt:
@RustoMCSpit I'm hitting this, do you know anyone who can solve it?

GeoffreyPlitt:
CC @acli @redtide @KKQ-KKQ Any help with this? Makes it impossible to run as a background job (launching with " &", etc)

GeoffreyPlitt:
@acli Do you know which source lines govern the REPL loop?

GeoffreyPlitt:
The REPL loop seems to be here if anyone knows how to fix it to properly detect EOF and exit: https://github.com/sfztools/sfizz/blob/806bf1dde49621d08c50a6a496b25c42b1afa8b8/clients/jack_client.cpp#L222

In the meantime, my workaround is to send it an empty string and newline once per minute like this:
[...]

PR #1267, opened 2024-05-27
Fixes the infinite-loop-on-eof problem

acli:
This isn’t ideal, but it doesn’t introduce whitespace changes and it does fix the bug where sfizz_jack goes into an infinite loop if stdin gets eof (#1266).


acli:
Please consider #1270 instead

PR #1274, opened 2024-05-29
Fixes the infinite-loop-on-EOF problem (smallest patch)

### LINKED ISSUE #1266: Sample JACK client (sfizz_jack) goes into infinite loop on EOF

acli:
This fixes #1266. It’s not really an idiomatic way to fix this but this is the smallest patch that works and plays nicer with the other fix.

### DIFF
--- a/clients/jack_client.cpp
+++ b/clients/jack_client.cpp
@@ -259,7 +259,7 @@ void cliThreadProc()
             } catch (...) {
                 std::cout << "ERROR: Can't set num of voices!\n";
             }
-        } else if (kw == "quit") {
+        } else if (kw == "quit" || !std::cin) {
             shouldClose = true;
         } else if (kw.size() > 0) {
             std::cout << "ERROR: Unknown command '" << kw << "'!\n";


paulfd:
Thanks!

PR #1281, opened 2024-06-21
Fix SFIZZ_RELEASE_ASSERTS option

chucknology:
The SFIZZ_RELEASE_ASSERTS option was ignored and always in effect.


paulfd:
Thanks!

PR #1283, opened 2024-07-09
Add multiple outputs support to sfizz_jack

lucianoiam:
This patch adds a new switch `--num_outputs` to the JACK client to set the number of stereo output pairs. Default is 1 to keep the behavior of the previous version.

Not sure about the value of `MAX_NUM_OUTPUTS`.


paulfd:
Thanks! It's most probably less than 1024, hopefully I'll find time to look into the code.

PR #1320, opened 2025-10-07
Fixes for build errors when using CMake `-G Xcode`

geraintluff:
On Mac, CMake's Makefile generator seems to build everything with dual-architecture.  However, the Xcode generator seems to build the architectures individually and then stick them together afterwards.

Since `PROJECT_SYSTEM_PROCESSOR` will contain both architectures, this always adds x86/x86_64 flags (e.g. `-msse2`) which cause an error when compiling the Apple Silicon part.

This PR uses Clang's `-Xarch_...` filters to say those flags are only relevant for certain architectures.


essej:
The atomic_queue clang build fix patches are also in there... just thought they should be mentioned too :)

Issue #1313, opened 2025-04-03
Feature request: MPE support


jamshark70:
(Or is this better as a discussion?)

I'd like to ask if there are any plans to implement Multichannel Polyphonic Expression in sfizz.

Currently, AFAICS, sfizz responds to all MIDI channels, but the CC and pitch bend states are shared globally -- effectively, there is one channel of response, which happens to accept messages from all channels. I verified this by using SuperCollider with IEM's great VSTPlugin add-on, to round-robin the notes across channels and send a randomized pitch bend with each note, to the same channel. Each PB message affects not only its own, but also any previous ones that hadn't released.

MPE would have independent PB, CC and aftertouch (including poly aftertouch, I think) per channel.

My specific interest is I'm looking for an alternative to Plogue's ancient VST2 sforzando plugin, which is unique in its support for the VST third data byte for detuning. I *love* generating detuned notes in my software (e.g. [here](https://github.com/sfztools/sfizz/discussions/1205), the comment about just intonation), but wine + yabridge + an old, unsupported plug-in using a deprecated API is... fragile. Sfizz is just about in the pocket, but it doesn't support that third byte, nor does it support MTS, nor MPE. I'd be willing to work with any of those. (Scala files don't meet my needs.)

MPE is growing in popularity, might be good to consider it.

rullopat:
Hi @paulfd, @jpcima, @jamshark70 — opened a draft PR addressing this: sftools/sfizz#1327.

The PR adds full per-note pitch bend, per-note CC modulation, and per-note channel aftertouch routing through `MidiState`, `Voice`, `Synth`, and the mod-matrix sources, plus MPE-aware voice stealing and an `setMPEEnabled` / `setMPEPitchBendRange` config surface on the public `sfz::Sfizz` wrapper. Seven independently-buildable commits totalling ~720 insertions; the full sfizz test suite (492 cases / 52269 assertions) plus 17 new MPE-specific regression tests pass.

Hand-tested against an Expressive E Osmose in Logic Pro through a downstream commercial product, so the design has been exercised against real MPE traffic (pre-noteOn per-note state, fast channel-pressure ramps, slow CC74 keytop sweeps, voice-stealing under sustained chords). Five fixes that the regression suite missed (because every test previously wrote events at delay 0 on master) surfaced and landed during that hand-test — they're in commit 7 of the PR.

Happy to iterate on:

[...]

PR #1327, opened 2026-05-10
MPE: per-channel pitch / CC / aftertouch + MPE-aware voice stealing

rullopat:
This PR addresses the MPE feature request in #1313 — adding full per-note pitch bend, per-note CC modulation, and per-note channel aftertouch routing through sfizz so a polyphonic chord on an MPE controller (Osmose, Seaboard, LinnStrument, Push 3) renders with each finger's modulation applied to its own voice instead of collapsing globally.

**Status:** draft, inviting design review. Hand-tested against an Expressive E Osmose in Logic Pro and an Ableton Push 3 in Live 12. Full sfizz suite green (526 cases / 52374 assertions) including a new `tests/MPET.cpp` covering per-channel state isolation, voice-stealing biasing, MCM / RPN 0 auto-configuration, MPE 1.0 spec-compliance filters, and the MPE-off compatibility contract.

## What's in the PR

Logically organised; each commit is independently buildable.

**Core engine plumbing**
- `36a6e09a` — `MidiState`: per-channel `ChannelState` struct (storage refactor; behaviour unchanged with master-only access)
- `a8b28743` — `MidiState` / `Voice`: per-voice `triggerChannel_` plumbing for modulation reads
- `3db6b80a` — `Synth`: channel-aware public API + per-channel dispatch
- `e4812d8b` — `VoiceStealing`: bias victim selection to a preferred MIDI channel
- `5ad530a9` — `MidiState`: lazy member-channel events + regression suite
- `cb3ba1dd` + `002f1395` / `b117153f` / `b9449173` / `c2f81dc7` / `51e0e3a1` — hand-test fixes (channel-aware `Voice::registerNoteOff`, master→member fallback for scalar getters, summed master+member pitch envelope, empty-events guard)

**Public C++ + C API**
- `cd7d7df1` — `sfz::Sfizz` C++ wrapper exposes the new methods
- `2977b355` — parallel C API (`sfizz_send_*_channel`, `sfizz_set_mpe_enabled`, bend-range getters/setters)

**MPE 1.0 spec compliance**
- `2b408275` — MCM (RPN 6) + Pitch Bend Sensitivity (RPN 0) auto-configuration with per-axis opt-outs
- `0a443c14` — drop Poly KP on Member Channels (§2.2.7 / Appendix E Table 5)
- `528a4b9d` — drop Manager-only CCs (pedals, mode/reset, Bank Select) on Member Channels (§2.3.1 / §2.3.3)
- `efaf7c2a` — released-voice expression reads route through the Manager Channel (§2.2.6 / §2.2.7 / §2.2.8 / §A.4.1)

**MPE-off compatibility**
- `36c92314` — regression test asserting legacy-API voices isolate from `*MPE` per-channel writes
- `db9717aa` — engine collapses channel argument to 0 in all channel-aware entry points when MPE is disabled, so legacy and channel-aware paths behave identically. `setMPEEnabled(false)` flushes active voices.

**API hygiene**
- `13337446` — dropped the `*MPE` suffix from channel-aware methods now that they handle both modes; C++ becomes overloads of the legacy names, C API uses `_channel` suffix. MPE config surface (`setMPEEnabled`, bend-range getters/setters, drop counters) unchanged.
- `9eaef3fc` — header doc refresh describing the new contract.

## Design notes

- **MPE config surface**: minimal — `setMPEEnabled(bool)` + `setMPEPitchBendRange(masterSemitones, perNoteSemitones)` + the two RPN opt-out flags. Happy to grow this if you'd prefer an explicit `setMPEZone(masterChannel, memberCount)` shape.
- **CC inheritance**: channel-aware getters fall back to master for empty member-channel vectors. Once a member channel writes a CC, it owns its state.
- **Master CCs (sustain pedal, modwheel, …)** apply globally because they're sent on the master channel; member voices read their own channel's vector, find it empty, fall back to master.
- **No SFZ opcode changes**: existing `*_oncc<N>` opcodes "just work" via per-channel CC banks.
- **Lower Zone only**: Upper Zone (master = channel 16) is deferred — no consumer in scope.

## Why a fork

Issue #1313 had been open since April 2025 with no maintainer engagement, and an MPE-capable engine was needed for a personal project. Rather than block, hard-forked at `f5c6e29f` and have been iterating on the fork since. This PR exists to invite review and find a maintainer-preferred shape so the patch series doesn't have to live in a fork indefinitely.

cc @paulfd, @jpcima, @jamshark70.

### DIFF
--- a/src/sfizz.h
+++ b/src/sfizz.h
@@ -624,7 +624,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_poly_aftertouch(sfizz_synth_t* synth, int 
  * sfizz_set_mpe_enabled() is informational for the engine: it gates
  * same-channel-preference voice stealing. Per-channel input dispatch
  * works regardless of the flag — calling
- * sfizz_send_pitch_wheel_mpe(channel=2, ...) always lands in the
+ * sfizz_send_pitch_wheel_channel(channel=2, ...) always lands in the
  * channel-2 modulation slot.
  *
  * @{
@@ -642,7 +642,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_poly_aftertouch(sfizz_synth_t* synth, int 
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_note_on_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, int velocity);
+SFIZZ_EXPORTED_API void sfizz_send_note_on_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, int velocity);
 
 /**
  * @brief Send a high-precision note on event on a specific MIDI channel.
@@ -656,7 +656,7 @@ SFIZZ_EXPORTED_API void sfizz_send_note_on_mpe(sfizz_synth_t* synth, int delay, 
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hd_note_on_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, float velocity);
+SFIZZ_EXPORTED_API void sfizz_send_hd_note_on_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, float velocity);
 
 /**
  * @brief Send a note off event on a specific MIDI channel (0..15).
@@ -670,7 +670,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_note_on_mpe(sfizz_synth_t* synth, int dela
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_note_off_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, int velocity);
+SFIZZ_EXPORTED_API void sfizz_send_note_off_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, int velocity);
 
 /**
  * @brief Send a high-precision note off event on a specific MIDI channel.
@@ -684,7 +684,7 @@ SFIZZ_EXPORTED_API void sfizz_send_note_off_mpe(sfizz_synth_t* synth, int delay,
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hd_note_off_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, float velocity);
+SFIZZ_EXPORTED_API void sfizz_send_hd_note_off_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, float velocity);
 
 /**
  * @brief Send a CC event on a specific MIDI channel (0..15).
@@ -698,7 +698,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_note_off_mpe(sfizz_synth_t* synth, int del
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_cc_mpe(sfizz_synth_t* synth, int delay, int channel, int cc_number, int cc_value);
+SFIZZ_EXPORTED_API void sfizz_send_cc_channel(sfizz_synth_t* synth, int delay, int channel, int cc_number, int cc_value);
 
 /**
  * @brief Send a high-precision CC event on a specific MIDI channel.
@@ -712,7 +712,7 @@ SFIZZ_EXPORTED_API void sfizz_send_cc_mpe(sfizz_synth_t* synth, int delay, int c
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hdcc_mpe(sfizz_synth_t* synth, int delay, int channel, int cc_number, float norm_value);
+SFIZZ_EXPORTED_API void sfizz_send_hdcc_channel(sfizz_synth_t* synth, int delay, int channel, int cc_number, float norm_value);
 
 /**
  * @brief Send a pitch wheel event on a specific MIDI channel (0..15).
@@ -725,7 +725,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hdcc_mpe(sfizz_synth_t* synth, int delay, int
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_pitch_wheel_mpe(sfizz_synth_t* synth, int delay, int channel, int pitch);
+SFIZZ_EXPORTED_API void sfizz_send_pitch_wheel_channel(sfizz_synth_t* synth, int delay, int channel, int pitch);
 
 /**
  * @brief Send a high-precision pitch wheel event on a specific MIDI channel.
@@ -738,7 +738,7 @@ SFIZZ_EXPORTED_API void sfizz_send_pitch_wheel_mpe(sfizz_synth_t* synth, int del
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hd_pitch_wheel_mpe(sfizz_synth_t* synth, int delay, int channel, float pitch);
+SFIZZ_EXPORTED_API void sfizz_send_hd_pitch_wheel_channel(sfizz_synth_t* synth, int delay, int channel, float pitch);
 
 /**
  * @brief Send a channel aftertouch (channel pressure) event on a specific MIDI channel.
@@ -751,7 +751,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_pitch_wheel_mpe(sfizz_synth_t* synth, int 
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_channel_aftertouch_mpe(sfizz_synth_t* synth, int delay, int channel, int aftertouch);
+SFIZZ_EXPORTED_API void sfizz_send_channel_aftertouch_channel(sfizz_synth_t* synth, int delay, int channel, int aftertouch);
 
 /**
  * @brief Send a high-precision channel aftertouch event on a specific MIDI channel.
@@ -764,7 +764,7 @@ SFIZZ_EXPORTED_API void sfizz_send_channel_aftertouch_mpe(sfizz_synth_t* synth, 
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hd_channel_aftertouch_mpe(sfizz_synth_t* synth, int delay, int channel, float aftertouch);
+SFIZZ_EXPORTED_API void sfizz_send_hd_channel_aftertouch_channel(sfizz_synth_t* synth, int delay, int channel, float aftertouch);
 
 /**
  * @brief Send a polyphonic aftertouch event on a specific MIDI channel.
@@ -778,7 +778,7 @@ SFIZZ_EXPORTED_API void sfizz_send_hd_channel_aftertouch_mpe(sfizz_synth_t* synt
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_poly_aftertouch_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, int aftertouch);
+SFIZZ_EXPORTED_API void sfizz_send_poly_aftertouch_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, int aftertouch);
 
 /**
  * @brief Send a high-precision polyphonic aftertouch event on a specific MIDI channel.
@@ -792,7 +792,7 @@ SFIZZ_EXPORTED_API void sfizz_send_poly_aftertouch_mpe(sfizz_synth_t* synth, int
  * @par Thread-safety constraints
  * - @b RT: the function must be invoked from the Real-time thread
  */
-SFIZZ_EXPORTED_API void sfizz_send_hd_poly_aftertouch_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, float aftertouch);
+SFIZZ_EXPORTED_API void sfizz_send_hd_poly_aftertouch_channel(sfizz_synth_t* synth, int delay, int channel, int note_number, float aftertouch);
 
 /**
  * @brief Enable or disable MPE mode.
--- a/src/sfizz.h
+++ b/src/sfizz.h
@@ -612,6 +612,243 @@ SFIZZ_EXPORTED_API void sfizz_send_poly_aftertouch(sfizz_synth_t* synth, int del
 SFIZZ_EXPORTED_API void sfizz_send_hd_poly_aftertouch(sfizz_synth_t* synth, int delay, int note_number, float aftertouch);
 
 /**
+ * @defgroup MPE MIDI Polyphonic Expression
+ *
+ * Channel-aware variants of the input API. They populate per-channel
+ * modulation state in the underlying engine so voices triggered on a
+ * member channel respond independently to per-note pitch bend, per-note
+ * CC, and per-note aftertouch. Hosts that don't care about MPE can keep
+ * using the existing single-channel methods; they are equivalent to
+ * calling the corresponding `_mpe` method with channel = 0 (master).
+ *
+ * sfizz_set_mpe_enabled() is informational for the engine: it gates
+ * same-channel-preference voice stealing. Per-channel input dispatch
+ * works regardless of the flag — calling
+ * sfizz_send_pitch_wheel_mpe(channel=2, ...) always lands in the
+ * channel-2 modulation slot.
+ *
+ * @{
+ */
+
+/**
+ * @brief Send a note on event on a specific MIDI channel (0..15).
+ *
+ * @param synth        The synth.
+ * @param delay        The delay of the event in the block, in samples.
+ * @param channel      The MIDI channel, in domain 0 to 15.
+ * @param note_number  The MIDI note number, in domain 0 to 127.
+ * @param velocity     The MIDI velocity, in domain 0 to 127.
+ *
+ * @par Thread-safety constraints
+ * - @b RT: the function must be invoked from the Real-time thread
+ */
+SFIZZ_EXPORTED_API void sfizz_send_note_on_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, int velocity);
+
+/**
+ * @brief Send a high-precision note on event on a specific MIDI channel.
+ *
+ * @param synth        The synth.
+ * @param delay        The delay of the event in the block, in samples.
+ * @param channel      The MIDI channel, in domain 0 to 15.
+ * @param note_number  The MIDI note number, in domain 0 to 127.
+ * @param velocity     The normalized MIDI velocity, in domain 0 to 1.
+ *
+ * @par Thread-safety constraints
+ * - @b RT: the function must be invoked from the Real-time thread
+ */
+SFIZZ_EXPORTED_API void sfizz_send_hd_note_on_mpe(sfizz_synth_t* synth, int delay, int channel, int note_number, float velocity);
+
+/**
+ * @brief Send a note off event on a specific MIDI channel (0..15).
+ *
+ * @param synth        The synth.
+ * @param delay        The delay of the event in the block, in samples.
+ * @param channel      The MIDI channel, in domain 0 to 15.
+ * @param note_number  The MIDI note number, in domain 0 to 127.
+ * @param velocity     The MIDI velocity, in domain 0 to 127.
+ *
+ * @par Thread-safety constraints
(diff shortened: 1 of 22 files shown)


paulfd:
Thanks for this PR. I assumed this is AI assisted in some way? (It's not an issue, mind you, just verifying)

One bit of trivia, my earliest version of sfizz had multi-channel :slightly_smiling_face: Before diving in, my main concern is that I wouldn't have much in the ways to test this and the possible since I have no real MPE hardware available.

jamshark70:
@rullopat Delighted to see this! Building it now.

I should set expectations that I'm not a C++ developer. Ad-hoc microtuning is something that I really really want, and I was considering taking some time this summer to see if I could make some sense out of it. *Maybe* with LLM assistance I might have made a little headway, but I'm not at all confident that I would have succeeded.

I will certainly put this through some paces, and if I run into bugs, I might have a look at the sources. Substantial contributions are probably outside of my skill set, though. Sorry if my earlier message made it sound otherwise.

@paulfd 
> my main concern is that I wouldn't have much in the ways to test this and the possible since I have no real MPE hardware available.

Neither do I, but SuperCollider / Max / Pd can generate any MIDI messages that are required. (I'll be running SuperCollider --> VSTPlugin extension --> sfizz.vst3.)

rullopat:
@paulfd yes it's AI assisted, I have some experience in C++ programming but not with audio. I was doing a personal project with JUCE for an autosampler / rompler to use with my MPE controllers (I have an Expressive E Osmose 49 and an Ableton Push 3) that is creating SFZ files and samples in FLAC format, I did a plugin that is using Sfizz for testing the samplers and I needed this MPE features.

@jamshark70 I think I'll try to update also the Sfizz UI so that it supports my fork.

rullopat:
Quick update: pushed `2977b355` to this branch — adds the parallel C API surface (`sfizz_send_*_mpe`, `sfizz_{set,get}_mpe_enabled`, `sfizz_{set,get}_mpe_*_pitch_bend_range`). One-line delegations to the engine, no new behavior. This unblocks updating sfizz-ui so the LV2 / VST3 builds can drive MPE without reaching into the C++ wrapper — I'll take a stab at that side next.

rullopat:
Follow-up: opened sfztools/sfizz-ui#166 — VST3/AU + LV2 + PD dispatch through the new `*_mpe` API plus MPE enable / master bend / per-note bend controls in the Settings panel. Draft because I don't have an MPE controller on hand to verify the end-to-end round-trip. @jamshark70 if your SuperCollider → `sfizz.vst3` path is still on, that PR should now let you drive MPE through it.

I'll try my Osmose and / or Push 3 as soon as I'll have a bit more free time, I just wanted to push the change to Sfizz UI to give the ability to you to test if you want.

jamshark70:
OK, I've got that and I now see the settings switch.

Unfortunately, I don't hear any effect from pitchbend when the new MPE option is enabled.

Also I tried some overlapping notes in different channels, and found that a note-off on the first channel cuts off all matching notes on all channels.

If the sfizz core works in your specific case, maybe the VST plugin wrapper isn't passing the MIDI data accurately? I can confirm that I'm sending different channel numbers, but the plugin is reacting as if it's all one big channel.

rullopat:
@jamshark70 I was trying in Logic Pro that has of course it's AU format with my own plugin in JUCE that it's using the C++ code of the Sfizz library directly, now I'm trying a VST in Ableton and see the problem, I'll let you know if I find out how to fix it, thanks for trying.

jamshark70:
> now I'm trying a VST in Ableton and see the problem

Strange...

I did a little more testing just now:

- Using plugdata (Pd-as-a-VST) in SC's VSTPlugin, I confirm that VSTPlugin (which is loading sfizz into SC) passes channel numbers through correctly.
- Adding a few debugging posts into sfizz, I get:

```
note on channel = 1
noteOnDispatch channel = 1
synth startVoice channel = 1
note on channel = 2
noteOnDispatch channel = 2
synth startVoice channel = 2
```

So the channel number is getting at least that far down.

Out of time for now, but when I have a moment I'll try to trace through note-off and PB value access with MPE on.

jamshark70:
Here's another finding. I was trying to track down what happens to those pitchbend messages.

With MPE off:

```
inserting bend = 0.000061 into channel 0
getPitchBend check events for channel = 1
getPitchBend check events for channel = 1
notenum = 60, channel = 1, bend = 0.000000
inserting bend = 0.854605 into channel 0
getPitchBend check events for channel = 2
getPitchBend check events for channel = 2
notenum = 63, channel = 2, bend = 0.000000
inserting bend = 0.000061 into channel 0
```

With MPE on:

```
getPitchBend check events for channel = 1
getPitchBend check events for channel = 1
notenum = 60, channel = 1, bend = 0.000000
getPitchBend check events for channel = 2
getPitchBend check events for channel = 2
notenum = 63, channel = 2, bend = 0.000000
```

~~In both cases, when playing a note, it's looking to the right channel for a pitchbend data. But with MPE on, it never reaches `pitchBendEvent()`. So nothing is stored into the MIDI state for the channels on which notes are being played.~~

~~I wonder if something similar is causing note events to be stored in the wrong place (but now I *really* am out of time -- need to go to my real job --)~~

Had 5 minutes before starting -- in MPE mode, it goes through performHdcc, not through pitchBendEvent() so I'll have to start again with that later.

rullopat:
Thanks @jamshark70 for the detailed traces — they made both bugs obvious.

Two follow-ups based on what you found:

1. **`002f1395`** in this branch — `Voice::registerNoteOff` now also matches on channel, so a note-off on one member channel no longer releases voices on every other channel at the same pitch.
2. The "no pitch bend in MPE mode" half wasn't in the library — it was the VST3 wrapper dropping `kPidPitchBend` / `kPidAftertouch` when MPE was enabled. Fixed in sfztools/sfizz-ui#166 at **`b0699c8`**.

About your edit re `performHdcc` reached vs. `pitchBendEvent` — I haven't reproduced that path yet. After the two fixes above, per-channel pitch bend still won't propagate end-to-end because the VST3 wrapper's `getMidiControllerAssignment` returns the same paramID regardless of channel, so per-channel bend collapses to one global parameter. That's a separate follow-up.

rullopat:
Two more commits pushed:

- **`b117153f`** — extends the master→member fallback (`cb3ba1dd`) to the scalar `MidiState` getters too. `Voice::Impl::startVoice` seeds the per-voice bend smoother from the scalar `getPitchBend`; without this it started at 0 for member-channel voices even when master had a non-zero bend.
- **`b9449173`** — actually applies the MPE bend ranges. `mpeMasterPitchBendRange_` / `mpePerNotePitchBendRange_` were stored but never read, so member-channel voices used the region's `bend_up`/`bend_down` (default ±2 st) instead of the configured ±48 st per-note range. `pitchEnvelope` also wasn't summing master + member contributions: once a member channel had any per-note bend, its events vector stayed populated and master bend stopped reaching the voice. Now reads own and master events separately (via the new `getPitchEventsRaw`) and combines: `perNote × perNoteRange + master × masterRange`.

Verified against the matching sfztools/sfizz-ui#166 with a Push 3 — per-note slide bend, master touch-strip bend, and combinations of the two are all audible at expected magnitudes.

rullopat:
Pushed **`2b408275`**: the engine now parses RPN 6 (MPE Configuration Message) and RPN 0 (Pitch Bend Sensitivity) per MPE 1.0 §2.

A complete `CC 101=0` / `CC 100=6` / `CC 6=N` sequence on channel 0 (Lower Zone master) calls `setMPEEnabled(N>0)`. An `CC 101=0` / `CC 100=0` / `CC 6=semitones` sequence on the master updates the master bend range; on a member channel it updates the per-note range. CCs still propagate to `MidiState` so SFZ `*_oncc` bindings on the relevant CC numbers keep working — the parser is a tap, not a filter.

Two granular opt-outs (`setMPEMasterBendAutoConfigEnabled`, `setMPEPerNoteBendAutoConfigEnabled`, mirrored through the C++ and C wrappers, default true) let UIs pin a bend range without disabling the MCM auto-enable path. MCM enable is unconditional because that's the spec contract.

Upper Zone (master = channel 15) and the CC 38 cents-fraction path are deferred — no commercial MPE controller in scope uses them. NRPN sequences are tracked separately so a `CC 6` following an NRPN selection doesn't get misinterpreted as RPN data entry.

12 new `[MPE]` cases in `tests/MPET.cpp` cover MCM enable/disable/non-master rejection, RPN 0 master vs member routing, null-RPN and NRPN cross-talk, both opt-out gates, and parser pass-through on CC propagation. Verified with Push 3 in Live 12.

rullopat:
Quick follow-up — testing surfaced a bug on the wrapper side. `SfizzVstProcessor::process()` was re-sending `mpeEnabled` and the bend ranges to the synth on every block, which meant the engine's auto-config from MCM / RPN 0 got clobbered straight away: engine flipped on, wrapper pushed its stale state next block, engine flipped back off. Push 3 didn't catch it (Live's MPE mode doesn't send MCM); something like an Osmose, which announces itself on connect, would have.

Fix is in sfizz-ui across two commits — **`0c9c4f4`** stops the wrapper re-asserting on every block, and **`acd5007`** feeds the engine's auto-configured values back to the host via `outputParameterChanges`, so the editor's MPE controls update in real time and project saves persist the post-MCM state.

rullopat:
Pushed **`0a443c14`**: drops Polyphonic Key Pressure events on Member Channels when MPE is enabled, per MPE 1.0 §2.2.7 (and Appendix E Table 5, which marks Poly KP on Member Channels as Prohibited for both Tx and Rx).

The gate lives at the top of `Synth::hdPolyAftertouchMPE`, so all four poly-aftertouch entry points (`polyAftertouch` / `hdPolyAftertouch` / `polyAftertouchMPE` / `hdPolyAftertouchMPE`) funnel through it; the legacy single-channel paths forward to channel 0 and remain unaffected. Dropped events neither update `MidiState` nor reach the mod-matrix `polyphonicAftertouch` ExtendedCC. Manager-Channel Poly KP is permitted under the spec's "discretion of the implementer" clause and continues to route as today.

A diagnostic counter (`getDroppedPolyKpOnMemberCount`, mirrored through the C++ and C wrappers) lets hosts observe spec-violating traffic for triage.

Three new `[MPE]` cases in `MPET.cpp` cover the Member drop, the Manager-Channel pass-through, and the MPE-disabled bypass.

rullopat:
Pushed **`528a4b9d`**: drops Manager-only CC messages received on Member Channels when MPE is enabled, per MPE 1.0 §2.3.1 / §2.3.3 (Appendix E Table 5 marks pedals, mode/reset and Bank Select as honored only on the Manager Channel).

The filter covers the pedal CCs (64-69 — Damper, Portamento, Sostenuto, Soft Pedal, Legato Footswitch, Hold 2), the mode/reset CCs (120, 121, 123, 124, 125), and Bank Select MSB/LSB (CC 0 / CC 32). It lives at the top of `Synth::Impl::performHdcc`, ahead of the global All-Notes-Off / Reset-All-Controllers early-returns and the RPN parser tap, so a Member-Channel CC 120/121/123 no longer resets the zone by accident and Bank Select on a Member Channel doesn't queue state for a Program Change the host would have to filter separately. Gated on the `asMidi` entry, so internal automation paths (which conceptually target the Manager Channel) are unaffected. CC#7 / CC#10 / CC#11 are intentionally NOT in the list — the spec marks them optional on both channel types.

RPN data CCs (6 / 38 / 98 / 99 / 100 / 101) are deliberately excluded so the existing MPE RPN auto-config (MCM + Pitch Bend Sensitivity) continues to work on Member Channels.

Program Change on Member Channels remains a host-side concern because the engine `programChange` API does not carry a channel argument.

Diagnostic counter `getDroppedManagerOnlyMessageCount` is mirrored through the C++ and C wrappers.

Eight new `[MPE]` cases in `MPET.cpp` cover Damper drop on Member, Damper pass-through on Manager, all six pedal CCs drop, All-Notes-Off / Reset-All / Bank-Select drops, the MPE-disabled bypass, and a regression that RPN data CCs still flow on Member Channels.

rullopat:
Pushed **`efaf7c2a`**: a released voice stops reacting to Pitch Bend, Channel Pressure, CC#74 and Poly Aftertouch sent on its Member Channel after Note Off, per MPE 1.0 §2.2.6 / §2.2.7 / §2.2.8. Manager-Channel traffic still reaches the release tail (§A.4.1).

Single choke point: `Voice::expressionChannel()` returns the trigger channel for active voices and 0 for released voices triggered on a Member Channel. The mod-matrix CC / Channel Pressure / Poly Aftertouch sources, the per-block pitch envelope, and the region crossfade reads all route through it.

Seven new `[MPE]` cases in `MPET.cpp`.

rullopat:
I'm continuing to test with my Push 3 in Ableton Live 12.4, but it seems that MPE support is complete right now and I didn't find any more bugs, but I need to use it a bit more for sure.

@jamshark70 @paulfd in case you would have some free time to test the latest Sfizz with Sfizz-UI in any DAW, that would be great!

Tell me when you think both this PR and the related one for Sfizz-UI are ready to be rewiewed, I would change the status of the PR from draft to ready for review.

jamshark70:
> in case you would have some free time to test the latest Sfizz with Sfizz-UI in any DAW, that would be great!

Here are the two quicky SuperCollider tests I've run. These don't in any way make up a comprehensive suite, but they show that per-channel release and pitch bend are happening -- but they happen whether MPE is turned on or off. That may be a difference in behavior compared to baseline sfizz. I.e., I can get the old sfizz behavior (where all messages collapse down to one channel) if I guarantee that all messages are on channel 1, but if I distribute the messages onto different channels, then they are always per-channel, regardless of the MPE setting.

This is usable for me personally, but it might be an obstacle to an upstream merge. (I'm not sure if sfizz maintainers would insist on full backward compatibility if MPE is off or not.)

Setup:

```
(
SynthDef(\vst, { |out = 0|
	Out.ar(out, VSTPlugin.ar(numOut: 2));
}).add;
)

a = Synth(\vst);
c = VSTPluginController(a);
c.open("sfizz.vst3"/*, mode: \sandbox*/);
m = VSTPluginMIDISender(c);

c.editor;  // and switch MPE on here
```

Per-channel release:

```
(
p = Ppar([
	Pbind(
		\type, \midisend,
		\midisend, m,
		\degree, Pn(Pseries(-3, 1, 12), inf) + #[0, 2, 4],
		\chan, Pn(Pseries(1, 1, 15), inf).clump(3).trace,
		\legato, 2.5,
		\amp, 0.5
	),
]).play;
)

p.stop;
```

The last thing you hear should be a B diminished triad. If per-channel note release is not working, the first chord will cut off the B and D early, and you'll be left only with an F hanging over. I get the full triad :+1:

Per-note pitch bend:

```
(
r = fork {
	var n = MIDINoteMessage(device: m, latency: s.latency);
	var p = MIDIBendMessage(device: m, latency: s.latency);
	// middle C, channel 2
	// play for 4 beats
	s.makeBundle(s.latency, {
		p.play(0, 1);
		n.play(60, 64, 4, 1);
	});

	1.0.wait;
	// middle Eb, channel 3, one beat later, up a quarter tone
	s.makeBundle(s.latency, {
		p.play((8192 / 48 * 0.5).asInteger, 2);
		n.play(63, 64, 4, 2);
	});
};
)
```

I should hear a 3.5 semitone interval, and I do.

Great stuff! Really appreciate this -- my sorta-just-intonation pianos are back :imp:

rullopat:
@jamshark70  thank you for the testing, really appreciate!

About your concern regarding MPE pitchbend, as you can in commit, I had to do it to be in line with the MPE specification:

> Pushed **`efaf7c2a`**: a released voice stops reacting to Pitch Bend, Channel Pressure, CC#74 and Poly Aftertouch sent on its Member Channel after Note Off, per MPE 1.0 §2.2.6 / §2.2.7 / §2.2.8. Manager-Channel traffic still reaches the release tail (§A.4.1).
> 
> Single choke point: `Voice::expressionChannel()` returns the trigger channel for active voices and 0 for released voices triggered on a Member Channel. The mod-matrix CC / Channel Pressure / Poly Aftertouch sources, the per-block pitch envelope, and the region crossfade reads all route through it.
> 
> Seven new `[MPE]` cases in `MPET.cpp`.

But you say that now the behaviour changed when MPE is turned off, compared to how it was before, as soon as I have a bit of time I'll take a look, probably this evening after work.

jamshark70:
> About your concern regarding MPE pitchbend

I don't think I have a concern about this -- it's working here, and doing exactly what I need!

> But you say that now the behaviour changed when MPE is turned off

To clarify -- if I recall correctly, the old sfizz behavior was to ignore MIDI channel completely, and act as if all messages were being sent on a single channel.

With your fork, *both* MPE-on and MPE-off handle note on/off and pitchbend per channel. That's correct for MPE-on, obviously. For MPE-off, I don't know what the sfizz maintainers will say. They could say, "This is great! We always wanted per channel behavior, but it was too big an architecture change." Or they could say that there needs to be a way back to legacy behavior, for users who don't care about (or actively don't want) multichannel MIDI response.

My personal opinion is twofold: 1/ I prefer the multichannel way, and that's how I plan to use it. 2/ I'm sensitive to the need for backward compatibility -- if somebody's DAW project using sfizz starts playing differently because MPE was added, that's bad. I would completely understand if sfizz maintainers argue that MPE-off should behave exactly like sfizz without MPE support at all.

*I* don't need any changes -- I'm happy! But I wouldn't be surprised if this became a conversation when they review for merge.

rullopat:
@jamshark70 your MPE-off concern is fixed in `sftools/sfizz-ui` `4a789ff` — `SfizzVstProcessor` now branches dispatch on the toggle: legacy channel-less API when off (collapses to master channel in `MidiState`), `*MPE` when on. NoteExpression and per-channel paramIDs are skipped when off; `allSoundOff()` on the on→off edge flushes voices with `triggerChannel_>0` that would otherwise hang.

Engine-side regression test in `rullopat/sfizz` `36c92314` asserts legacy-API voices are isolated from `*MPE` per-channel writes. Library bump in `sftools/sfizz-ui` `68c7dea`.

Worth re-running your two SuperCollider reproducers with the switch off — multi-channel input should collapse to one global channel.

rullopat:
Public API rename in `9eaef3fc` (preceded by `13337446`): dropped the `*MPE` suffix from the channel-aware dispatch methods now that they handle both MPE-on and MPE-off routing.

- **C++**: `noteOnMPE` / `ccMPE` / `pitchWheelMPE` / `channelAftertouchMPE` / `polyAftertouchMPE` (+ `hd*` variants) → overloads of the existing legacy names, distinguished by argument count.
- **C**: `sfizz_send_*_mpe` → `sfizz_send_*_channel`.
- **MPE config surface unchanged** (`setMPEEnabled`, bend-range getters/setters, drop counters) — those are genuinely MPE-related.

Doc refresh in `9eaef3fc` updates the API preamble to describe the new contract: with MPE disabled the channel-aware methods collapse channel to 0 internally, so legacy and channel-aware paths are equivalent in that mode. 526/526 tests green.

jamshark70:
Hi, yes, the MPE switch in the GUI does make a difference in behavior now.

- MPE on: Pitch bend on the order of +/- 57 is audible (due to 48 semitone MPE bend range)
- MPE off: Same pitch bend is not audible (due to 2 semitone main bend range), and overlapped notes do get cut off.

Couple of issues:

1. The CC messages to switch on MPE are still not working for me. The exact sequence of bytes I'm sending is `B0 65 00 B0 64 06 B0 06 0F` (copied from debugging output), and the MPE spec says `[0xBn 0x65 0x00] [0xBn 0x64 0x06] [0xBn 0x06 <mm>]` -- I'm only using the lower zone so n = 0. But there's no effect on the settings page or on the sound.
2. I can't seem to change either pitch bend range. Clicking on the bend range number boxes (either of them) has no effect, and I can't type into them either.

rullopat:
Thanks for testing — glad the legacy single-channel path (note stealing, ±2 semitone bend with MPE off) is preserved.

On the two remaining points:

**1. MCM not updating the GUI.** Engine state *is* being applied — your audible ±48 vs ±2 confirms it. Whether the GUI should also reflect received MCM in the visible toggle / bend-range fields isn't industry standard: the MPE 1.0 spec mandates only receiver-side behavioural state. Most MPE-capable plugins treat the visible UI as authoritative and incoming RPN as a side-channel hint.

Can we split this as a follow-up sfizz-ui issue rather than gating merge? Happy to track.

**2. Bend-range number boxes unresponsive.** Real UI bug — I'll fix it in sfztools/sfizz-ui#166. Thanks for catching it.

rullopat:
Quick follow-up — the bend-range widget bug is fixed in sfztools/sfizz-ui#166 (`39e3107` populates the value menus that were silently empty, and `2ced678` adds the integer formatter / width alignment / "disabled when RPN-driven" treatment).

Also pushed `4fb3724` which adds a read-only "Current master / per-note bend value" row beside each override:

- !override / no RPN → MPE default
- !override / RPN received → engine's RPN-driven value
- override / no RPN → override value
- override / RPN received → incoming RPN value, struck through to signal "engine is ignoring this"

Should make it easier to see what the engine is actually using vs. what the user has staged. Would appreciate another pass when you have time.

jamshark70:
Will do, when I can, bit busy these days. Thanks for all the hard work on this!

rullopat:
@jamshark70 since I see not only no interest in this PR but it's completely ignored, I decided to do my own fork with MPE support of the sfizz library called Sfizioso (https://github.com/rullopat/sfizioso) and a complete replacement of sfizz-ui using the Juce library called Sfizioso Player (https://github.com/rullopat/sfizioso-player) if you are interested.

paulfd:
For the record, I did start to look at it and started replies, but I guess I spend so much time looking at AI code at work, I'm just too weary to do so also on my own time. Maybe two general comments then:
- I'm not convinced expanding all the data structures over channels is necessary, but it will work.
- I'd also think it would make sense to see if we could design support for both MIDI 2.0 and MPE in one shot.

I cannot find it in me to review and iterate on this on a timely fashion for you apparently, even if you're the one spending the tokens (and the time, obviously). I’m also not yet comfortable delegating my craft and projects to LLMs without oversight. So forking is the right move. And the JUCE port of the UI is also a right move.

I'm happy sfizz is usefull as a base for you, I'll archive this repository to avoid mixed feelings from other people later.

Thanks!

